PR #4, opened 2015-05-08
Adding straw-man API for discussion

bmorris3:

PR #7, opened 2015-06-05
Added plotting API strawman.

jberlanga:
@eteq @adrn  I'll post code for doing the actual plotting in another PR.

### DIFF
--- a/dev/plots_api_strawman.py
+++ b/dev/plots_api_strawman.py
@@ -1,0 +1,121 @@
+#!/usr/bin/python
+
+# ======================================
+# Plots for Observation Planning Toolbox
+# ======================================
+
+# This is a strawman API for plotting with Astroplan.
+
+# Must first construct an Observer object...
+
+from astroplan import Observer
+import astropy.units as u
+import pytz
+
+obs = Observer(name='Subaru Telescope',
+               longitude='-155:28:48.900',
+               latitude='+19:49:42.600',
+               elevation=4163 * u.meter,
+               pressure=0.615 * u.bar,
+               relative_humidity=0.11,
+               temperature=0 * u.deg_C,
+               timezone=pytz.timezone('US/Hawaii'),
+               description="Subaru Telescope on Mauna Kea, Hawaii")
+
+day = obs.get_date("2015-05-01")
+
+# ...then define your target... 
+
+from astroplan import FixedTarget
+
+t1 = FixedTarget(name='Polaris', ra='02:31:49.09', dec='+89:15:50.8')
+
+# ...once you've constructed your Observer and Target objects, you can plot any
+#   any number of quantities.
+
+
+# ===============
+# Airmass vs Time
+# ===============
+
+from astroplan import plot_airmass
+
+# The default airmass plot takes a Target, an Observer, and a datetime object
+#   as arguments. 
+# The default time range is six hours before to six hours after the datetime
+#   specified, and a light background.
+plot_airmasss(t1, observer=obs, datetime=day)
+
+# You can also pass several targets to plot_airmass once you create them.
+
+target_list = [t1, t2, t3]
+
+plot_airmass(target_list, observer=obs, datetime=day)
+
+# You can set several options, such as change the time start/end (relative
+#   to the datetime object), set a dark background, etc., with optional keywords.
+
+plot_airmass(target_list, observer=obs, datetime=day, time_window=(-8, +8), dark)
+
+# You can leave the start or end time to the default simply by leaving that 
+#   option blank.
+
+plot_airmass(target_list, observer=obs, datetime=day, time_window=(-8,), dark)
+
+
+# =========================
+# Parallactic Angle vs Time
+# =========================
+
+from astroplan import plot_pang
+
+# The default parallactic angle plot takes one or more Target objects, 
+#   an Observer, and a datetime object.
+# The defaults are: time range is six hours before to six hours after the datetime 
+#   specified, a light background.
+# You can set options the same way you set them for plot_airmass.
+
+plot_pang(target_list, observer=obs, datetime=day)
+
+
+# =========
+# Sky Chart
+# =========
+
+from astroplan import plot_sky
+
+# The default sky chart plot takes one or more Target objects, an Observer, 
+#   and a datetime object. 
+# The default uses alt/az coordinates, is centered on the zenith 
+#   of the Observer at the datetime given, and has a light background.
+
+plot_sky(target_list, observer=obs, datetime=day)
+
+# You can set several options, including the use of RA/Dec coordinates, 
+#   a dark background, turn markers on, etc.
+# Markers are pulled from the targets provided. 
+
+plot_sky(target_list, observer=obs, datetime=day, radec, dark, markers)
+
+# You can show how your target objects move over time (with respect to your 
+#   zenith) in one of two ways:
+
+# 1) Provide multiple datetime objects after creating them:
+
+datetime_list = [time1, time2, time3, time4]
+plot_sky(target_list, observer=obs, datetime=datetime_list)
+
+# 2) Provide a time window and number of "snapshots" you want displayed. 
+#   Here, 8 hours before to 8 hours after the datetime object, 
+#   and 5 snapshots evenly spaced throughout the time window (in addition to 
+#   the datetime specified).
+
+plot_sky(target_list, observer=obs, datetime=day, time_window=(-8, +8, 5)
+
+
+# ===========
+# Other plots
+# ===========
+
+# Will be added as come up in discussion.
+
--- a/dev/plots_api_strawman.py
+++ b/dev/plots_api_strawman.py
@@ -6,32 +6,57 @@
 
 # This is a strawman API for plotting with Astroplan.
 
-# Must first construct an Observer object...
+# Must first construct an Observer object using astropy.coordinates.
 
 from astroplan import Observer
+from astropy.coordinates import EarthLocation
 import astropy.units as u
-import pytz
+from pytz import timezone
 
-obs = Observer(name='Subaru Telescope',
-               longitude='-155:28:48.900',
-               latitude='+19:49:42.600',
-               elevation=4163 * u.meter,
-               pressure=0.615 * u.bar,
-               relative_humidity=0.11,
-               temperature=0 * u.deg_C,
-               timezone=pytz.timezone('US/Hawaii'),
-               description="Subaru Telescope on Mauna Kea, Hawaii")
+my_location = EarthLocation.from_geodetic(
+    '-155d28m48.900s',
+    '+19d49m42.600s',
+    4163 * u.m,
+    )
 
-day = obs.get_date("2015-05-01")
+obs = Observer(
+    name='Subaru Telescope',
+    location=my_location,
+    pressure=0.615 * u.bar,
+    relative_humidity=0.11,
+    temperature=0 * u.deg_C,
+    timezone=timezone('US/Hawaii'),
+    description="Subaru Telescope on Mauna Kea, Hawaii"
+    )
 
-# ...then define your target... 
+# Or call up one of of the sites in Astroplan's database.
+
+from astroplan import sites
+
+obs = sites.Keck1
+
+# Then, define your target.
 
 from astroplan import FixedTarget
+from astroplan.coordinates import SkyCoord
 
-t1 = FixedTarget(name='Polaris', ra='02:31:49.09', dec='+89:15:50.8')
+my_target = FixedTarget(
+    name='Polaris',
+    coord=SkyCoord('02h31m49.09s', '+89d15m50.8s', frame='icrs')
+    )
+
+# You can also define an array with multiple targets.
+targets = [
+    FixedTarget(name='Polaris',
+                coord=SkyCoord('02h31m49.09s', '+89d15m50.8s', frame='icrs')),
+    FixedTarget(name='Sirius',
+                coord=SkyCoord('06h45m08.9173s', '-16d42m58.017s', frame='icrs')),
+    FixedTarget(name='Other Target',
+                coord=SkyCoord('01h30m0s', '+10d15m20s', frame='icrs')),
+    ]
 
 # ...once you've constructed your Observer and Target objects, you can plot any
-#   any number of quantities.
+#   any number of quantities using the plotting functions and Time objects.
 
 
 # ===============
@@ -39,28 +64,56 @@ t1 = FixedTarget(name='Polaris', ra='02:31:49.09', dec='+89:15:50.8')
 # ===============
 
 from astroplan import plot_airmass
+from astropy.time import Time
+import numpy as np
 
-# The default airmass plot takes a Target, an Observer, and a datetime object
-#   as arguments. 
-# The default time range is six hours before to six hours after the datetime
-#   specified, and a light background.
-plot_airmasss(t1, observer=obs, datetime=day)
+# plot_airmass takes an individual Target, an individual Observer,
+#   and a Time object as arguments.
 
-# You can also pass several targets to plot_airmass once you create them.
+# Construct the Time object with as many individual times as you want data
+#   points in the resulting airmass plot.
 
-target_list = [t1, t2, t3]
+# One way to do this is with a TimeDelta object and numpy.linspace().
 
-plot_airmass(target_list, observer=obs, datetime=day)
+start_time = Time('2015-01-15T21:00:00')
+end_time = Time('2015-01-16T04:00:00')
+delta_t_airmass = end_time - start_time
+times_airmass = start_time + delta_t_airmass*np.linspace(0.0, 1.0, 100)
 
-# You can set several options, such as change the time start/end (relative
-#   to the datetime object), set a dark background, etc., with optional keywords.
+plot_airmass(my_target, observer=obs, time=times_airmass)
 
-plot_airmass(target_list, observer=obs, datetime=day, time_window=(-8, +8), dark)
+# If you want to populate a Time object starting with just one time,
+# you can use astropy.units.
 
-# You can leave the start or end time to the default simply by leaving that 
-#   option blank.
+times_airmass = start_time + np.linspace(-8, 8, 100)*u.hour
 
-plot_airmass(target_list, observer=obs, datetime=day, time_window=(-8,), dark)
+# If you want to plot airmass for multiple targets on the same plot,
+#   use the new_plot=False option for each target after the first
+#   (the default is True).
+
+plot_airmass(targets[0], observer=obs, time=times_airmass)
+plot_airmass(targets[1], observer=obs, time=times_airmass, new_plot=False)
+plot_airmass(targets[2], observer=obs, time=times_airmass, new_plot=False)
+
+# If you want to plot airmass for multiple targets on different plots,
+#   simply issue the command as normal.
+
+for i in range(0, len(targets)):
+    plot_airmass(targets[i], observer=obs, time=times_airmass)
+
+# You can also set a dark background with the dark_plot=True option.
+
+plot_airmass(my_target, observer=obs, time=times_airmass, dark_plot=True)
+
+# For custom line styles/colors, use the style_kwargs option in the following
+#   manner.
+
+plot_airmass(
+    my_target,
+    observer=obs,
+    time=times_airmass,
+    style_kwarg={'linestyle': '--', 'color': 'r'},
+    )
 
 
 # =========================
@@ -69,13 +122,46 @@ plot_airmass(target_list, observer=obs, datetime=day, time_window=(-8,), dark)
 
 from astroplan import plot_pang
 
-# The default parallactic angle plot takes one or more Target objects, 
-#   an Observer, and a datetime object.
-# The defaults are: time range is six hours before to six hours after the datetime 
-#   specified, a light background.
-# You can set options the same way you set them for plot_airmass.
+# The plot_pang takes an individual Target, an individual Observer,
+#   and a Time datetime object.
 
-plot_pang(target_list, observer=obs, datetime=day)
+# Just as with plot_airmass, construct the Time object with as many
+#   individual times as you want data points in the resulting plot.
+
+start_time = Time('2015-01-15T21:00:00')
+end_time = Time('2015-01-16T04:00:00')
+delta_t_pang = end_time - start_time
+times_pang = start_time + delta_t_pang*np.linspace(0.0, 1.0, 50)
+
+plot_pang(my_target, observer=obs, time=times_pang)
+
+# If you want to plot airmass for multiple targets on the same plot,
+#   use the new_plot=False option for each target after the first
+#   (the default is True).
+
+plot_pang(targets[0], observer=obs, time=times_pang)
+plot_pang(targets[1], observer=obs, time=times_pang, new_plot=False)
+plot_pang(targets[2], observer=obs, time=times_pang, new_plot=False)
+
+# If you want to plot airmass for multiple targets on different plots,
+#   simply issue the command as normal.
+
+for i in range(0, len(targets)):
+    plot_pang(targets[i], observer=obs, time=times_pang)
+
+# You can also set a dark background with the dark_plot=True option.
+
+plot_pang(my_target, observer=obs, time=times_pang, dark_plot=True)
+
+# For custom line styles/colors, use the style_kwargs option in the following
+#   manner.
+
+plot_pang(
+    my_target,
+    observer=obs,
+    time=times_pang,
+    style_kwarg={'linestyle': '--', 'color': 'r'},
+    )
 
 
 # =========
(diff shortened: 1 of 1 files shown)


dev/plots_api_strawman.py
```diff
@@ -0,0 +1,121 @@
...
+
+# You can set several options, such as change the time start/end (relative
+#   to the datetime object), set a dark background, etc., with optional keywords.
+
+plot_airmass(target_list, observer=obs, datetime=day, time_window=(-8, +8), dark)
+
+# You can leave the start or end time to the default simply by leaving that 
+#   option blank.
+
+plot_airmass(target_list, observer=obs, datetime=day, time_window=(-8,), dark)
```
adrn: This will fail because dark is a non-keyword argument passed in after keyword arugments -- bu what do you imagine this variable `dark` is? Perhaps we want to instead allow for setting a global "dark mode" as well as a keyword argument (maybe `dark_mode=True`)?

dev/plots_api_strawman.py
```diff
@@ -0,0 +1,121 @@
...
+# You can also pass several targets to plot_airmass once you create them.
+
+target_list = [t1, t2, t3]
+
+plot_airmass(target_list, observer=obs, datetime=day)
+
+# You can set several options, such as change the time start/end (relative
+#   to the datetime object), set a dark background, etc., with optional keywords.
+
+plot_airmass(target_list, observer=obs, datetime=day, time_window=(-8, +8), dark)
```
adrn: Instead of allowing a time window, do we want to instead allow specifying the times through passing in an array of datetimes using the `datetime` keyword? For example, if I wanted to plot the airmass from, say, 0 to 10 UTC time for a given day, I could pass in:

```python
utc_day + np.linspace(0, 10, 1000)*u.hour
```

dev/plots_api_strawman.py
```diff
@@ -0,0 +1,121 @@
...
+# The default uses alt/az coordinates, is centered on the zenith 
+#   of the Observer at the datetime given, and has a light background.
+
+plot_sky(target_list, observer=obs, datetime=day)
+
+# You can set several options, including the use of RA/Dec coordinates, 
+#   a dark background, turn markers on, etc.
+# Markers are pulled from the targets provided. 
+
+plot_sky(target_list, observer=obs, datetime=day, radec, dark, markers)
```
adrn: (Again, here we have some non-keyword arguments after keyword arguments, so that is invalid syntax)

adrn: Also, rather than just allowing markers, what if we had an argument called, e.g., `style_kwargs` which gets passed to matplotlib `plot` commands. For a single object, this would just be a dictionary like:

```python
plot_kwargs = {'linestyle': '--', 'color': 'k'}
```

For multiple objects, it could be a list of dictionaries? (or a dictionary of lists...)

dev/plots_api_strawman.py
```diff
@@ -0,0 +1,121 @@
...
+
+datetime_list = [time1, time2, time3, time4]
+plot_sky(target_list, observer=obs, datetime=datetime_list)
+
+# 2) Provide a time window and number of "snapshots" you want displayed. 
+#   Here, 8 hours before to 8 hours after the datetime object, 
+#   and 5 snapshots evenly spaced throughout the time window (in addition to 
+#   the datetime specified).
+
+plot_sky(target_list, observer=obs, datetime=day, time_window=(-8, +8, 5)
```
adrn: Same comment here about `time_window` -- perhaps we want to just let the user specify the time directly through `datetime=`?

jberlanga:
@bmorris3 @cdeil @ejeschke  Sorry to not have included you all before, but of course I would like to know what you guys think. :)

dev/plots_api_strawman.py
jberlanga: @adrn  I think here it does make more sense to do as you say and simply feed a list/array of times into datetime (as mentioned in [Lines 103-106](https://github.com/astroplanners/astroplan/pull/7/files#diff-5b6ad4137e779466ca6135ac57c29709R103)--if the user wants evenly spaced snapshots, they could just construct an array in that manner.

dev/plots_api_strawman.py
jberlanga: @adrn  I can definitely see that being useful to users who want finer control.  For those who don't care as much or just want a quick and dirty plot, markers distinguished by some default styles would do the job.

dev/plots_api_strawman.py
jberlanga: @adrn  Here I think it would be good to get input from everyone else, but I'm inclined to think that users wouldn't be as interested in specifying the frequency of points (or smoothness of the line plotted) as they would be to simply set the time window--unless they are somehow tying this to generating tables and want the same precision in both.  

@astroplanners/astroplanners-contributors @astroplanners/owners Any thoughts?

dev/plots_api_strawman.py
jberlanga: @adrn  I was indeed imagining `dark` to set a dark mode, and thought perhaps users would find it easier/quicker to type that in instead of adding yet another `variable=value`, since there are really just two options, light (by default) or dark.  But maybe the meaning gets lost with just `dark`?

cdeil:
I've never figured out how to write plotting functions that can be re-used / extended / adapted by users to do the more fancy plots or layout they like.

The best pattern I'm aware of and what I think matplotlib is also using is for every plotting function to take an [Axes](http://matplotlib.org/api/axes_api.html#matplotlib.axes.Axes) argument:
```python
def plot_something(..., ax=None):
    import matplotlib.pyplot as plt
    if ax is None:
        ax = plt.gca()
   # Now start plotting
```
This allows users to control where they plot their stuff.

Another question is how users can configure plotting layout, i.e. colors, labels, ...

I guess in general there's several approaches:
1. Let the user pass in `ax` and plotting options
2. Return `ax` and document how the user can change colors, labels, ... after calling the pre-made plotting function.
3. Do something more complex where we define classes (like `aplpy.FITSFigure` did) to produce plots.

adrn:
@cdeil for my own code (not saying this is right, but one example) I usually do 1), so, something like this:

```python
def plot_stuff(..., ax=None, style_kwargs=dict()):
    if axes is None:
        fig,ax = plt.subplots()
    # ...
    ax.plot(xx, **style_kwargs)
```

cdeil:
:+1: to @adrn's suggestion.

Some comments / questions:
* Should be `style_kwargs=None` and then `if style_kwargs is None: style_kwargs = dict()` because in Python mutable objects like dicts should never be used as keyword argument default values, right?
* What do you return from plotting functions? `fig` or `ax` or both or nothing?
* Do we have functions here that make more than one matplotlib plotting call? If so, do we have one `style_kwargs` that's passed on to all, or separate ones for the separate calls?

adrn:
@cdeil 

1. yep! default to None makes sense
2. I usually return `fig,ax` but am open to other suggestions :)
3. I think the functions we have here only `plot()` once? But worth thinking about for other functions...

cdeil:
:+1: to @adrn's suggestion (unless someone comes with a better suggestion or there's a more common pattern for building plotting functions on top of matplotlib).

bmorris3:
I second the :+1: to @adrn 's suggestion.

dev/plots_api_strawman.py
```diff
@@ -0,0 +1,121 @@
...
+# Airmass vs Time
+# ===============
+
+from astroplan import plot_airmass
+
+# The default airmass plot takes a Target, an Observer, and a datetime object
+#   as arguments. 
+# The default time range is six hours before to six hours after the datetime
+#   specified, and a light background.
+plot_airmasss(t1, observer=obs, datetime=day)
```
bmorris3: `airmasss` -> `airmass`. Little typo. :smile:

dev/plots_api_strawman.py
bmorris3: On the topic of a time window: sometimes when targets pass directly over the zenith, an altitude vs. time plot will have a corner at 90 degrees, which you might want to resolve with precision on one of these plots (since many telescopes can't point directly at the zenith). Maybe it would be nice to have a time resolution option so when people like me inevitably debug to figure out why there's a corner in the plot at 90 degrees, they have a way of making sure they know what they're looking at.

Also: What is the `dark` parameter setting? Is that a twilight?

cdeil:
@adrn , all - Please have a look [here](http://nbviewer.ipython.org/gist/cdeil/6a2b33715fe47408587c). Is this the plot function pattern we want to follow in `astroplan`?

Specifically I only pass in and return `ax` and call `plt.gca()`.
@adrn, I think you said you usually call `plt.subplot()` and also return `fig`. Does this allow any use cases that aren't possible with calling `gca()` and only returning `ax`? If no, I'd prefer the slightly simpler option and only return `ax`.

adrn:
@cdeil yea, that looks good to me! @eteq convinced me the other day that calling `subplots()` is a bad idea and `gca()` is a better idea (see my modified comment above :))

dev/plots_api_strawman.py
jberlanga: @bmorris3  I've been looking at the `astropy.time` documentation, and it seems that users can set precision to some extent by directly using the [`astropy.time` machinery](http://astropy.readthedocs.org/en/latest/time/#creating-a-time-object) directly.  So my initial thought for a solution to your problem (assuming that, instead of using time window options, we allow the user to build a `Time` object with as many times as they like), is to explicitly point the user to this option.  

`dark` was meant to be a keyword turning on dark plotting, but @adrn has convinced me that something like `dark_plot=False` as the default setting would be better.

jberlanga:

I've updated the rudimentary examples in [jberlanga/plotting_prep](https://github.com/jberlanga/plotting_prep) for airmass/pa/skyplot.

jberlanga:

I've updated the plotting API strawman. See [2nd Commit](https://github.com/jberlanga/astroplan/commit/dc944d4d85dd50de8f1df15fb122dd8868f83f0a).

dev/plots_api_strawman.py
```diff
@@ -0,0 +1,242 @@
...
+# This is a strawman API for plotting with Astroplan.
+
+# Must first construct an Observer object using astropy.coordinates.
+
+from astroplan import Observer
+from astropy.coordinates import EarthLocation
+import astropy.units as u
+from pytz import timezone
+
+my_location = EarthLocation.from_geodetic(
```
eteq: Very minor point, but I would think it might make more sense to include this below in the ``Observer`` call because it's obviously associated with Subaru.  Or if you really like it separate, I would change ``my_location`` to something like ``subaru_location`` to make it more obvious it's really meant for use there.

Also, to make it slightly more explicit you can do ``EarthLocation(lon='-155d28m48.900s', lat='+19d49m42.600s',height=4163*u.m)

(These are all mostly personal opinion, though, so if you genuinely think it's better this way  that's ok too!)

dev/plots_api_strawman.py
```diff
@@ -0,0 +1,242 @@
...
+
+# Construct the Time object with as many individual times as you want data
+#   points in the resulting airmass plot.
+
+# One way to do this is with a TimeDelta object and numpy.linspace().
+
+start_time = Time('2015-01-15T21:00:00')
+end_time = Time('2015-01-16T04:00:00')
+delta_t_airmass = end_time - start_time
+times_airmass = start_time + delta_t_airmass*np.linspace(0.0, 1.0, 100)
```
eteq: Ohh, this is a nice trick, I'm glad you're showing it here!

dev/plots_api_strawman.py
```diff
@@ -0,0 +1,242 @@
...
+#   (the default is True).
+
+plot_airmass(targets[0], observer=obs, time=times_airmass)
+plot_airmass(targets[1], observer=obs, time=times_airmass, new_plot=False)
+plot_airmass(targets[2], observer=obs, time=times_airmass, new_plot=False)
+
+# If you want to plot airmass for multiple targets on different plots,
+#   simply issue the command as normal.
+
+for i in range(0, len(targets)):
```
eteq: Note that there's a much easier way to do this:
```
for targ in targets:
    plot_airmass(targ, ...)
```
In python you can usually get away with no indecies in for loops - instead you just loop over lists like what I've shown here, or if you want matched indecies from multiple lists, you can usually use the ``zip`` function.  (Once in a while you have to use an index when you're doing something complicated, but it's generally the exception rather than the rule).

dev/plots_api_strawman.py
```diff
@@ -0,0 +1,242 @@
...
+plot_airmass(my_target, observer=obs, time=times_airmass, dark_plot=True)
+
+# For custom line styles/colors, use the style_kwargs option in the following
+#   manner.
+
+plot_airmass(
+    my_target,
+    observer=obs,
+    time=times_airmass,
+    style_kwarg={'linestyle': '--', 'color': 'r'},
```
eteq: is there a specific reason not to just pass in all kwargs, rather than having a specific ``style_kwargs`` dictionary?  That is, something like:
```
plot_airmass(
    my_target,
    observer=obs,
    time=times_airmass,
    linestyle='--',
    color='r')
```

dev/plots_api_strawman.py
```diff
@@ -0,0 +1,242 @@
...
+    )
+
+
+# =========================
+# Parallactic Angle vs Time
+# =========================
+
+from astroplan import plot_pang
```
eteq: I think `pang` is an abbreviation that isn't immediately obvious.  Why not ``plot_parallactic_angle``, ``plot_parallactic``, or ``plot_parr_ang``?  Those are much more apparent at a glance.

eteq:
@jberlanga - a great start! Some general comments below.

First, just wanted to make sure I understand something:  the actual plots are being shown in ``plotting_prep``, is that right?  Maybe it would be useful to instead put the notebooks there in this PR?  That way we can comment on them all at the same time.  Or, would you rather we just comment in this PR based on the visual appearence of the plots from ``plotting_prep``? (I may have some more comments on the actual appearence of the plots once I know the answer to that.)

Right now it looks like you're expecting the user to give every time they want plotted.  A very common use case I expect for this is to want to plot it for the whole "night" (where "night" might include the daytime if you're a radio astronomer...).  So maybe you should add to this that, if the user gives a *scalar* time, then you plot that `Time +/- 12 hours? (i.e., a 24 hour window centered on that time) with a default sampling of, say 100 points? (Note that this is *not* the same thing as an length-1 array time, which would still allow plotting a single point.)

I'm :-1: on the ``new_plot`` keyword.  This rolls into the discussion @cdeil and @adrn were having above: instead of ``new_plot`` or something like that, the idea is that you have an ``ax`` argument to all of the functions, which the user can use to pass in an axes.   At the very top of the plotting functions, if the user passes in ``None`` you use ``plt.gca`` to get the axes object like this:
```
if ax is None:
    ax = plt.gca()
ax.scatter(...)
```
That is, you do *all* plotting with the ``ax`` object (which you then return at the end of the function in case the user wants to do something with it).  So then if the user wants a new plot, they could do``ax.cla()`` to replace a particular subplot, or ``plt.clf()`` to clear the current figure before running the plot function.  The philosophy here is that there "should be only one obvious way to do it", and matplotlib already provides ``clf`` and ``cla`` and similar to clear a plot.

dev/plots_api_strawman.py
cdeil: +1 to `plot_parallactic_angle` ... even for `plot_parr_ang` I don't know if I'd be able to guess what it is from the function name.

dev/plots_api_strawman.py
cdeil: This is also a question I had when @adrn proposed to have a `style_kwarg` argument.
One pro for `style_kwargs` is that if the user mistypes one of the other keyword arguments she gets a useful error instead of it being added to `**kwargs`. Another one is that if there are two matplotlib plot calls in a function, they could be styled differently if we want that. The pro for `**kwargs` is that it's shorter to type. I'm pretty much 50:50 on this, maybe slightly leaning towards `style_kwargs` (it should be with the plural `s`).

@adrn, all – Which do you prefer?

cdeil:
+1 to @eteq's comment. Remove the `new_plot` option and pass `ax` around instead.

dev/plots_api_strawman.py
jberlanga: @eteq  Since I haven't messed around with `EarthLocation` a whole lot, I was going off of the example in the latest iteration to PR #4.  I'm not married to any of the syntax here, although :+1:  to `subaru_location = EarthLocation(lon='-155d28m48.900s', lat='+19d49m42.600s',height=4163*u.m)`.

dev/plots_api_strawman.py
jberlanga: @eteq  Thanks for the tip!

dev/plots_api_strawman.py
jberlanga: @astroplanners/astroplanners-contributors 

@cdeil  I agree with you that since you can have lines or markers for airmass/parallactic angle, but lines don't make sense for a skychart, an error telling the user that (and why) their style argument won't work is useful.  

I'm not sure what you mean by two matplotlib calls in a function (e.g., you call plot_airmass once and you get two different sets of data on the same set of axes?).  @adrn, @eteq and I have previously discussed calling a plot function once for each set of data (e.g., airmass for multiple targets) on the same plot.

dev/plots_api_strawman.py
jberlanga: :+1:  to `plot_parallactic` or `plot_parr_ang`.  I think `plot_parallactic_angle` is a bit long.

dev/plots_api_strawman.py
```diff
@@ -0,0 +1,242 @@
...
+
+# The plot_pang takes an individual Target, an individual Observer,
+#   and a Time datetime object.
```
jberlanga: Replace `Time datetime object` with `Time object`.

jberlanga:
@astroplanners/astroplanners-contributors  @eteq  

So to clarify on the purpose of this PR vs. the purpose of the [plotting_prep notebooks](https://github.com/jberlanga/plotting_prep):  
1. This PR is primarily meant to illustrate the usage of the plotting methods.
2. The notebooks started out as my attempt to put together some basic plots for each category, and I shared them with everyone (not just @adrn & @eteq) because I thought it could facilitate discussion on the look/basic construction.  

>Right now it looks like you're expecting the user to give every time they want plotted. A very common use >case I expect for this is to want to plot it for the whole "night" (where "night" might include the daytime if >you're a radio astronomer...). So maybe you should add to this that, if the user gives a scalar time, then you >plot that `Time +/- 12 hours? (i.e., a 24 hour window centered on that time) with a default sampling of, say >100 points? (Note that this is not the same thing as an length-1 array time, which would still allow plotting a >single point.)

I did include something like this in my original iteration, but 
1. after [some](https://github.com/astroplanners/astroplan/pull/7#discussion_r32169434) [discussion](https://github.com/astroplanners/astroplan/pull/7#discussion_r32168995) with @adrn and 
2. seeing that you could do something like `times_airmass = start_time + np.linspace(-8, 8, 100)*u.hour` if you wanted to specify a time window centered on one particular time (as opposed to specifying the beginning/end of the window and using a `TimeDelta` object), 
it seemed to me that we might end up with a more flexible method, and more consistency in usage among the different plotting methods.  

I'm okay with nixing the `new_plot` option, but I'm going to need more direction on how you end up with a finished plot by returning an `ax` (as opposed to a finished plot).

jberlanga:

For the lazy, here is what my rudimentary plots look like at the moment.

airmass_draft

parallactic_draft

skychart_sample

bmorris3:
Thanks for the demo! Should we have the airmass plot's upper limit be X~3? Most people don't observe above that, and `sec(z)` becomes a poor approximation above X~2.

ejeschke:
Hi Jazmin,

Thanks for catering to the lazy!

The plots look pretty good.  I'd agree with the others that giving direct
access to the axes (and figure) of the matplotlib object is the most
straightforward way to let them customize the plots.

How about adding a moon altitude plot?  As I showed in my early example,
this can be done by adding an elevation scale on the right (although I'm
not sure it is the best approach--perhaps making a separate plot just for
this would be useful).

I am also wondering about a moon separation from target plot.  With one of
the instruments we are hoping to use this with it has a large (1.5 deg)
FOV.  This makes it susceptible to ghosting from the moon.  We need to
maintain a 30 deg separation from the moon ideally for each target.  I'm
wondering about a plot that would show each target's separation over the
course of the night.

Regards,
~E

On Tue, Jun 23, 2015 at 6:38 AM, Jazmin Berlanga Medina <
notifications@github.com> wrote:

> @astroplanners/astroplanners-contributors
> <https://github.com/orgs/astroplanners/teams/astroplanners-contributors>
>
> For the lazy, here is what my rudimentary plots look like at the moment.
>
> [image: airmass_draft]
> <https://cloud.githubusercontent.com/assets/11530948/8311508/386d6d50-199c-11e5-8efe-60bf4df1614c.png>
>
> [image: parallactic_draft]
> <https://cloud.githubusercontent.com/assets/11530948/8311510/3f308da2-199c-11e5-9067-48d22347ca4e.png>
>
> [image: skychart_sample]
> <https://cloud.githubusercontent.com/assets/11530948/8311515/43a893c0-199c-11e5-9415-0566c1ce1dd9.png>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/astroplanners/astroplan/pull/7#issuecomment-114565934>
> .
>



-- 
Eric Jeschke

ejeschke:
Also, a line plot of the moon's track over the sky (polar) chart would be
great.

~E

On Tue, Jun 23, 2015 at 8:39 AM, Eric Jeschke <eric@redskiesatnight.com>
wrote:

> Hi Jazmin,
>
> Thanks for catering to the lazy!
>
> The plots look pretty good.  I'd agree with the others that giving direct
> access to the axes (and figure) of the matplotlib object is the most
> straightforward way to let them customize the plots.
>
> How about adding a moon altitude plot?  As I showed in my early example,
> this can be done by adding an elevation scale on the right (although I'm
> not sure it is the best approach--perhaps making a separate plot just for
> this would be useful).
>
> I am also wondering about a moon separation from target plot.  With one of
> the instruments we are hoping to use this with it has a large (1.5 deg)
> FOV.  This makes it susceptible to ghosting from the moon.  We need to
> maintain a 30 deg separation from the moon ideally for each target.  I'm
> wondering about a plot that would show each target's separation over the
> course of the night.
>
> Regards,
> ~E
>
> On Tue, Jun 23, 2015 at 6:38 AM, Jazmin Berlanga Medina <
> notifications@github.com> wrote:
>
>> @astroplanners/astroplanners-contributors
>> <https://github.com/orgs/astroplanners/teams/astroplanners-contributors>
>>
>> For the lazy, here is what my rudimentary plots look like at the moment.
>>
>> [image: airmass_draft]
>> <https://cloud.githubusercontent.com/assets/11530948/8311508/386d6d50-199c-11e5-8efe-60bf4df1614c.png>
>>
>> [image: parallactic_draft]
>> <https://cloud.githubusercontent.com/assets/11530948/8311510/3f308da2-199c-11e5-9067-48d22347ca4e.png>
>>
>> [image: skychart_sample]
>> <https://cloud.githubusercontent.com/assets/11530948/8311515/43a893c0-199c-11e5-9415-0566c1ce1dd9.png>
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/astroplanners/astroplan/pull/7#issuecomment-114565934>
>> .
>>
>
>
>
> --
> Eric Jeschke
>
>


-- 
Eric Jeschke

dev/plots_api_strawman.py
eteq: @cdeil - re: the error question: it's ok with ``**kwargs`` because they get sent directly to matplotlib which would give errors for invalid keywords.  You're right it's a little less clear because it would come from `matplotlib` instead of the `astorplan.plot_*` function, though.

As for multiple plotting functions, @jberlanga, @cdeil means that if there are *two* calls to matplotlib inside a `astroplan.plot_*` function, using ``**kwarg`` you can only unambiguously send the keywords off to one of them.  For example if we had a `plot_airmass_and_moon` plot, you'd need to make *two* calls to ``scatter`` or ``plot``, and if the style stuff is in a dictionary, you can just make two dictionaries: ``airmass_style_kwargs`` and ``moon_style_kwargs``.  But if the keywords are forwarded directly using ``**kwargs``, it can only be done for one of them. I agree that *might* be an issue, but it might be that we agree now to not do more than one plot per function call?  (I.e., require the user to do ``plot_airmass`` and ``plot_moon`` as separate functions).

dev/plots_api_strawman.py
eteq: I confess I like ``plot_parr_ang`` because I have trouble with the world parallactic: I always end up spelling it parrallactic for some reason.  But @cdeil has a point about ``parr_ang`` being a bit confusing.  What do you think about ``plot_parallactic`` then, @cdeil?

cdeil: To be honest ... even with `plot_parallactic_angle` or probably any other name I'd have to look at an example plot to know what it is. So yes, any function name is fine with me.

Is it foreseen that all plot functions should have an example in the docstring?
I think that would be nice and it's easy to do (see [example](http://astropy.readthedocs.org/en/latest/api/astropy.convolution.MexicanHat1DKernel.html))

eteq:
@ejeschke - I was discussing this with @jberlanga and I agree we want those moon plots, but I think a better approach is to implement a `MoonTarget`, and then it can be passed in as a target to any of the plotting functions already here.   Similarly for `SunTarget`.  That re-uses the effort in each of these plotting functions more efficiently.

I did have one more plot to suggest, though: `plot_nighttime` (or perhaps it should be called `plot_twilight`?), which would do something like what you see in the last plot of http://docs.astropy.org/en/stable/coordinates/observing-example.html : shaded regions for night and twilight.

eteq:
Also, @jberlanga, I agree with @bmorris3 that an airmass limit of something like 3 is better because people rarely observe even past 3.  Of course they can always change it manually, so this is just a "sensible default".  Also, I personally think airmass should increase as it goes up, rather than down like you have it in that plot (but that's not necessarily standard beyond "standard for Erik".)

One other change I'd like to suggest in the plots themselves: for airmass and parallactic angle, I suggest using `plot`, rather than `scatter`.  For those two, the discrete points don't really make sense - they're not collections of x-y points, but rather a plot of something(s) over time, which is exactly what `plot` is for.  The user can always pass in a `fmt` of `o` if they really want it to look like what you've got here.

For the sky chart it's less clear to me if it should be `scatter` or `plot`.  scatter makes sense for a *single* instant in time, but `plot` makes more sense if there's multiple times.  But that might just be me...

ejeschke:
@eteq, that makes sense (the MoonTarget, SunTarget).  I think I did something similar in the obsplan example.  Shall we also have the planetary bodies easily accessible (Jupiter, Venus, etc.).?

ejeschke:
+1 to setting airmass limit of 3 as default.  I also prefer a plot with lines rather than a scatter plot.

One nice option is to plot the current time "window" in the airmass plot (e.g. as an axvspan from the current time to, say, one hour later).  Then if you are generating the plot during the night you can zero in more easily on what your targets are doing in the next hour.  Of course the user can add this, given access to the axes, but since it might be frequently used it might be nice to have as an option to the plot.

jberlanga:
Closing pull request.

PR #11, opened 2015-06-19
Beginning in the FixedTarget and Observer classes

bmorris3:


astroplan/core.py
```diff
@@ -25,8 +40,10 @@ class Observer(object):
     Some comments.
     """
 
-    def __init__(self, name, longitude, latitude, elevation, pressure,
-                 relative_humidity, temperature, timezone, description=None):
+    def __init__(self, name=None, location=None, latitude=None, longitude=None,
```
bmorris3: Should we instead use the `**kwargs` syntax and then `kwargs.get` or `kwargs.pop` the keywords one at a time?

astroplan/core.py
```diff
@@ -37,70 +54,94 @@ def __init__(self, name, longitude, latitude, elevation, pressure,
...
 
+        time : `~astropy.time.Time`
+            Object
         """
-        raise NotImplementedError()
+        if not isinstance(target, FixedTarget):
+            raise TypeError, ('The target must be an instance of FixedTarget')
+
+        return target.coord.transform_to(AltAz(target.coord.ra, target.coord.dec,
+                                         location=self.location, obstime=time))
```
bmorris3: Here I assume the `target` input to `obs.altaz(target, time)` is an instance of `FixedTarget`, but it could be a moving target in the future, or we may choose to support SkyCoords. Any opinions on interchangeability with SkyCoords where feasible?

astroplan/core.py
```diff
@@ -2,6 +2,21 @@
...
 
+from astropy.coordinates import (EarthLocation, Latitude, Longitude, SkyCoord,
+                                 AltAz)
+import astropy.units as u
+from astropy.units.quantity import Quantity
+
+import pytz
+
+################################################################################
+# Temporary solution to IERS tables problems
```
cdeil: Add a TODO marker, or even better yet, make a Github issue that this should be changed / removed once X in Astropy happens. @bmorris3 – You probably know what we are waiting for and what should be done, but who knows, maybe it'll take a year to be resolved, and then the next person working on astroplan doesn't. (And it can be nice for yourselt too, to have a list of issues that can be addressed one by one instead of having to remember things.)

astroplan/core.py
```diff
@@ -310,7 +355,20 @@ class FixedTarget(Target):
     """
     An object that is "fixed" with respect to the celestial sphere.
     """
-
+    def __init__(self, coord, name=None, **kwargs):
```
bmorris3: I went to the @adrn [-style extreme](https://github.com/astroplanners/astroplan/pull/4#discussion_r32672422) here and started by initializing a `FixedTarget` with only a `SkyCoord`, because I'm starting to agree that this is the clearest, least redundant way to write the code.

astroplan/core.py
```diff
@@ -37,70 +54,94 @@ def __init__(self, name, longitude, latitude, elevation, pressure,
...
-        ----------
-        pressure : `~astropy.units.Quantity`
-            The ambient pressure.
+            self.location = EarthLocation.from_geodetic(longitude, latitude,
+                                                        elevation)
 
-        relative_humidity : `~astropy.units.Quantity`
-            The ambient relative humidity.
+        elif isinstance(location, EarthLocation):
+            self.location = location
```
bmorris3: This is my compromise between @ejeschke 's accepting attitude towards input formats and @adrn 's [preference for accepting one type (`EarthLocation`)]([-style extreme](https://github.com/astroplanners/astroplan/pull/4#discussion_r32672422) ). I accept latitudes and longitudes or an `EarthLocation`. If you input latitude/longitude, they become an EarthLocation under the hood.

cdeil:
@bmorris3 – When you make pull requests, we get a notification, but we don't know whether you want us to review them at this point or not. In Astropy and generally on Github, people do this differently, you could:
* Say whether this is ready for review or not in the description.
* If it's work in progress (usually is when the PR is first made) and we shouldn't review yet, add `WIP:` to the title or add a task list (https://help.github.com/articles/writing-on-github/#task-lists) of things you still want to do in this PR.

Another important thing to know (which you probably do) is that Github doesn't send notifications when you push new commits to a branch. So if there has been review, and then you've addressed the comments, and want something to be merged or a second round of review, you have to leave a comment  saying so.

astroplan/core.py
```diff
@@ -37,70 +54,94 @@ def __init__(self, name, longitude, latitude, elevation, pressure,
...
 
-        longitude : str or `~astropy.units.quantity`?
-            The longitude of the observing location.
+        pressure : `~astropy.units.Quantity`
+            The ambient pressure.
+
+        location : `~astropy.coordinates.EarthLocation`
+            The location (latitude, longitude, elevation) of the observatory.
+
+        longitude : str or `~astropy.units.quantity`
```
cdeil: `quantity` needs to be `Quantity` or the Sphinx link won't work.
You probably know this, but just in case ... the commands to check the docs locally are:
```
python setup.py build_sphinx
open _docs/_build/html/index.html
```

astroplan/core.py
bmorris3: :+1: Got it.

astroplan/core.py
bmorris3: Gah, I should have caught that!

eteq:
@bmorris3 - something else is confusing me about this PR: it appears to have some (all?) of the commits from #4.  Did you mean to do that?  I'm guessing not, because it looks like this is only supposed to include some changes in ``core.py``, and you have a commit where you *removed* the API doc.  This doesn't actually remove it from history, and if you merge this after the API is merged, this will *delete* that document, which I don't think is what you want.

So maybe you want to rebase this against master.
```
git checkout initclasses # you might already be on this branch
git branch initclasses-backup # insurance in case you screw up something -- you'll  always back it all out by doing git reset --hard initclasses-backup
git fetch upstream  # this might actually be "astroplanners" depending on how your remotes are set up.  I try to avoid ever having an "origin"/"upstream" because I find it confusing, so I always use the same names as github assigns.  But YMMV.
git rebase -i upstream/master
```
this will bring up an editor showing you all the commits between master and this branch. You'll want to delete the top 3 (which are from the API branch).  Then when you close your editor it will give you an error about ``The previous cherry-pick is now empty, possibly due to conflict resolution.``  Just do ``git rebase --continue`` and everything will be fine.   Once you've done that, you'll see that this branch is just the two commits you *wanted* and nothing else.  So you can do ``git push origin initclasses -f`` and all should be good.

bmorris3:
Just to leave a note here – @eteq helped me do the rebase, they won't appear in the PR.

> Those were not the commits you were looking for.
– Obi Eteq

astroplan/core.py
```diff
@@ -2,6 +2,21 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
+from astropy.coordinates import (EarthLocation, Latitude, Longitude, SkyCoord,
```
cdeil: @bmorris3 – Do you understand this? https://travis-ci.org/astroplanners/astroplan/jobs/67568895#L372  (I don't)

bmorris3: I don't either, maybe @eteq can give us some hints on how to read these logs on Monday or in the Hangout on Tuesday.

astroplan/core.py
```diff
@@ -2,6 +2,21 @@
...
+                                 AltAz)
+import astropy.units as u
+from astropy.units.Quantity import Quantity
```
cdeil: ImportError, should be:
```python
form astropy.units import Quantity
```
You can always check locally whether most things are OK by running
```
python2 setup.py test
python3 setup.py test
```

astroplan/core.py
```diff
@@ -25,82 +40,108 @@ class Observer(object):
...
-            The ambient relative humidity.
+        elif isinstance(location, EarthLocation):
+            self.location = location
 
-        temperature :
-            The ambient temperature.
-        """
-        raise NotImplementedError()
+        else:
+            raise ValueError, ('Observatory location must be specified with '
```
cdeil: SyntaxError on Python 3:
```python
$ python setup.py test
Traceback (most recent call last):
  File "setup.py", line 37, in <module>
    __import__(PACKAGENAME)
  File "/Users/deil/code/astroplan/astroplan/__init__.py", line 18, in <module>
    from .core import *
  File "/Users/deil/code/astroplan/astroplan/core.py", line 111
    raise ValueError, ('Observatory location must be specified with '
                    ^
SyntaxError: invalid syntax
```
(there shouldn't be a comma, just
```python
raise ValueError('... description ...')
```

astroplan/core.py
cdeil: I still don't understand why that line gave an error on travis-ci:
```
ImportError: No module named astropy.coordinates
```
But if you fix the other errors mentioned below and `python setup.py test` (which imports this file) works locally, it should work on travis-ci, too.

astroplan/core.py
adrn: I think my preference is no to `kwargs` -- if the arguments are spelled out, it is easier to quickly see what is expected from the documentation of the call sequence.

eteq:
A quick note: that missing indent @bmorris3 added in the last commit at @adrn's suggestion was not a problem per se with the package template, because there's no `core` module in the package template.  (no action required, just an FYI about why it happened)

astroplan/tests/test_core.py
```diff
@@ -0,0 +1,115 @@
...
+    pyephem_vega.compute(pyephem_obs)
+    pyephem_altitude = Latitude(np.degrees(pyephem_vega.alt)*u.degree)
+    pyephem_azimuth = Longitude(np.degrees(pyephem_vega.az)*u.degree)
+
+    # Assert that altitudes/azimuths are within 30 arcsec - this is a wide
+    # tolerance because the IERS tables used by astroplan may offset astroplan's
+    # positions due to leap seconds.
+    tolerance = 30*u.arcsec
+    assert (abs(pyephem_altitude - astroplan_altitude) < tolerance and
+            abs(pyephem_azimuth - astroplan_azimuth) < tolerance)
```
bmorris3: @astroplanners/astroplanners-contributors : Leap seconds will make comparison with PyEphem tricky in our tests, since PyEphem uses [old Earth orientation standards](https://twitter.com/brandon_rhodes/status/605044030406303744) and astroplan uses up-to-date IERS tables. Any ideas on how to handle the differences in apparent positions of targets as a result? 

Here I use a very broad tolerance in the accepted difference in altitude/azimuth between astroplan and PyEphem (30 arcsec). In the above test, the differences are significantly smaller than this tolerance: 
```python
>>> (pyephem_altitude - astroplan_altitude).to('arcsec')
<Angle 0.10845603317193309 arcsec>

>>> (pyephem_azimuth - astroplan_azimuth).to('arcsec')
<Angle 7.0463692237353825 arcsec>
```

astroplan/tests/test_core.py
```diff
@@ -0,0 +1,115 @@
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+
+from astropy.coordinates import (EarthLocation, Latitude, Longitude, SkyCoord)
+import astropy.units as u
+from astropy.time import Time
+from astropy.tests.helper import remote_data
```
bmorris3: @eteq: I'm borrowing astropy's test helpers in my tests, since `SkyCoord.from_name` requires internet access, which gets used in `FixedTarget.from_name`. Is it appropriate to steal this astropy machinery like so?

cdeil: As far as I know it's common practice to use `from astropy.tests.helper import remote_data` in affiliated packages.
E.g. [gammapy](https://github.com/gammapy/gammapy/search?utf8=%E2%9C%93&q=remote_data) and [astroquery](https://github.com/astropy/astroquery/search?utf8=%E2%9C%93&q=remote_data) do this.

But `remote_data` does not show up in the API reference docs for Astropy:
http://astropy.readthedocs.org/en/latest/search.html?q=remote_data&check_keywords=yes&area=default

I think `remote_data` should be part of Astropy's public API and show up in the docs.

@eteq – What do you think?

astroplan/tests/test_core.py
cdeil: There's not much you can do. Either make the tolerance so large that it's OK (and document in a comment or somewhere that the difference is understood). Or test against another software package, e.g. does Skyfield have this functionality?

astroplan/tests/test_core.py
```diff
@@ -0,0 +1,130 @@
...
+                    latitude=lat,
+                    longitude=lon,
+                    elevation=elevation,
+                    **environment_kwargs)
+
+    obs2 = Observer(name='Observatory',
+                    location=location,
+                    **environment_kwargs)
+
+    assert obs1.location == obs2.location, ('Locations not parsed equivalently')
```
cdeil: I'm -1 on adding descriptions of assert fails.
Over time it adds up to significance work and amount of text we have to write / discuss, with IMO no benefit.
When tests fail pytest will give a very good error message (line number, values, traceback).
Usually one will have to look at the test and code anyways, and then the description on the assert adds nothing useful.

astroplan/tests/test_core.py
```diff
@@ -0,0 +1,130 @@
...
+    location = EarthLocation.from_geodetic(longitude, latitude, elevation*u.m)
+    astroplan_obs = Observer(name='Observatory', location=location,
+                             pressure=pressure*u.bar)
+    astroplan_vega = FixedTarget(vega_coords)
+    altaz = astroplan_obs.altaz(astroplan_vega, time)
+    astroplan_altitude = altaz.alt
+    astroplan_azimuth = altaz.az
+
+    # Calculate altitude/azimuth with PyEphem
+    pyephem_obs = ephem.Observer()
```
cdeil: @eteq, all – For tests against results from other packages like `pyephem`, should we compute the results using the other package on the fly or hard-code the result numbers?

Here Brett is running `pyephem` in a test.
In Astropy we don't want the `pyephem` dependency even for running tests, and we hard-code the pyephem numbers into the test:
https://github.com/astropy/astropy/blob/master/astropy/coordinates/tests/accuracy/test_altaz_icrs.py#L61

IMO the best solution is to have hardcoded numbers from other software packages (so that the tests don't have that external dependency, and when results in the external package change because they introduce or fix bugs, out tests don't break). And in addition it would be nice if the code snippets to reproduce the numbers from the external packages were available, in case someone needs to understand or debug a test in the future. I'm not sure how to do that best ... maybe commented out code in the test or test docstring? Or a separate folder in our repo with little functions one can run that print all relevant numbers to the console (slightly more complicated setup, but makes it easy to run and reproduce the numbers)?

astroplan/tests/test_core.py
```diff
@@ -0,0 +1,130 @@
...
+    pyephem_vega._dec = ephem.degrees(np.radians(vega_coords.dec.value))
+    pyephem_vega.compute(pyephem_obs)
+    pyephem_altitude = Latitude(np.degrees(pyephem_vega.alt)*u.degree)
+    pyephem_azimuth = Longitude(np.degrees(pyephem_vega.az)*u.degree)
+
+    # Assert that altitudes/azimuths are within 30 arcsec - this is a wide
+    # tolerance because the IERS tables used by astroplan may offset astroplan's
+    # positions due to leap seconds.
+    tolerance = 30*u.arcsec
+    assert (abs(pyephem_altitude - astroplan_altitude) < tolerance and
```
cdeil: @bmorris3 – Does this expression give a useful error message if the assert fails?

I **always** use [numpy.testing.assert_allclose](http://docs.scipy.org/doc/numpy-dev/reference/generated/numpy.testing.assert_allclose.html) for any float assertions and find this easier that doing `abs` and `-` and `/` myself.

Although I just tried this simple example and pytest is very clever and nice with the error reporting for `assert abs(a-b) < tol`, it does show the value of `a` and `b` and `abs(a-b)`:
https://gist.github.com/cdeil/dd091ba63b9243757e15
So if the assertion fail report is still good for the assert as written here, and you like this way better, OK to keep as is.

astroplan/tests/test_core.py
bmorris3: It's covered in the [Test Guidelines](http://astropy.readthedocs.org/en/latest/development/testguide.html#working-with-data-files) in such a way that it's almost advertising for you to borrow the method.

astroplan/tests/test_core.py
bmorris3: Maybe for now while we're waiting on consensus on this from @astroplanners/astroplanners-contributors I'll just comment out the code that calls PyEphem.

astroplan/tests/test_core.py
bmorris3: That's a really handy method @cdeil , thanks for the suggestion! I only like my way better because it's what I'm used to seeing, but it looks like [astropy does it your way a plethora of times](https://github.com/astropy/astropy/search?utf8=%E2%9C%93&q=numpy.testing.assert_allclose), so I'll switch to your pattern.

astroplan/tests/test_core.py
cdeil: Just one more comment: the function that prints the reference results to the console could live right next to the actual test function in this file. If it's name doesn't start with `test_` it will not be collected as a test, but you can still import and execute it manually. So the reference results for `test_Observer_altaz` could be computed by `print_reference_Observer_altaz`. I think this would be my preference, but yes, let's see what the others think.

astroplan/core.py
```diff
@@ -25,82 +40,106 @@ class Observer(object):
...
+            self.timezone = timezone
+        elif isinstance(timezone, str) or isinstance(timezone, unicode):
+            self.timezone = pytz.timezone(timezone)
+        else:
+            raise ValueError('timezone keyword should be a string, or an '
+                               'instance of datetime.tzinfo')
 
-        Must first create date_time object with get_date() before using any other 
-        Observer methods, all of which require one as an argument.
+    def altaz(self, target, time):
```
eteq: I suggest slightly changing this to be ``def altaz(self, time, target=None)``.  Then if the `target` is `None`, you return *just* the `AltAz` frame, without the data.  That is, the part of your implementation below that's inside `transform_to`.  If it's *not* None, you can then do something like ``return target.coord.transform_to(self.altaz(time, None))``.

Down the road it might be we can make big gains by caching this `AltAz` object for a particular `time`, but probably best not to do that yet (In the hallowed words of Donald Knuth: "premature optimization is the root of all evil")

astroplan/core.py
```diff
@@ -25,82 +40,106 @@ class Observer(object):
...
+        """
+        Returns an instance of `~astropy.coordinates.SkyCoord` with altitude and
+        azimuth of the `FixedTarget` called `target` at time `time`.
 
         Parameters
         ----------
-        input_date_time : WHAT TYPE IS astropy.time OBJECT?
+        target : `~astroplan.FixedObject`
```
eteq: This is an example of a place where we could get away with accepting either a `Target` *or* a coordinate: you can just do something like:
```
if hasattr(target, 'coord'):
    coord = target.coord
else:
    coord = target
```
and then below replace all `target.coord` references with just `coord`.
(also, I think you meant `FixedTarget` here?)

astroplan/core.py
adrn: Just adding a line note to remind us that this was resolved by a later commit -- the issue was a missing indentation in `astroplan/__init__.py` for a line that was importing from `astroplan.core`.

astroplan/core.py
```diff
@@ -25,82 +40,137 @@ class Observer(object):
...
-            The longitude of the observing location.
+        pressure : `~astropy.units.Quantity`
+            The ambient pressure.
 
-        latitude : str or `~astropy.units.quantity`?
-            The latitude of the observing location.
+        location : `~astropy.coordinates.EarthLocation`
+            The location (latitude, longitude, elevation) of the observatory.
+
+        longitude : str or `~astropy.units.Quantity`
```
adrn: Maybe add `(optional)` after `~astropy.units.Quantity`, and in the first line of the description specify that this is not the preferred method for instantiating an `Observer` (e.g., use `location=`). Same below for `latitude`

astroplan/core.py
```diff
@@ -25,82 +40,137 @@ class Observer(object):
...
-        <longer description>
+        if pressure is None:
+            pressure = 0
 
-        Parameters
-        ----------
-        pressure : `~astropy.units.Quantity`
-            The ambient pressure.
+        if temperature is None:
+            temperature = 0 * u.deg_C
```
adrn: I'm not sure we want to set the default (unset) values to 0 -- this is a valid number that will work in arithmetic operations but may return weird results. I think we want them to be `None` if they are unspecified. Here's an example: imagine a user forgets to specify `temperature`, and later on we add a method that internally uses the temperature for some calculation. Unless we check for 0, the calculation will proceed, but _may_ return nonsense. I think we want to just check for `None` values...

adrn: Another way to think about it: there is a conceptual difference between "`temperature=0`" and "`temperature` is unspecified"

astroplan/core.py
```diff
@@ -25,82 +40,137 @@ class Observer(object):
...
+        self.relative_humidity = relative_humidity
 
-    def get_date(self, input_date_time, timezone=None):
-        """
-        Builds an object containing date and time information.
+        # If lat/long given instead of EarthLocation, convert them
+        # to EarthLocation
+        if location is None and (latitude is not None and longitude is not None):
+            accepted_latlon_types = [str, Quantity]
+            if (type(latitude) in accepted_latlon_types and
```
adrn: Maybe I'm wrong here (@eteq let me know), but is this necessary? The initializer for `EarthLocation` should do something similar internally, so we may want to just pass the latitude and longitude directly in to the initializer and let that handle this?

astroplan/core.py
```diff
@@ -25,82 +40,137 @@ class Observer(object):
...
+                                                        elevation)
+
+        elif isinstance(location, EarthLocation):
+            self.location = location
 
-        Default time zone is UTC.
-        If time zone (local or otherwise) requested, then uses `datetime.datetime`
-        and pytz.timezone to convert from UTC.
+        else:
+            raise TypeError('Observatory location must be specified with '
```
adrn: Make me happy, and put `EarthLocation` as the 1st option here :) -- we want to encourage the OOP!

astroplan/core.py
```diff
@@ -25,82 +40,137 @@ class Observer(object):
...
 
         Parameters
         ----------
-        input_date_time : WHAT TYPE IS astropy.time OBJECT?
+        time : `~astropy.time.Time`
+            Astropy time object.
+
+        target : None (default) or `~astroplan.FixedTarget` or `~astropy.coordinates.SkyCoord`
+            Celestial object of interest. If `target`=None, return just the
+            `~astropy.coordinates.AltAz` frame without coordinates.
```
adrn: Add `obswl` to docstring?

astroplan/core.py
```diff
@@ -296,21 +366,40 @@ def ra(self):
...
 class FixedTarget(Target):
     """
     An object that is "fixed" with respect to the celestial sphere.
     """
-
+    def __init__(self, coord, **kwargs):
+        '''
+        TODO: Docstring.
+        '''
+        if not isinstance(coord, SkyCoord):
```
adrn: This should probably accept a `SkyCoord` or any of the `BaseCoordinateFrame` classes? (e.g., `ICRS`)

astroplan/core.py
```diff
@@ -296,21 +366,40 @@ def ra(self):
...
+            raise TypeError('Coordinate must be a SkyCoord.')
+
+        self.name = kwargs.get('name', None)
```
adrn: make `name` an explicit keyword argument?

astroplan/core.py
```diff
@@ -296,21 +366,40 @@ def ra(self):
...
+        self.coord = coord
+
+    @classmethod
+    def from_name(cls, query_name, **kwargs):
+        # Allow manual override for name keyword so that the target name can
+        # be different from the query name, otherwise assume name=queryname.
+        name = kwargs.pop('name', query_name)
```
adrn: add `target_name` or something as an optional but explicit keyword argument?

astroplan/tests/test_core.py
adrn: @cdeil What do you mean here? I think assert fails are useful to have for things that *must be true*, but haven't really thought about it...

cdeil: @adrn – I'm not sure I understand what you mean with "assert fails".
All asserts in tests should always be true, no?

I mean that here `, ('Locations not parsed equivalently')` can be removed because it's useless.

adrn: I think I mean "assert tests"... and I read you as saying "assert tests" as well! But now I see what you mean...oops

astroplan/core.py
eteq: Actually, I think this should go in the `Observer` class.  Everything else that goes into creating `AltAz` does, so it makes sense that `obswl` would, too. (but if it doesn't, then I agree with @adrn it needs to be there)

bmorris3: My strategy was that the `Observer` should really encapsulate just the observatory. Since you could observe multiple targets with the same observatory, or the same target at multiple wavelengths, I think we should keep `obswl` confined to specific calculations like `obs.altaz()`.

astroplan/core.py
bmorris3: I haven't played much with the other `BaseCoordinateFrame` subclasses – could generalizing introduce any subtleties that could cause errors I might not be thinking of?

astroplan/core.py
bmorris3: Anything to make you happy, @adrn!

astroplan/core.py
bmorris3: Deal. Setting *pressure* to zero by default is pretty common to specify an atmosphere-less calculation – we're still OK on that though, right?

bmorris3: @adrn – The place where that became handy for me was [in the `altaz` method](https://github.com/bmorris3/astroplan/blob/initclasses/astroplan/core.py#L158) where I copy over  the environment instance variables to initialize an AltAz frame. How would be the best way to handle that if `self.temperature = None`?

astroplan/core.py
bmorris3: It seemed to me from the [EarthLocation docs](http://astropy.readthedocs.org/en/latest/api/astropy.coordinates.EarthLocation.html#astropy.coordinates.EarthLocation.from_geodetic) that `EarthLocation.from_geodetic` takes lat/long as either `Latitude`s/`Longitude`s or floats, whereas `Latitude`/`Longitude` have [very flexible inputs](http://astropy.readthedocs.org/en/latest/api/astropy.coordinates.Longitude.html#astropy.coordinates.Longitude), which excited me.

astroplan/core.py
adrn: If I recall, `AltAz` has its own internal default assumptions, so let that class deal with this case?

astroplan/core.py
adrn: Works with `EarthLocation` too -- internally, it's probably passing the input in to let `Latitude` and `Longitude` classes do the parsing. Encapsulation!

e.g., this works:

```python
loc = coord.EarthLocation.from_geodetic(lon="150:21:41.1", lat=45*u.deg, height=15*u.m)
```

astroplan/core.py
bmorris3: What's the right pattern for calling `AltAz()` only the kwargs that have values other than `None`? Would you do a loop through possible keywords and search for ones that are not `None` and then pass those into `AltAz()`?

astroplan/core.py
adrn: (that is, you're just repeating code that already exists in the `EarthLocation` initializer)

bmorris3: Awesome, maybe this is an indication that the EarthLocation docs can be updated to specify their awesome input flexibility :)

adrn: Actually, it's useful for developers to know, but I think it is not in the docs because we want to encourage users to use quantities or objects (rather than strings! strings are baaaad! bad!)

adrn: (willing to reconsider that decision, if that was a decision and not just an oversight in the first place..)

bmorris3: You mean, I'm an *insider* now? :smile:

astroplan/tests/test_core.py
eteq: I don't see the problem here in a *general* sense, @cdeil : leaving a useful note about what you're trying to test with the assert is sort of self-documenting code in my mind.  You could instead put in a comment just above, but I think this works just as well.

That said, that *particular* string is not terribly informative in that it says something you could guess pretty much right away from the fact that it failed. I'd go with something more like “using latitude/longitude/elvation keywords gave a different answer from passing in an EarthLocation directly".

cdeil: @eteq – I agree leaving a note on the purpose of a given test / assert can be useful.
So far I have been doing this in docstrings or comments in tests, usually not at the place of the assert, but higher up where I start a new test. As far as I can see that's also the common thing to do in Astropy:
https://github.com/astropy/astropy/search?p=1&q=assert&utf8=%E2%9C%93

Another disadvantage if putting this note after the assert is that often you're almost at the end of the line, and then you have to wrap a lot (see the examples below in this PR).

But my main point is that asserts, for written and debugged code will very rarely fail, and thus it's just not worth to implement extra notes for every assert. If someone has a test failure, it's always worth reporting.
The user should copy & paste the console output into the issue report or a gist, and then a developer / maintainer will look at the test code anyways.

But bottom line: this is probably partly a matter of taste, and @bmorris3, if you like it this way, we'll still merge this tomorrow latest. :-)

astroplan/core.py
ejeschke: +1 to having an explicit keyword arg for name!

astroplan/tests/test_core.py
eteq: Yep, using `remote_data` like this is the way to go, @bmorris3.  

@cdeil - see astropy/astropy#3890 for a PR where I've added the relevant module.  I didn't realize that testing stuff wasn't anywhere in the docs...

astroplan/core.py
eteq: :+1: to @adrn here.  Also makes it easier to see what the default is.

astroplan/core.py
```diff
@@ -22,109 +38,152 @@
 
 class Observer(object):
     """
-    Some comments.
+    A container for information about an observatory's location and environment.
```
eteq: (minor quibble) maybe "a container class" instead of just "a container"?

eteq: also, maybe "observer" instead of "observatory"?  After all, it could just be a person outside with their eyes...

astroplan/core.py
```diff
@@ -22,109 +38,152 @@
...
-            The latitude of the observing location.
+        latitude : float, str, `~astropy.units.Quantity` (optional)
+            The latitude of the observing location. Should be valid input for
+            initializing a `~astropy.coordinates.Latitude` object.
 
-        elevation : `~astropy.units.Quantity`
+        elevation : `~astropy.units.Quantity` (optional)
             The elevation of the observing location, with respect to sea
-            level.
+            level. Defaults to zero meters.
```
eteq: perhaps "sea level" instead of "zero meters"?

astroplan/core.py
```diff
@@ -22,109 +38,152 @@
...
             The ambient relative humidity.
 
-        temperature : `~astropy.units.Quantity`
+        temperature : `~astropy.units.Quantity` (optional)
             The ambient temperature.
 
-        timezone : WHAT TYPE IS pytz.timezone ?
-            The local timezone.
+        timezone : str or `datetime.tzinfo` (optional)
+            The local timezone, as either an instance of `pytz.timezone()` or
```
eteq: The way you've phrased it here you're duplicating information because you're just saying the type.  Maybe instead "The local timezone to assume.  If a string, it will be passed through \`pytz.timezone()\` to produce the `timezone` object.

astroplan/core.py
```diff
@@ -22,109 +38,152 @@
...
         """
-        raise NotImplementedError()
-
-    def set_environment(self, pressure, relative_humidity, temperature):
-        """
-        Updates the Observer object with environmental conditions.
 
-        <longer description>
+        if elevation is None:
+            elevation = 0*u.m
```
eteq: better to just set this as the default for the argument.  After all, you don't store the fact that it was initially `None` anywhere.  So then it's manifestly clear what the true default is.

astroplan/core.py
```diff
@@ -22,109 +38,152 @@
...
+            self.timezone = pytz.timezone(timezone)
+        else:
+            raise TypeError('timezone keyword should be a string, or an '
+                            'instance of datetime.tzinfo')
 
-        Must first create date_time object with get_date() before using any other 
-        Observer methods, all of which require one as an argument.
+    def altaz(self, time, target=None, obswl=None):
+        """
+        Calculates an altitude/azimuth frame (if `target` not specified)
```
eteq: "calculates" -> "generates"?  without a target no actual "calcualtion" is done in the sense of a coordinate transform.  So maybe something like ``If `target` is None, generates an altitude/azimuth frame.  Otherwise, calculates the transformation to that frame for the requested `target`."`` ?

astroplan/core.py
```diff
@@ -22,109 +38,152 @@
...
+        or coordinate (otherwise).
 
         Parameters
         ----------
-        input_date_time : WHAT TYPE IS astropy.time OBJECT?
+        time : `~astropy.time.Time`
+            Astropy time object.
```
eteq: This is a repeat of the type information.  Instead maybe ``The time at which the observation is taking place.  Will be used as the `obstime` attribute in the resulting frame or coordinate.``

astroplan/core.py
```diff
@@ -22,109 +38,152 @@
...
+            Celestial object of interest. If `target` is `None`, return the
+            `~astropy.coordinates.AltAz` frame without coordinates.
+
+        obswl : `~astropy.units.Quantity` (optional)
+            Wavelength of the observation used in the calculation.
+
+        Returns
+        -------
+        `~astropy.coordinates.AltAz`
+            Frame or coordinate.
```
eteq: Isn't it either an `AltAz` *or* a `SkyCoord?`.  More properly (and this is a subtle but important point), if `target` is not None, this is the type that `target` was.  That is, if `target` is a `SkyCoord`, `tranform_to` yields a new `SkyCoord`.  But if it's, say, an `ICRS` coordinate object, `tranform_to` will yield an `AltAz` coordinate.  So this has to be gotten across somehow in the "Returns" description.

astroplan/core.py
eteq: :+1: to the @adrn-style extreme!

astroplan/core.py
```diff
@@ -296,21 +355,45 @@ def ra(self):
...
     """
     An object that is "fixed" with respect to the celestial sphere.
     """
-
+    def __init__(self, coord, name=None, **kwargs):
+        '''
+        TODO: Docstring.
+        '''
+        if not isinstance(coord, SkyCoord):
+            raise TypeError('Coordinate must be a SkyCoord.')
```
eteq: Oops, you don't want to do this.  We *do* want to allow people to pass in the "low-level" frame classes like `ICRS` or `AltAz` or whatever.  You want *some* kind of validation, though.  Perhaps the thing to do is to check for a few methods that really should be there?  Like `tranform_to` and `represent_as`?  This follows the python tradition of "duck-typing": when you see a bird that walks like a duck and swims like a duck and quacks like a duck, you call that bird a duck.  Similar with coordinates.

eteq:
I left a few more comments @bmorris3, but all but the last are phrasing/polish bits on the docs, nothing substantial on the code itself.  The last one is an important subtlety, though, that I think needs to be addressed (fortunately, it's easy to fix).

bmorris3:
@eteq : thanks for the comments! They've been addressed in the latest commit.

cdeil:
@eteq @adrn – Does one of you have time for a final review and hitting the merge button for this PR later today?

adrn:
This looks good to me -- way to go @bmorris3! I'm going to fire away and merge this.

cdeil:
:balloon: 

Have a nice weekend, everyone!

bmorris3:
:confetti_ball: One of many reasons to celebrate today!

adrn:
I merged this for :rainbow: !

PR #14, opened 2015-06-27
Rise/set/transit time calculations and convenience methods

bmorris3:
**Before beginning a code review/discussion, I recommend you [check out my notes](https://github.com/bmorris3/astroplan/wiki/Notes#on-risingsetting-methods)** on the organization of the methods that I've implemented in this PR.

I took some liberties to get things up and running, and am prepared to make plenty of changes as discussion begins. 

Happy Pride weekend! :rainbow:

### DIFF
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -198,8 +198,8 @@ class Observer(object):
         if target is None:
             return altaz_frame
         else:
-            if not (isinstance(target, FixedTarget) or
-                    isinstance(target, SkyCoord)):
+            if not (hasattr(target, 'ra') or hasattr(target, 'dec') or
+                    hasattr(target, 'coord')):
                 raise TypeError('The target must be a coordinate (i.e. a '
                                 'FixedTarget or SkyCoord).')
 
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -19,7 +19,7 @@ iers.IERS.iers_table = iers.IERS_A.open(download_file(iers.IERS_A_URL,
 ################################################################################
 
 
-#from ..extern import six
+from astropy.extern.six import string_types
 
 #import sys
 #from math import sqrt, pi, exp, log, floor
@@ -150,7 +150,7 @@ class Observer(object):
         # Accept various timezone inputs, default to UTC
         if isinstance(timezone, datetime.tzinfo):
             self.timezone = timezone
-        elif isinstance(timezone, str) or isinstance(timezone, unicode):
+        elif isinstance(timezone, string_types):
             self.timezone = pytz.timezone(timezone)
         else:
             raise TypeError('timezone keyword should be a string, or an '
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -239,32 +239,18 @@ class Observer(object):
             raise ValueError('Target does not rise/set with respect to '
                              '`horizon` within 24 hours')
 
+        # Isolate horizon crossing
         nearest_index = np.argwhere(condition)[0][0]
 
-        # Address special cases when nearest point is end point in array:
-        if nearest_index == 0:
-            lower_limit = t[0]
-            upper_limit = t[1]
-            lower_alt = alt[0]
-            upper_alt = alt[1]
-        elif nearest_index == len(alt):
-            lower_limit = t[-2]
-            upper_limit = t[-1]
-            lower_alt = alt[-2]
-            upper_alt = alt[-1]
-        # Normal cases:
-        elif alt[nearest_index] > horizon:
-            lower_limit = t[nearest_index-1]
-            upper_limit = t[nearest_index]
-            lower_alt = alt[nearest_index-1]
-            upper_alt = alt[nearest_index]
-        else:
-            lower_limit = t[nearest_index]
-            upper_limit = t[nearest_index+1]
-            lower_alt = alt[nearest_index]
-            upper_alt = alt[nearest_index+1]
+        # Determine points on either side of horizon crossing
+        if np.sign((alt[nearest_index] - horizon).value) != np.sign((alt[nearest_index+1] - horizon).value) and nearest_index != len(t):
+            lower_t, upper_t = t[nearest_index:nearest_index+2]
+            lower_alt, upper_alt = alt[nearest_index:nearest_index+2]
+        elif np.sign((alt[nearest_index] - horizon).value) != np.sign((alt[nearest_index-1] - horizon).value) and nearest_index != 0:
+            lower_t, upper_t = t[nearest_index-1:nearest_index+1]
+            lower_alt, upper_alt = alt[nearest_index-1:nearest_index+1]
 
-        return (lower_limit, upper_limit), (lower_alt, upper_alt)
+        return (lower_t, upper_t), (lower_alt, upper_alt)
 
     @u.quantity_input(horizon=u.deg)
     def _two_point_interp(self, times, altitudes, horizon=0*u.deg):
@@ -364,7 +350,8 @@ class Observer(object):
             times = _generate_24hr_grid(time, -1, 0, N)
 
         # Use trigonometric altitude calculation when pressure = 0
-        if self.pressure == 0*u.bar:
+        # (Need to check for pressure=None first or else throws FutureWarning)
+        if self.pressure is not None and self.pressure == 0*u.bar:
             LST = times.sidereal_time('mean', longitude=self.location.longitude)
             altitudes = self._altitude_trig(LST, target)
         else:
@@ -373,6 +360,60 @@ class Observer(object):
         horizon_crossing_limits = self._horiz_cross(times, altitudes, rise_set,
                                                     horizon)
         return self._two_point_interp(*horizon_crossing_limits, horizon=horizon)
+
+    def _calc_riseset_brentq(self, time, target, prev_next, rise_set, horizon,
+                             N=150, **kwargs):
+        '''
+        Time at next rise/set of ``target`` with `~scipy.optimize.brentq`.
+
+        Parameters
+        ----------
+        time : `~astropy.time.Time`
+            Time of observation
+
+        target : `~astropy.coordinates.SkyCoord`
+            Position of target or multiple positions of that target
+            at multiple times (if target moves, like the Sun)
+
+        prev_next : str - either 'previous' or 'next'
+            Test next rise/set or previous rise/set
+
+        rise_set : str - either 'rising' or 'setting'
+            Compute prev/next rise or prev/next set
+
+        location : `~astropy.coordinates.EarthLocation`
+            Location of observer
+
+        horizon : `~astropy.units.Quantity`
+            Degrees above/below actual horizon to use
+            for calculating rise/set times (i.e.,
+            -6 deg horizon = civil twilight, etc.)
+
+        N : int
+            Number of altitudes to compute when searching for
+            rise or set.
+
+        Returns
+        -------
+        ret1 : `~astropy.time.Time`
+            Time of rise/set
+        '''
+        if prev_next == 'next':
+            times = _generate_24hr_grid(time, 0, 1, N)
+        else:
+            times = _generate_24hr_grid(time, -1, 0, N)
+
+        altitudes = self.altaz(times, target).alt
+
+        time_limits, alt_limits = self._horiz_cross(times, altitudes, rise_set,
+                                                    horizon)
+
+        calc_altitude = lambda t: (self.altaz(Time(t, format='jd'),
+                                              target).alt - horizon).value
+        from scipy.optimize import brentq
+        root = brentq(calc_altitude, time_limits[0].jd, time_limits[1].jd,
+                      **kwargs)
+        return Time(root, format='jd')
 
     def _calc_transit(self, time, target, prev_next, antitransit=False, N=150):
         '''
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -218,7 +218,7 @@ class Observer(object):
             Grid of times
         alt : `~astropy.units.Quantity`
             Grid of altitudes
-        rise_set : str, either "rising" or "setting"
+        rise_set : {"rising",  "setting"}
             Calculate either rising or setting across the horizon
         horizon : float
             Number of degrees above/below actual horizon to use
@@ -233,9 +233,6 @@ class Observer(object):
             condition = (alt[:-1] < horizon) * (alt[1:] > horizon)
         elif rise_set == 'setting':
             condition = (alt[:-1] > horizon) * (alt[1:] < horizon)
-        else:
-            raise ValueError('`rise_set` keyword may only be "rising" or '
-                             '"setting"')
 
         if sum(condition) < 1:
             raise ValueError('Target does not rise/set with respect to '
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -233,11 +233,13 @@ class Observer(object):
         of the horizon crossing.
         '''
         if rise_set == 'rising':
+            # Find index where altitude goes from below to above horizon
             condition = (alt[:-1] < horizon) * (alt[1:] > horizon)
         elif rise_set == 'setting':
+            # Find index where altitude goes from above to below horizon
             condition = (alt[:-1] > horizon) * (alt[1:] < horizon)
 
-        if sum(condition) < 1:
+        if not np.any(condition) < 1:
             raise ValueError('Target does not rise/set with respect to '
                              '`horizon` within 24 hours')
 
@@ -1013,7 +1015,7 @@ class Observer(object):
         raise NotImplementedError()
 
     @u.quantity_input(horizon=u.deg)
-    def can_see(self, time, target, horizon=0*u.degree, ret_altaz=False):
+    def can_see(self, time, target, horizon=0*u.degree, return_altaz=False):
         '''
         Is ``target`` above ``horizon`` at this ``time``?
 
@@ -1033,16 +1035,16 @@ class Observer(object):
             for calculating rise/set times (i.e.,
             -6 deg horizon = civil twilight, etc.)
 
-        ret_altaz : bool (optional)
+        return_altaz : bool (optional)
             Also return the '~astropy.coordinates.AltAz' coordinate.
         '''
         if not isinstance(time, Time):
             time = Time(time)
 
         altaz = self.altaz(time, target)
-        observable = bool(altaz.alt > horizon)
+        observable = altaz.alt > horizon
 
-        if not ret_altaz:
+        if not return_altaz:
             return observable
         else:
             return observable, altaz
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -210,7 +210,8 @@ class Observer(object):
                              '"setting"')
 
         if sum(condition) < 1:
-            raise ValueError('Target does not rise/set within 24 hours')
+            raise ValueError('Target does not rise/set with respect to '
+                             '`horizon` within 24 hours')
 
         if return_limits:
             nearest_index = np.argwhere(condition)[0][0]
@@ -218,13 +219,13 @@ class Observer(object):
             if alt[nearest_index] > horizon:
                 lower_limit = t[nearest_index-1]
                 upper_limit = t[nearest_index]
-                lower_alt = alt[nearest_index-1]#.value
-                upper_alt = alt[nearest_index]#.value
+                lower_alt = alt[nearest_index-1]
+                upper_alt = alt[nearest_index]
             else:
                 lower_limit = t[nearest_index]
                 upper_limit = t[nearest_index+1]
-                lower_alt = alt[nearest_index]#.value
-                upper_alt = alt[nearest_index+1]#.value
+                lower_alt = alt[nearest_index]
+                upper_alt = alt[nearest_index+1]
 
             if return_alts:
                 return (lower_limit, upper_limit), (lower_alt, upper_alt)
@@ -262,8 +263,7 @@ class Observer(object):
         return Time(times[1].jd - ((altitudes[1] - horizon)/slope).value,
                     format='jd')
 
-    def _calc_riseset(self, time, target, prev_next, rise_set,
-                      horizon, N=150):
+    def _calc_riseset(self, time, target, prev_next, rise_set, horizon, N=150):
         '''
         Calculate the time at next rise/set of `target`.
 
@@ -305,11 +305,12 @@ class Observer(object):
             times = time + np.linspace(-1, 0, N)*u.day
 
         altitudes = self.altaz(times, target).alt
+        print('horizon = {}'.format(horizon))
         horizon_crossing_limits = self._horiz_cross(times, altitudes, rise_set,
                                                     horizon,
                                                     return_limits=True,
                                                     return_alts=True)
-        return self._two_point_interp(*horizon_crossing_limits)
+        return self._two_point_interp(*horizon_crossing_limits, horizon=horizon)
 
     @u.quantity_input(horizon=u.deg)
     def calc_rise(self, target, time, which='nearest', horizon=0*u.degree):
@@ -342,8 +343,8 @@ class Observer(object):
             Rise time of target
         '''
         if which == 'next' or which == 'nearest':
-            next_rise = self._calc_riseset(time, target, 'next',
-                                           'rising', horizon)
+            next_rise = self._calc_riseset(time, target, 'next', 'rising',
+                                           horizon)
             if which == 'next':
                 return next_rise
 
@@ -391,8 +392,8 @@ class Observer(object):
             Set time of target.
         '''
         if which == 'next' or which == 'nearest':
-            next_set = self._calc_riseset(time, target, 'next',
-                                          'setting', horizon)
+            next_set = self._calc_riseset(time, target, 'next', 'setting',
+                                          horizon)
             if which == 'next':
                 return next_set
 
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -357,7 +357,7 @@ class Observer(object):
             times = _generate_24hr_grid(time, -1, 0, N)
 
         # Use trigonometric altitude calculation when pressure = 0
-        if self.pressure is None or self.pressure == 0*u.bar:
+        if self.pressure == 0*u.bar:
             LST = times.sidereal_time('mean', longitude=self.location.longitude)
             altitudes = self._altitude_trig(LST, target)
         else:
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -36,7 +36,7 @@ __all__ = ["Observer", "Target", "FixedTarget", "NonFixedTarget",
 
 #__doctest_requires__ = {'*': ['scipy.integrate']}
 
-def _generate_24hr_grid(t0, start, end, N):
+def _generate_24hr_grid(t0, start, end, N, for_deriv=False):
     '''
     Generate a nearly linearly spaced grid of time durations.
 
@@ -58,13 +58,22 @@ def _generate_24hr_grid(t0, start, end, N):
     N : int
         Number of grid points to generate
 
+    for_deriv : bool
+        Generate time series for taking numerical derivative (modify
+        bounds)?
+
     Returns
     -------
     `~astropy.time.Time`
     '''
-    time_grid = np.concatenate([[start - 1/(N-1)],
-                                np.linspace(start, end, N)[1:-1],
-                                [end + 1/(N-1)]])*u.day
+
+    if for_deriv:
+        time_grid = np.concatenate([[start - 1/(N-1)],
+                                    np.linspace(start, end, N)[1:-1],
+                                    [end + 1/(N-1)]])*u.day
+    else:
+        time_grid = np.linspace(start, end, N)*u.day
+
     return t0 + time_grid
 
 class Observer(object):
@@ -367,6 +376,59 @@ class Observer(object):
                                                     horizon)
         return self._two_point_interp(*horizon_crossing_limits, horizon=horizon)
 
+    def _calc_transit(self, time, target, prev_next, antitransit=False, N=150):
+        '''
+        Time at next transit of the meridian of `target`.
+
+        Parameters
+        ----------
+        time : `~astropy.time.Time`
+            Time of observation
+
+        target : `~astropy.coordinates.SkyCoord`
+            Position of target or multiple positions of that target
+            at multiple times (if target moves, like the Sun)
+
+        prev_next : str - either 'previous' or 'next'
+            Test next rise/set or previous rise/set
+
+        antitransit : bool
+            Toggle compute antitransit (below horizon, equivalent to midnight
+            for the Sun)
+
+        location : `~astropy.coordinates.EarthLocation`
+            Location of observer
+
+        N : int
+            Number of altitudes to compute when searching for
+            rise or set.
+
+        Returns
+        -------
+        ret1 : `~astropy.time.Time`
+            Time of transit/antitransit
+        '''
+        if prev_next == 'next':
+            times = _generate_24hr_grid(time, 0, 1, N, for_deriv=True)
+        else:
+            times = _generate_24hr_grid(time, -1, 0, N, for_deriv=True)
+
+        # The derivative of the altitude with respect to time is increasing
+        # from negative to positive values at the anti-transit of the meridian
+        if antitransit:
+            rise_set = 'rising'
+        else:
+            rise_set = 'setting'
+
+        altitudes = self.altaz(times, target).alt
+        dt = Time((times.jd[1:] + times.jd[:-1])/2, format='jd')
+        d_altitudes = altitudes.diff()
+
+        horizon = 0*u.degree # Find when derivative passes through zero
+        horizon_crossing_limits = self._horiz_cross(dt, d_altitudes, rise_set,
+                                                    horizon)
+        return self._two_point_interp(*horizon_crossing_limits, horizon=horizon)
+
     @u.quantity_input(horizon=u.deg)
     def calc_rise(self, time, target, which='nearest', horizon=0*u.degree):
         '''
@@ -471,6 +533,92 @@ class Observer(object):
         raise ValueError('"which" kwarg must be "next", "previous" or '
                          '"nearest".')
 
+    def calc_transit(self, time, target, which='nearest'):
+        '''
+        Calculate time at the transit of the meridian.
+
+        Compute time of the next/previous/nearest transit of the ``target``
+        object.
+
+        Parameters
+        ----------
+        time : `~astropy.time.Time`
+            Time of observation
+
+        target : coordinate object (i.e. `~astropy.coordinates.SkyCoord`, `~astroplan.FixedTarget`)
+            Target celestial object
+
+        which : {'next', 'previous', 'nearest'}
+            Choose which sunrise relative to the present ``time`` would you
+            like to calculate
+
+        Returns
+        -------
+        `~astropy.time.Time`
+            Transit time of target
+        '''
+        if which == 'next' or which == 'nearest':
+            next_transit = self._calc_transit(time, target, 'next')
+            if which == 'next':
+                return next_transit
+
+        if which == 'previous' or which == 'nearest':
+            previous_transit = self._calc_transit(time, target, 'previous')
+            if which == 'previous':
+                return previous_transit
+
+        if which == 'nearest':
+            if abs(time - previous_transit) < abs(time - next_transit):
+                return previous_transit
+            else:
+                return next_transit
+        raise ValueError('"which" kwarg must be "next", "previous" or '
+                         '"nearest".')
+
+    def calc_antitransit(self, time, target, which='nearest'):
+        '''
+        Calculate time at the antitransit of the meridian.
+
+        Compute time of the next/previous/nearest antitransit of the ``target``
+        object.
+
+        Parameters
+        ----------
+        time : `~astropy.time.Time`
+            Time of observation
+
+        target : coordinate object (i.e. `~astropy.coordinates.SkyCoord`, `~astroplan.FixedTarget`)
+            Target celestial object
+
+        which : {'next', 'previous', 'nearest'}
+            Choose which sunrise relative to the present ``time`` would you
+            like to calculate
+
+        Returns
+        -------
+        `~astropy.time.Time`
+            Antitransit time of target
+        '''
+        if which == 'next' or which == 'nearest':
+            next_antitransit = self._calc_transit(time, target, 'next',
+                                                  antitransit=True)
+            if which == 'next':
+                return next_antitransit
+
+        if which == 'previous' or which == 'nearest':
+            previous_antitransit = self._calc_transit(time, target, 'previous',
+                                                      antitransit=True)
+            if which == 'previous':
+                return previous_antitransit
+
+        if which == 'nearest':
+            if abs(time - previous_antitransit) < abs(time - next_antitransit):
+                return previous_antitransit
+            else:
+                return next_antitransit
+        raise ValueError('"which" kwarg must be "next", "previous" or '
+                         '"nearest".')
+
     @u.quantity_input(horizon=u.deg)
     def sunrise(self, time, which='nearest', horizon=0*u.degree):
         '''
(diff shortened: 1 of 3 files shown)


bmorris3:
Just in case your notifications are set differently @astroplanners/astroplanners-contributors : take a look at this sometime this week! 

After discussion with @eteq today, we noted that I should relate the `N` parameter in `_calc_riseset()` to a precision of the rise/set time in units of time, and make a kwarg for sunrise/sunset that will allow users to adjust the precision (in exchange for speed).

cdeil:
@bmorris3 – Great!

Could you please have a look a the [coverage report](https://coveralls.io/builds/2920552/source?filename=astroplan%2Fcore.py) for the Observer class and add tests for the few missing statements that do computations (e.g. [here](https://coveralls.io/builds/2920552/source?filename=astroplan%2Fcore.py#L360)) (tests that cover the invalid input errors aren't really necessary.)

I built the docs locally and the "Methods Summary" section for the observer class is empty. I like having the overview of available methods ... it's much quicker than scrolling through the long descriptions. Anyone know why the "Methods Summary" isn't filled?

For the docstrings ... most don't have a `Returns` section because it's kind of obvious what they return. But it might still be useful to mention that a `Time` object is returned just to be clear ... maybe in the first line of the docstring if you think a `Returns` section is too verbose?

@bmorris3 – What do you think is the best strategy with narrative docs: http://astroplan.readthedocs.org/en/latest/#general-documentation ?
Add / improve piece by piece with each PR or only implement the classes and write in a separate PR later?

astroplan/core.py
```diff
@@ -162,56 +164,336 @@ def altaz(self, time, target=None, obswl=None):
...
+            Time of observation
 
-    def sunset(self, time, **kwargs):
-        """
-        Returns the local sunset time.
+        target : coordinate object (i.e. `~astropy.coordinates.SkyCoord`, `~astroplan.FixedTarget`)
+            Target celestial object
 
-        The default sunset returned is the next one to occur.
+        which : str - can be "next", "previous", or "nearest"; defaults to "nearest"
```
cdeil: The common way to document multiple-choice options in the Numpy docstring format is this:
```
which : {'next', 'previous', 'nearest'}
```

astroplan/core.py
```diff
@@ -162,56 +164,336 @@ def altaz(self, time, target=None, obswl=None):
...
+
+        altitudes = self.altaz(times, target).alt
+        horizon_crossing_limits = self._horiz_cross(times, altitudes, rise_set,
+                                                    horizon,
+                                                    return_limits=True,
+                                                    return_alts=True)
+        return self._two_point_interp(*horizon_crossing_limits, horizon=horizon)
+
+    @u.quantity_input(horizon=u.deg)
+    def calc_rise(self, time, target, which='nearest', horizon=0*u.degree):
```
cdeil: Usually in Python one should never have a mutable object as default value for keyword arguments.
@eteq, @adrn – Is this considered the best solution for functions that take quantities as input?

astroplan/core.py
```diff
@@ -162,56 +164,336 @@ def altaz(self, time, target=None, obswl=None):
...
 
-    def sunset(self, time, **kwargs):
-        """
-        Returns the local sunset time.
+        target : coordinate object (i.e. `~astropy.coordinates.SkyCoord`, `~astroplan.FixedTarget`)
+            Target celestial object
 
-        The default sunset returned is the next one to occur.
+        which : str - can be "next", "previous", or "nearest"; defaults to "nearest"
+            Choose which sunrise relative to the present `time` would you
```
cdeil: Here time needs to be in double-backticks ... you should always use double-backticks for variable names (and single-backticks for types.
If you check the HTML docs, you'll see that otherwise a link to here will be generated:
https://docs.python.org/2/library/time.html#module-time

astroplan/core.py
```diff
@@ -162,56 +164,336 @@ def altaz(self, time, target=None, obswl=None):
...
+            like to calculate
+
+        horizon : `~astropy.units.Quantity` (optional), default = zero degrees
+            Degrees above/below actual horizon to use
```
cdeil: "Degrees" -> "Angle" (the user's allowed to pass an `Angle` or `Quantity` in radian or milli-arcsec or whatever they like, right?

astroplan/core.py
```diff
@@ -269,57 +551,98 @@ def moon_position(self, time):
...
         """
-        Returns the local morning (nautical) twilight time.
+        Time at morning astronomical (-18 degree) twilight.
 
         Parameters
         ----------
         time : `~astropy.time.Time`
+            Time of observations
+
+         which : str - can be "next", "previous", or "nearest"; defaults to "nearest"
```
cdeil: This isn't formatted correctly in the HTML output. It's not obvious to me what's wrong ... maybe the semicolon.

astroplan/core.py
```diff
@@ -162,56 +164,336 @@ def altaz(self, time, target=None, obswl=None):
...
+        if which == 'nearest':
+            if abs(time - previous_rise) < abs(time - next_rise):
+                return previous_rise
+            else:
+                return next_rise
+        raise ValueError('"which" kwarg must be "next", "previous" or '
+                         '"nearest".')
+
+    @u.quantity_input(horizon=u.deg)
+    def calc_set(self, time, target, which='nearest', horizon=0*u.degree):
```
cdeil: It's not clear to me why we have `calc_set` and `calc_rise`, but all other methods don't have `calc` in their name, e.g. `moon_illumination` or `moonset` or ...

Can we drop the `calc_` here and just have a descriptive name of what's returned (i.e. `time`) ... maybe `set_time` and `rise_time` is good enough ... and if not, maybe `target_set_time` and `target_rise_time`?

astroplan/tests/test_core.py
```diff
@@ -155,3 +157,326 @@ def test_FixedTarget_ra_dec():
...
+    astroplan_prev_sunset = obs.sunset(time, which='previous').datetime
+
+    # Run get_pyephem_sunrise_sunset() to compute analogous
+    # result from PyEphem:
+    pyephem_next_sunrise = datetime.datetime(2000, 1, 2, 6, 3, 39, 150790)
+    pyephem_next_sunset = datetime.datetime(2000, 1, 1, 18, 3, 23, 676686)
+    pyephem_prev_sunrise = datetime.datetime(2000, 1, 1, 6, 3, 10, 720052)
+    pyephem_prev_sunset = datetime.datetime(1999, 12, 31, 18, 2, 55, 100786)
+
+    # Is there an equivalent to assert_allclose() for datetimes?
```
cdeil: No, there's no `assert_allclose` for datetimes. Looks like it's a common thing we do, so you can define a utility function at the top of the test file.

But shouldn't we use `astropy.time.Time` objects as much as possible, also in these tests?

astroplan/tests/test_core.py
```diff
@@ -155,3 +157,326 @@ def test_FixedTarget_ra_dec():
...
+    obs.lon = lon
+    obs.elevation = elevation
+    obs.date = time.datetime
+    obs.pressure = pressure
+    next_sunrise = obs.next_rising(ephem.Sun(), use_center=True)
+    next_sunset = obs.next_setting(ephem.Sun(), use_center=True)
+    prev_sunrise = obs.previous_rising(ephem.Sun(), use_center=True)
+    prev_sunset = obs.previous_setting(ephem.Sun(), use_center=True)
+
+    return (repr(next_sunrise.datetime()),
```
cdeil: Wouldn't it be more convenient to print this to the console instead of returning it?
Then I could do a one-liner to quickly check the reference values:
```
python -c 'from astroplan.tests.test_core import print_pyephem_sunrise_sunset as f; f()'
```

astroplan/tests/test_core.py
```diff
@@ -155,3 +157,326 @@ def test_FixedTarget_ra_dec():
...
+    obs = Observer(location=location, pressure=pressure)
+    astroplan_next_rise = obs.calc_rise(time, vega, which='next').datetime
+    astroplan_next_set = obs.calc_set(time, vega, which='next').datetime
+
+    astroplan_prev_rise = obs.calc_rise(time, vega, which='previous').datetime
+    astroplan_prev_set = obs.calc_set(time, vega, which='previous').datetime
+
+    # Run get_pyephem_vega_rise_set() to compute analogous
+    # result from PyEphem:
+    pyephem_next_rise, pyephem_next_set, pyephem_prev_rise, pyephem_prev_set = (
```
cdeil: I think it's more readable if you assign one variable per line ... simpler to see which value goes with which name.
If it's > 80 characters I'd say that's OK here.

astroplan/tests/test_core.py
```diff
@@ -155,3 +157,326 @@ def test_FixedTarget_ra_dec():
...
+
+    astroplan_prev_sunrise = obs.sunrise(time, which='previous',
+                                         horizon=horizon).datetime
+    astroplan_prev_sunset = obs.sunset(time, which='previous',
+                                       horizon=horizon).datetime
+
+    astroplan_next_real_sunrise = obs.sunrise(time, which='next',
+                                              horizon=0*u.degree).datetime
+
+    print(astroplan_next_sunrise, astroplan_next_real_sunrise)
```
cdeil: Remove print from test.

astroplan/core.py
bmorris3: I addressed this point in the [notes](https://github.com/bmorris3/astroplan/wiki/Notes#on-risingsetting-methods), but I'm interested in what all @astroplanners/astroplanners-contributors think about this. When I here `set_time`, it sounds to me like it's a setter and it's changing an internal time, so I wanted to avoid the `*_time` naming scheme. `target_set_time ` seems longer and less descriptive.

cdeil:
@bmorris3 – I've left a bunch of nitpicky comments about the docstrings. If you disagree with some, OK, I don't feel very strongly about them, it's just how I would have written it.

astroplan/tests/test_core.py
bmorris3: I'm comparing with `datetime` objects as outputs because PyEphem outputs datetimes. This way we're comparing apples to apples... and if the `astropy.time` conversion to datetime objects works, shouldn't that be fine?

astroplan/tests/test_core.py
bmorris3: I was running it interactively. That helps to know what you had in mind when you suggested I write these tests.

bmorris3:
> I built the docs locally and the "Methods Summary" section for the observer class is empty.

@cdeil How are you building the docs locally? When I run `python setup.py build_sphinx -lo`, I get the methods summary. Were you maybe looking at `Observation` rather than `Observer`?

astroplan/core.py
cdeil: I think `calc_set` is very cryptic.
Maybe if you think `set` = `setter` is confusing, using `setting` could be an option?
http://rhodesmill.org/pyephem/quick#transit-rising-setting

astroplan/tests/test_core.py
cdeil: Yes, it's fine. So Astropy uses mostly `Time` objects and pyephem uses mostly `datetime` objects. You can convert between the two easily using the `Time` class. It's not very important, but I'm +1 to use `astropy.time.Time` as much as possible everywhere, including tests.

astroplan/tests/test_core.py
bmorris3: I did this because I was lazy and that block of datetimes was the copied and pasted result of the `get_pyephem_*()` method. I'll change this.

cdeil:
I'm getting an empty "Methods summary" section of the Observer class:

screen shot 2015-06-29 at 23 44 24

Weird that it's there for you ... I also use `python setup.py build_sphinx -l`.
Can someone else check, please?

astroplan/core.py
bmorris3: I still hear the verb "to set" when I hear setting, as in "what are the settings?" This might just be me being annoying.

bmorris3:
This is what I see: screen shot 2015-06-29 at 2 33 53 pm

astroplan/core.py
jberlanga: I agree with @bmorris3 that `set` and `setting` both seem ambiguous at first, but I would go with `set_time`/`rise_time`--introducing them as a pair makes it clear what they are.  Also, users can't update the time, so `set_time` can't be a method for this.

Alternatively, `target_set` and `target_rise`.

astroplan/tests/test_core.py
ejeschke: I think we should have some tests that pass `datetime` objects (in addition to `Time` ones) just to verify that this works as expected.

bmorris3:
After the Hangout with @cdeil , we have some things to discuss (in particular, @cdeil and @eteq might have some things to hash out).

#### Precision
These methods work and will produce high precision results where necessary, but can we decided on what use cases they will matter in the most? In our conversation, the rise/set/transit times that we thought were most important to solve with precision and with consideration for the atmosphere were for the Sun, but in almost all other cases, the rise and set times can be calculated with low precision and without the atmosphere, since almost no observers will be looking at a target on the horizon (apart from solar observers).

@cdeil suggests that I implement a simple equivalent to my fancy algorithm that gets the altitude of targets over time from simple trigonometry rather than astropy's AltAz frame, since this is the biggest bottleneck in the computation. Plus, I've already implemented [Meeus' trigonometric methods before](https://github.com/bmorris3/gsoc2015/blob/688e7431bad70ea4384e4b8c2f395c1d41fb3af8/riseset.ipynb) in a function that would be easy to implement the framework already developed in this PR. 

Would anyone (@eteq) be opposed to imposing that when the user asks for rise/set of a target other than the sun, we default to using the trigonometric approximations?

#### Vectorization
@cdeil brought up the good point that we should perhaps be more concerned with properly implementing vectorization in these algorithms that with their precision for now. After this PR gets merged, maybe I should focus on implementing use cases (like computing rise/set on hundreds of targets simultaneously) rather than making super-awesome calculators.

bmorris3:
@cdeil @eteq: Thinking on this some more, we could always use the trigonometric rise/set equations when `pressure=0`. That would be a simple setting for users to figure out, and it remains self-consistent.

bmorris3:
@cdeil : I've implemented the faster trigonometric method, which gives a factor of ~3-4 speedup. I've set the package to use the faster method when the atmospheric pressure is set to zero.

astroplan/core.py
```diff
@@ -162,56 +195,359 @@ def altaz(self, time, target=None, obswl=None):
...
+
+        target : {`~astropy.coordinates.SkyCoord`, `FixedTarget`} or similar
+            Target celestial object's coordinates.
+
+        Returns
+        -------
+        alt : `~astropy.unit.Quantity`
+            Array of altitudes
+        '''
+        alt = np.arcsin(np.sin(self.location.latitude.radian) *
```
cdeil: I think `np.sin` works with Astropy angles, i.e. you can remove the `.radian` attribute access and get the same result?
Can you do a quick check if there is a speed difference, i.e. if it's slower or faster if you explicitly call `.radian` before passing to `np.sin`?

bmorris3: It's actually faster without the radian call!
```
>>> from astropy.coordinates import Longitude
>>> a = Longitude(np.linspace(0, np.pi, 100)*u.radian)
>>> %timeit np.sin(a)
10000 loops, best of 3: 77.1 µs per loop
>>> %timeit np.sin(a.radian)
10000 loops, best of 3: 25.7 µs per loop
```

cdeil:
It would be great if @ejeschke, @eteq or @adrn could have a first look at this and comment on the big picture ... is this going in the right direction? should any big changes to API or implementation be made?

I think I've said these things before, but just in case, here's what I think:
- Make vectorisation for targets work first and add a test where you have a few targets that are spread out over the sky.
- Check if the trigonometric formulae agree with the other method you have here or some other software (e.g. pyephem) for the case of zero pressure. It's not obvious to me that refraction is the only concern here, but I think any other effects (like the fact that the Earth is not perfectly round) aren't really important here?
- Check that transit time really doesn't change when changing the pressure (because it's a shift in zenith direction).
- If the others say that the root-finding algorithm for the rise / set time should be kept, because it's important to have for some reason, it should be improved by: Change the course 24 hour grid to a grid that just covers the minute or so where you know the solution is. Investigate both using `scipy.optimize.brentq` and if you want to use interpolation, I'd suggest switching to one of the scipy interpolation methods that results in some function with analytical roots (I think the splines have that) instead of implementing your own interpolation.
- IMO it's more important to move on and extend the astroplan toolbox instead of spending a week on precise / fast numerical methods here. This can always be done at the end of GSoC if we really think it's important.

astroplan/core.py
cdeil: I find the overhead of letting `np.sin` do the unit checking / conversion as a function of array size weird:
```
In [17]: a = Longitude(np.linspace(0, np.pi, 100000)*u.radian)

In [18]: %timeit np.sin(a)
1000 loops, best of 3: 1.09 ms per loop

In [19]: %timeit np.sin(a.radian)
1000 loops, best of 3: 1.13 ms per loop
```
For very large arrays, it should be a fixed-time overhead if at all, if an Angle in radians is passed in.
Could be interesting to follow up and try to understand ... a possibility for performance improvement of `Angle`, but I don't have time for this, and it's not important for this PR.

ejeschke:
I agree that the big picture toolbox should probably get more attention now and come back to any optimizations later.  It would be very nice to be able to complete the basic toolbox before the end of the GSoC period because then we could maybe even move on to a very basic scheduling layer API on top of it.  But this will be more difficult without the foundation.

eteq:
On the trig vs. `AltAz` (I'll have more time to look at the rest of this later, but I thought this is probably the biggest current question): my first-level thought is that if it's only a factor of 3x - 4x sppedup to drop `AltAz`, then we should *not* do it.  There are a few reasons for this:

1. How are you going to deal with the case where the user supplies a target coordinate that's not in ICRS?  You'll have to transform to `ICRS` then, which will probably make things *slower* because it adds an extra transform that's not needed if you can go direct to `AltAz` (because most of the cost is in creating the objects).  I suspect that will make the whole thing quite a bit more complicated.
2. More generally, taking the `ICRS` coordinates and using that to compute rise/set/transit times using trig algorithms is not right because it neglects effects like abberation, light deflection, etc.  These are only probably only of concern at the sub-minute level (although for aberration it can be as much as 20 seconds), but some use cases probably want sub-minute precision.  So using `AltAz` with an option to make the grid finer allows users to say "I want precision as good as `x`' without requiring you to do any extra work beyond deciding how to map precision onto grid spacing.
3. One of the nice things about the `AltAz` approach is that it's relatively straightforward for an (advanced) user to provide their own (say) atmospheric refraction scheme.  They just define a new frame with a transform to `AltAz` that accounts for the effect they want, and then send their coordinates to that before sending them on to  the rise/set code.  This is the sort of flexibility is what the coordinates framework is designed for, and circumventing it prevents that.

Another thing to consider: it might be worthwhile (*not* necessarily right now) to do some profiling of the `AltAz` transformation to determine the bottleneck. There's likely a lot of optimization left to be done in the transform code, particularly with regard to `__init__` of `SkyCoord` and frame objects.  I would not be surprised to find a 3x speedup is possible down the road, which would mean no downside (although it's hard to say for sure).

Also, @bmorris3, did you get a chance to do those tests we discussed in our short hangout of checking the precision of the `AltAz` grid-search approach over a grid of both on-sky coordinates and different on-Earth positions?  That might be useful for doing the precision comparisons @cdeil was talking about.

eteq:
All of this said, it *might* make sense to use the trig approach (with suitable "guard" methods to transform to ICRS before they are applied) if the requested precision is below a certain level.  But I think that might end up more confusing both for devs and users if the calculation is done in a totally different way for <1 minute vs > 1 minute.

eteq:
Oh, and I also don't think it's wise for the sun to be special-cased (i.e. trig for everything but the sun): that adds extra burden both for support and for users who won't understand why they're getting mysteriously different answers if they do something we don't quite expect (like my item 3 above)

bmorris3:
@eteq @adrn @ejeschke – More mundane topic of debate: what do we think of the naming scheme for the generic rise/set methods? I'm not in love with `calc_rise()`/`calc_set()` because of the extraneous calc, but I think plain `rise()`/`set()` risks confusion with a setter method – is that just paranoia?

bmorris3:
In the last two commits I've done the last major implementations for this PR, further additions from here will likely be to address comments on the implemented code. The methods that were missing were for calculating transit/antitransit, and the solar convenience functions associated with transit (noon/midnight).

I also made [a slight change to the API](https://github.com/astroplanners/astroplan/pull/14/files#diff-d1475973061987d43ea4fea64513b64aR205) to make things more consistent. Nearly everywhere in the `Observer` class, the first argument of the method is a time, *except* the `altaz` method specified in the API. I switched the order around to prevent confusion. I also removed some keywords that are just plain arguments as implemented.

bmorris3:
@cdeil @adrn @eteq – The tests are currently failing because [I verify that](https://github.com/bmorris3/astroplan/blob/rise_set/astroplan/core.py#L153) the `Observer` kwarg `timezone` is either the result of a call to `pytz.timezone()`, or a string (which can be passed into that function). I do this like so: 
```python
elif isinstance(timezone, str) or isinstance(timezone, unicode):
```
which is not Python 3 friendly since `unicode` has been dropped. I see that in astropy [it is common to use `string_types` from six](https://github.com/astropy/astropy/search?utf8=%E2%9C%93&q=six+string_types&type=Code), but we haven't yet set up six. Is that something we plan to do?

cdeil:
Astropy-affiliated packages that have Astropy as a required dependency (i.e. most, including `astroplan`) should use this version of six (no need to bundle our own or add it as a dependency):
```python
from astropy.extern import six
```

bmorris3:
Thanks @cdeil! Fixed and passing.

ejeschke:
I agree that set() is probably a bad choice.  I don't have a problem with
using a calc_ prefix if you really are calculating something significant.
If that's disagreeable though, how about another prefix?

On Wed, Jul 1, 2015 at 7:17 AM, Brett M. Morris <notifications@github.com>
wrote:

> @eteq <https://github.com/eteq> @adrn <https://github.com/adrn> @ejeschke
> <https://github.com/ejeschke> – More mundane topic of debate: what do we
> think of the naming scheme for the generic rise/set methods? I'm not in
> love with calc_rise()/calc_set() because of the extraneous calc, but I
> think plain rise()/set() risks confusion with a setter method – is that
> just paranoia?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/astroplanners/astroplan/pull/14#issuecomment-117756075>
> .
>



-- 
Eric Jeschke

astroplan/core.py
eteq: I did some quick tests and I do see it as a fixed-time overhead (as you thought it should be, @cdeil) - not sure why it's different.  Presumably the extra cost over `radian` comes from having to do the extra bits to create a `Quantity`.

astroplan/core.py
```diff
@@ -118,17 +158,19 @@ def __init__(self, name=None, location=None, latitude=None, longitude=None,
...
-        calculates the transformation to that frame for the requested `target`.
+        Get an `~astropy.coordinates.AltAz` frame or coordinate.
+
+        If ``target`` is None, generates an altitude/azimuth frame. Otherwise,
+        calculates the transformation to that frame for the requested ``target``.
 
         Parameters
         ----------
-        time : `~astropy.time.Time`
+        time : Anything accepted by `~astropy.time.Time`
```
eteq: I think it might be better to be a bit more specific about what this means.  The average user might not understand what "accepted by" means.  Maybe just "as the first argument to" is clear enough?  I also think it would be useful to specify that it could be just a `Time` object.  It may not be obvious to everyone that a `Time` accepts a `Time` as a first argument...

(same for all the other places, hopefully that's just a find-and-replace-all, though)

astroplan/tests/test_core.py
```diff
@@ -155,3 +152,591 @@ def test_FixedTarget_ra_dec():
...
+            datetime.timedelta(minutes=threshold_minutes))
+    assert (abs(astroplan_next_midnight - pyephem_next_antitransit) <
+            datetime.timedelta(minutes=threshold_minutes))
+    assert (abs(astroplan_prev_noon - pyephem_prev_transit) <
+            datetime.timedelta(minutes=threshold_minutes))
+    assert (abs(astroplan_prev_midnight - pyephem_prev_antitransit) <
+            datetime.timedelta(minutes=threshold_minutes))
+
+def print_pyephem_solar_transit_noon():
+    '''
```
eteq: Presumably this is fairly obvious, but you might want to note in the docs here that pyephem must be installed, and also that this isn't actually run as part of the testing suite itself (to make it clear that pyephem is not a dependency)

astroplan/tests/test_core.py
```diff
@@ -155,3 +152,591 @@ def test_FixedTarget_ra_dec():
...
+    consistent with PyEphem results (for no atmosphere)
+    for the fast method assuming pressure=0, versus with less fast
+    method where pressure=None.
+    '''
+    lat = '00:00:00'
+    lon = '00:00:00'
+    elevation = 0.0 * u.m
+    pressure = 0 * u.bar
+    location = EarthLocation.from_geodetic(lon, lat, elevation)
+    time = Time('2000-01-01 12:00:00')
```
eteq: For this test I think it's insufficient to only do lat/lon=0,0 and 2000-1-1.  Those are exactly where you expect it to be simplest/most accurate, and we want to test where it's most likely to go *wrong*.  (A bit more about this in my PR comments)

astroplan/tests/test_core.py
```diff
@@ -155,3 +152,591 @@ def test_FixedTarget_ra_dec():
                                                         'SkyCoord')
     assert vega.coord.dec == vega_coords.dec == vega.dec, ('Retrieve Dec from '
                                                            'SkyCoord')
+
+def test_sunrise_sunset_equator():
+    '''
+    Check that time of sunrise/set for an observer on the equator is
+    consistent with PyEphem results (for no atmosphere/pressure=0)
```
eteq: Perhaps you should also use some other comparison datapoint like the USNO sunrise/sunset calculator?  PyEphem shouldn't really be taken as any more "authoritative" than anywhere else in this respect, so comparing a few different things might be useful.

eteq:
@bmorris3  - strangely, I thought I had left comments on this two days ago, but I see now that none of them appeared... I was having network issues at the time, so maybe that's it.  Anyway, I've looked again - looks pretty good to me aside from the question of trig vs. AltAz approach and a few comments I left above and below.

An important point about the tests, though: the rise/set (and maybe transit) calculations need to be tested for a variety of different on-earth locations and times.  E.g., the effect about precession only kicks in for times away from 2000, and latitudes that are negative/positive have different effects.

We also want to have useful behavior if someone asks for the rise/set time of something that's circumpolar or never rises.  Probably should just raise a specialized exception for those cases.

One thing to add about my above comments: after looking more closely I think my # 2 above may not be an issue, because you're using the local sidereal time calculation which *includes* precession.  I'm not 100% sure that accounts for all of it, but it should at least be some.  All the more reason to do tests not at 2000-1-1, though.

astroplan/tests/test_core.py
bmorris3: [USNO uses some assumptions](http://aa.usno.navy.mil/faq/docs/RST_defs.php#riseset) about the atmosphere and which limb of the sun/moon defines the rise/set, which might make it difficult to compare without knowing their assumptions better. Do you know where to look for that?

bmorris3:
> E.g., the effect about precession only kicks in for times away from 2000

@eteq : I'm concerned that since we use up-to-date IERS standards and PyEphem does not, we are not comparing apples to apples when checking one with the other. PyEphem uses standards from 1980, so I wanted to pick times close to then (rather than to today). Is there another package that we should be using to check these things (like Skyfield) which may be more up to date and appropriate for these comparisons? 

Our predictions are still off by ~5 minutes in these best of circumstances anyway, so I feel like our requirements necessary to consider the two packages to be "in agreement" are broader than the corrections introduced by precession/nutation.

> a variety of different on-earth locations and times

I think this format that I'm using with a `print_pyephem_*()` method that complements a `test_*()` method will get unwieldy rather quickly if we ran tests like these over a grid of locations and times. How would you handle this @eteq ? While we think about that, I've been doing sanity checks on grids of observers/targets [by hand](https://github.com/bmorris3/gsoc2015/blob/master/sanity_checks/gridofobservers.ipynb).

astroplan/tests/test_core.py
bmorris3: Where would this be done in the docs, @cdeil / @eteq?

cdeil:
@bmorris3 – There's https://github.com/astropy/coordinates-benchmark which would be the appropriate location to implement checks against other codes  / reference values for largish grids of parameters (e.g. locations, times, ...).

As I've said before, my strong preference would be that this kind of work (root finding code to find rise / set times, extensive checks for many locations / times) is postponed to the end of GSoC. Let's get a first version of the toolbox in master first (within a week or two possible?), so that others that want to work on e.g. plotting or scheduling aren't blocked and so that there's something we can start to use and get real-world feedback on. It really doesn't matter for now if the sun sets a few minutes early or our package doesn't work for the eskimos and penguins.

@bmorris3, just to be clear – this is in now way a critique of your work ... I'm very impressed at what you're doing and how fast you're exploring / implementing things. It's just that in my experience trying to get it "right" on the first go and be comprehensive in testing is hard and building a working prototype quickly and iterating on it is easier and leads to quicker / better results.

I guess to be re-discussed tomorrow evening in the hangout ...

astroplan/tests/test_core.py
cdeil: I'm not sure it's really needed, but if you add this to the docs it should be on a separate page for developers, not mixed in with docs for end-users.

Example how I did this in Gammapy ... I used a Q & A format: https://gammapy.readthedocs.org/en/latest/development/index.html

Astropy has this: http://astropy.readthedocs.org/en/latest/#developer-documentation
I'm not sure it's a useful example for such a small package as astroplan.

eteq:
@cdeil - I agree with you about this not being a high priority, but that's why I think it's *better* to go with all `AltAz` for now: we don't need any checking to make sure it's consistent with the "trig" answer if we don't use the trig answer at all.  So I was suggesting we stick with the `AltAz` approach for the present and treat adding a trig answer (or better yet, figuring out where to optimize `AltAz`), is better done later once we can see if it's really a critical bottleneck.

Similarly, my suggestion about making sure we cover a grid of points was not meant to be exhaustive like coordinates-benchmark, but rather a relatively quick check to ensure something isn't very wrong.  Like select 25 random points on the earth and 25 points in the sky or something like that.  Just doing 0/0 lat/lon and 2000-1-1 is not sufficient here because most of the failure modes I know of are worse the further you go from there. Not something that should take a lot of time, just a few minutes of adding a for loop in pyephem to get the "good" answers for a few cases.

Also, @bmorris3, one related point of clarification: the approach PyEphem uses isn't actually fixed to 1980, if I understood the tweet correctly: rather it's an algorithm *published* in 1980.  They differ from the astropy (2006) algorithms only at arcsecond levels at the very most (or centuries in the past/future), so it shouldn't be noticable at all here.

I don't agree with the "doesn't work for the eskimos and penguins" bit, though: the case I had in mind   is if I'm at, say, Subaru, and ask for the rise/set time of Polaris.  That's a nonsense question because it's circumpolar, and its important for observing planning to know that. and for the rise/set methods to appropriately do something in that case.  Similarly, If I ask for the rise/set time of the Large Magellanic Cloud, it should also tell me that the LMC *never* rises at Subaru.

cdeil:
@adrn @ejeschke @eteq – Do you have time for a final review of this PR? (I don't.)
It would be great if we can merge this tomorrow or Friday.

Sorry I missed the hangout yesterday!

adrn:
@astroplanners/astroplanners-contributors  I'll look this over today -- sorry, yesterday was a bit crazy.

bmorris3:
@adrn If you have any questions, find me on Slack! I look forward to getting this merged and moving on.

cdeil:
@bmorris3 – What about @eteq's last comment? What currently happens when you ask for rise times for objects that never rise or set? Do you think this behaviour is appropriate? Or do you want to improve it here or make a new issue to discuss how to handle this? Is there a test covering this?

bmorris3:
@cdeil I'm working on [custom exceptions](https://github.com/bmorris3/gsoc2015/blob/master/custom_exceptions.ipynb) for these cases.

bmorris3:
@cdeil I will implement these in a separate PR once this one is merged. I'm managing a whole bunch of overlapping branches of code, and it would be really convenient if rise/set was merged so I can build off of it.

cdeil:
@adrn – I'm +1 to merge this today and let @bmorris3 continue in other PRs ... I'll leave the final review and merging to you.

@bmorris3 – I didn't look yet and your exceptions to handle non-rising and non-setting objects. But raising an exception means it's not possible to process arrays of targets where only some rise, right? What numpy does there is raise a warning (which can be suppressed), but process the array and return `np.nan` or `np.inf` ... example:
```python
>>> import numpy as np
>>> np.sqrt([4, -4, 9, -9])
__main__:1: RuntimeWarning: invalid value encountered in sqrt
array([  2.,  nan,   3.,  nan])
```
I just briefly looked at your tests ... doesn't look like arrays or targets are supported (or at least not tested) at the moment? Is it possible / easy to implement? Do you plan to do it in a future PR?
(one thing that can be useful is to start a new issue or PR with a task list and add the things you don't want to address in this PR, to make sure they aren't forgotten)

bmorris3:
@cdeil The next really important thing to work on is to allow for easy vectorization of all of the methods implemented in this PR. The purpose of moving forward from this PR is to get agreement on the approach for single targets so we can move onto the more important cases with multiple targets.

astroplan/core.py
```diff
@@ -162,171 +207,851 @@ def altaz(self, time, target=None, obswl=None):
...
+            Number of degrees above/below actual horizon to use
+            for calculating rise/set times (i.e.,
+            -6 deg horizon = civil twilight, etc.)
+        Returns
+        -------
+        Returns the lower and upper limits on the time and altitudes
+        of the horizon crossing.
+        '''
+        if rise_set == 'rising':
+            condition = (alt[:-1] < horizon) * (alt[1:] > horizon)
```
adrn: Shouldn't these `*`s be `&`?

astroplan/core.py
```diff
@@ -162,171 +207,851 @@ def altaz(self, time, target=None, obswl=None):
...
+        horizon : float
+            Number of degrees above/below actual horizon to use
+            for calculating rise/set times (i.e.,
+            -6 deg horizon = civil twilight, etc.)
+        Returns
+        -------
+        Returns the lower and upper limits on the time and altitudes
+        of the horizon crossing.
+        '''
+        if rise_set == 'rising':
```
adrn: Took me a minute to parse what was going on here -- maybe add a comment above the `if` that just says "# vectorized sign flip test or something"

astroplan/core.py
```diff
@@ -162,171 +207,851 @@ def altaz(self, time, target=None, obswl=None):
...
+            condition = (alt[:-1] < horizon) * (alt[1:] > horizon)
+        elif rise_set == 'setting':
+            condition = (alt[:-1] > horizon) * (alt[1:] < horizon)
 
-    def midnight(self, time):
-        """
-        Returns the local, solar midnight time.
+        if sum(condition) < 1:
```
adrn: This is minor, but might make more logical sense "for the reader" to do
```python
if not np.any(condition):
...
```

astroplan/core.py
adrn: Yea, I don't necessarily advocate changing it for this reason, but stylistically it is confusing that you pass in `target.dec` for one trig call (an `Angle` subclass) but then specify `self.location.latitude.radian` for another...

astroplan/core.py
adrn: There is no consensus, so I suggest that our placeholder be more explicit -- I propose to make it `calculate_target_set_time` (and same for rise) until someone comes up with better names!

astroplan/core.py
```diff
@@ -162,171 +207,851 @@ def altaz(self, time, target=None, obswl=None):
...
+        if which == 'nearest':
+            if abs(time - previous_set) < abs(time - next_set):
+                return previous_set
+            else:
+                return next_set
+
+        raise ValueError('"which" kwarg must be "next", "previous" or '
+                         '"nearest".')
+
+    def calc_transit(self, time, target, which='nearest'):
```
adrn: `calculate_target_meridian_transit`!

astroplan/core.py
```diff
@@ -162,171 +207,851 @@ def altaz(self, time, target=None, obswl=None):
...
+        time : `~astropy.time.Time` or other (see below)
+            This will be passed in as the first argument to
+            the `~astropy.time.Time` initializer, so it can be anything that
+            `~astropy.time.Time` will accept (including a `~astropy.time.Time`
+            object).
         """
         raise NotImplementedError()
 
+    @u.quantity_input(horizon=u.deg)
+    def can_see(self, time, target, horizon=0*u.degree, ret_altaz=False):
```
adrn: In keeping with naming of other methods with similar boolean returns, I'd name this `is_visible` or something (e.g., [isscalar](http://docs.scipy.org/doc/numpy/reference/generated/numpy.isscalar.html), [isfinite](http://docs.scipy.org/doc/numpy/reference/generated/numpy.isfinite.html), [isiterable](http://docs.astropy.org/en/latest/api/astropy.utils.misc.isiterable.html#astropy.utils.misc.isiterable)).

I'd also make `ret_altaz` --> `return_altaz`

astroplan/core.py
```diff
@@ -162,171 +207,851 @@ def altaz(self, time, target=None, obswl=None):
...
+            -6 deg horizon = civil twilight, etc.)
+
+        ret_altaz : bool (optional)
+            Also return the '~astropy.coordinates.AltAz' coordinate.
+        '''
+        if not isinstance(time, Time):
+            time = Time(time)
+
+        altaz = self.altaz(time, target)
+        observable = bool(altaz.alt > horizon)
```
adrn: This will fail if you the `altaz` multiple targets or times -- why do you wrap it with `bool`? The result of a comparison operator will either be a boolean (if scalar) or a boolean array

astroplan/tests/test_core.py
adrn: I agree with @eteq .

astroplan/core.py
```diff
@@ -35,13 +33,52 @@
...
+                                    [end + 1/(N-1)]])*u.day
+    else:
+        time_grid = np.linspace(start, end, N)*u.day
+
+    return t0 + time_grid
 
 class Observer(object):
     """
     A container class for information about an observer's location and
     environment.
```
adrn: We need to put a docstring in here for the class, but this could be done in a separate PR...

adrn:
I have a few minor inline comments above, but a larger issue is that I think there should be many more tests. The tests right now are largely functional tests of end-to-end calculations and comparison with pyephem. It would be good to add a bunch of unit tests (as @cdeil [hinted at in a previous comment](https://github.com/astroplanners/astroplan/pull/14#issuecomment-117356802)) -- e.g., testing your "internal" functions like `_horiz_cross` on many different possible inputs, and checking that it fails where you expect. Now, normally I would say that this is required before the PR is accepted, but I do agree that we need to find a balance here for moving forward / writing tests / merging code that other stuff depends on...soooooo I guessssss I'd be willing to merge this (after all comments are addressed) and have a separate PR expanding the tests.

astroplan/core.py
bmorris3: @jberlanga, @eteq and I discussed this a bit. We didn't want to imply too much about what this computation is doing. For example, we're going to make an `is_observable` method that applies constraints, but this is just a simple "is this object above the horizon" method, which we thought "can see" describes succinctly and without unnecessary implications.

astroplan/core.py
bmorris3: Good question, I don't know what I was doing there.

astroplan/core.py
adrn: "can see" is just as vague as "is visible" to me -- I mean there is same amount of possibility for misinterpretation.

bmorris3: One reason we really preferred `can_see` is because it's a method on the Observer class, so you could read it as `keck.can_see(Time.now(), vega)`. Otherwise you'd be calling `is_visible` on the observatory.

adrn: I don't know, feels awkward to me, but not sure why. Just throwing out another possibility:
`target.is_visible(observer, time)` instead

adrn:
BTW there were weird test failures (I think travis issues) so I restarted the tests.

bmorris3:
@adrn It looks like travis issues are still happening.

adrn:
Merged, but let's all keep in mind:

edge cases (objects never rise / set)
https://github.com/astroplanners/astroplan/pull/14#issuecomment-117356802
#18
#19

PR #16, opened 2015-07-02
Added plots.py; Added examples for usage of plot_airmass and plot_parallactic

jberlanga:
…allactic.

Also modified .gitignore to ignore .ipynb_checkpoints folders and contents, which contain saved states for IPython (Jupyter) notebooks.

@astroplanners/astroplanners-contributors  Please have a look and comment!

~~Note that although some unfinished code for `plot_skyplot` exists in `plots.py`, some errors with that function~~ Some issues will make it necessary to address `plot_skyplot` and examples for its use in another PR--stay tuned!


astroplan/plots/plots.py
```diff
@@ -0,0 +1,294 @@
...
+
+import matplotlib.pyplot as plt
+import numpy as np
+import astropy.units as u
+import random
+
+# Originally authored by Jazmin Berlanga Medina
+# (jazmin.berlanga@gmail.com), with input from
+# Adrian Price-Whelan (email), Erik Tollerud (email),
+# Christoph Deil (email), Eric Jeschke (email) and Brett Morris (email).
```
eteq: Note that we generally do not leave comments of authorship like this: the github changelog is used to keep track of who contributed to what.  That makes it much easier to keep it up-to-date: no one has to change text like this.  To give credit for something like "maintainer" or "main developer" like you would be for this, text belongs better in actual visible documentation where *users* will actually see it.

astroplan/plots/plots.py
```diff
@@ -0,0 +1,294 @@
...
+
+    observer: `~astroplan`
+        The person, telescope, observatory, etc. doing the observing.
+
+    time : `~astropy.time.Time`
+        An collection of times to use in generating data to plot.
+        Can be scalar (len(time)=1) or contain multiple times.
+
+    ax : `~matplotlib.axes.Axes` or None
+        Optional.
```
eteq: Not enough info here.  You need to add something like: "The axes this airmass plot will be added to, or if None a new axes will be created."  Also, the "Optional" part should be in the type line.  E.g. 
```
ax : `~matplotlib.axes.Axes` or None, optional 
    The axes this airmass plot will be added to, or if None a new axes will be created.
```

Ditto for all the `ax` sections below.

astroplan/plots/plots.py
```diff
@@ -0,0 +1,294 @@
...
+
+    style_kwargs : WHAT TYPE IS DICTIONARY? or None
+        Optional.
+        A dictionary with plotting style options.
```
eteq: This doesn't have enough information.  To provide kwargs of this sort, the user needs to know what matplotlib plotting function this is going into.  So it needs to be something like "A dictionary of keywords that will be passed into pyplot.scatter to set the plotting style"

For generic dictionary-like things you can just call it a "mapping".  Or just "dict" is fine too - it's clear what you mean either way.

 (also same comment as above about "optional").

astroplan/plots/plots.py
```diff
@@ -0,0 +1,294 @@
...
+
+    dark_plot: Boolean.
+        Optional. False (default) or True.
+    """
```
eteq: You need to add a ``Returns`` section that indicates that ``ax`` is returned.  Have a look through some of the Astropy docstrings to see how this is formatted.

astroplan/plots/plots.py
```diff
@@ -0,0 +1,294 @@
...
+
+    # Set up plot axes and style if needed.
+    if ax is None:
+        ax = plt.gca()
+    if style_kwargs is None:
+        # Need to add more style options for ax's with existing plots?
+        style_kwargs = {'linestyle': '-', 'color': 'r'}
```
eteq: There's a slightly cleaner way of dealing with `style_kwargs`.  Something like this:

```python
style_kwargs = dict(style_kwargs) # make a copy so we don't overwrite what the user gives us in case they want it
style_kwargs.setdefault('linestyle', '-')
style_kwargs.setdefault('color', 'r')
```
That way if the user sets color but not linestyle, they still get the settings you have as defaults (or vice versa).

Note that in this approach the default of `None` is incalid, but an empty dict will do just fine, so you use the empty dict as the default.

astroplan/plots/plots.py
```diff
@@ -0,0 +1,294 @@
...
+    #if dark_plot is True:
+        # Set dark plot style stuff here.
```
eteq: for now you might just raise a `NotImplementedError` if dark_plot is True

astroplan/plots/plots.py
```diff
@@ -0,0 +1,294 @@
...
+
+    # Populate time window if needed.
+    if time.isscalar:
+        time = time + np.linspace(-12, 12, 100)*u.hour
+
+    # Calculate airmass
+    altitude = observer.altaz(time, target).alt
+    airmass = (1.0/np.cos(90*u.deg - altitude))
```
eteq: don't do this, it's already in `AltAz`.  Just replace these two lines with:
``airmass = = observer.altaz(time, target).secz``

astroplan/plots/plots.py
```diff
@@ -0,0 +1,294 @@
...
+
+    # Some checks & info for labels.
+    assert len(time) == len(airmass)
+
+    if hasattr(observer, 'name') is False:
+        observer_name = 'No Observer/Location Name'
```
eteq: Maybe instead of this you should just have the observer name not be included? I.e., set ``observer_name=''``

astroplan/plots/plots.py
```diff
@@ -0,0 +1,294 @@
...
+    ----------
+    target: `~astroplan.FixedTarget`
+        The celestial body of interest.
+
+    observer: `~astroplan`
+        The person, telescope, observatory, etc. doing the observing.
+
+    time : `~astropy.time.Time`
+        An collection of times to use in generating data to plot.
+        Can be scalar (len(time)=1) or contain multiple times.
```
eteq: You need to describe better what happens in the scalar case.  This description could be in the main description part of the docstrings, though.

astroplan/plots/plots.py
```diff
@@ -0,0 +1,294 @@
...
+
+    If an ax object is passed in, plots an additional airmass plot on top.
+    Otherwise, creates a new ax object with an airmass plot.
+
+    Parameters
+    ----------
+    target: `~astroplan.FixedTarget`
+        The celestial body of interest.
+
+    observer: `~astroplan`
```
eteq: I think you meant `` `~astroplan.Observer` `` here?

astroplan/plots/plots.py
```diff
@@ -0,0 +1,294 @@
...
+    observe_timezone = 'UTC'
+
+    # Plot data.
+    ax.plot_date(
+        time.plot_date,
+        airmass,
+        linestyle=style_kwargs['linestyle'],
+        color=style_kwargs['color'],
+        label=target_name,
+        )
```
eteq: this isn't using ``style_kwargs`` the way we meant: instead you do something like:
```
ax.plot_date(time.plot_date, airmass, label=target_name, **style_kwargs)
```
The idea is that this gives the user *full* control over the look of the plot: the ``**style_kargs`` but says "treat style_kwargs as a set of keywords to be passed into `plot_date`".

Or, if you follow the pattern I described above you can do:
```
style_kwargs.set_default('label', target_name)
ax.plot_date(time.plot_date, airmass, **style_kwargs)
```
That way the user can override the name if they *want*, but they have to do so explicitly.

astroplan/plots/plots.py
```diff
@@ -0,0 +1,294 @@
...
+    ax.figure.autofmt_xdate()
+    ax.invert_yaxis()
+
+    # Set labels, title, legend, etc.
+    ax.set_ylabel("Airmass")
+    ax.set_xlabel("Hour - "+observe_timezone)
+    ax.set_title("Airmass vs Time | "+observer_name+" | "+observe_date+" "+observe_timezone)
+    ax.legend(shadow=True)
```
eteq: I don't think automatically adding the legend is a good idea: this plot only creates a single plot, so the legend is a bit superfluous.  The user can always call ``ax.legend(...)`` themselves later to add a legend if they want it.

(Same below).

eteq: As an alternative, this could be optional, controlled by a `show_legend` keyword of the function.

astroplan/plots/plots.py
```diff
@@ -0,0 +1,294 @@
...
+
+    # If Time object is scalar, pad time window.
+    if time.isscalar:
+        time = time + np.linspace(-12, 12, 100)*u.hour
+
+    # Calculate parallactic angle.
+    declination = target.dec
+    latitude = observer.location.latitude
+    azimuth = observer.altaz(time, target).az
+    hour_angle = target.ra
```
eteq: this isn't right: RA and HA are not the same thing.  more generally, I think the actual computation of the parallactic angle is something that is going to be on `Observer`, right @bmorris3 ?  So you could just do:
```
parallactic_angle = obs.parallactic_angle(time, target)
```
```

File: astroplan/plots/plots.py [hunk 14]
```diff
@@ -0,0 +1,294 @@
...
+    if ax is None:
+        # axisbg option sets background color within outer circle (plot bounds).
+        # May need to set a color outside of plot bounds.
+        ax = plt.gca(projection='polar', axisbg='0.96')
+    if style_kwargs is None:
+        # Need to add more style options for ax's with existing plots?
+        style_kwargs = {'marker': 'o', 'color': 'b'}
+
+    # Set up figure.
+    dimensions = (9, 9)
```
eteq: You can't do `plt.figure` here because that creates a figure and that defeats the whole point of passing in `ax` above.

eteq: oops, sorry, forgot that this is left-over

astroplan/plots/plots.py
```diff
@@ -0,0 +1,294 @@
...
+    # Set labels, title, legend, etc.
+    ax.set_ylabel("Parallactic Angle")
+    ax.set_xlabel("Hour - "+observe_timezone)
+    ax.set_title("Parallactic Angle vs Time | "+observer_name+" | "+observe_date+" "+observe_timezone)
+    ax.legend(shadow=True)
+
+    return ax
+
+
+def plot_sky(target, observer, time, ax=None, style_kwargs=None):
```
eteq: I think you should just delete this function for now, as it's non-functional from what you said in the comments.  No point in having something broken in a PR - you can always save it in a separate branch and just bring it back later.

eteq:
A great start @jberlanga - I went through and left a bunch of comments for `plot_airmass`.  Many of them also apply to ``plot_parallactic`` but I didn't copy any of the redundant items, so just remember to reproduce whatever is relevant on both.

One style suggestion on the plots that didn't have an obvious place inline: I think it would be better if these plots defaulted to a ``pyplot.plot``-like style.  That is, *line* plots, not points. The way to do this would be to have the ``fmt`` argument of ``plot_date`` default to "-" instead of "o".  You could just do this with ``style_kwargs.setdefault('fmt','-') if you follow my suggestion inline, @jberlanga .

And one general question for consideration later: how do we want to combine  `dark_plot` and `style_kwargs`? Should `dark_plot` just set a different set of defaults that the user can override with `style_kwargs`?

astroplan/plots/plots.py
```diff
@@ -0,0 +1,294 @@
...
+    # Plot data.
+    ax.plot_date(
+        time.plot_date,
+        airmass,
+        linestyle=style_kwargs['linestyle'],
+        color=style_kwargs['color'],
+        label=target_name,
+        )
+    ax.figure.autofmt_xdate()
+    ax.invert_yaxis()
```
eteq: I think you also need to check to see if it's already inverted: you don't want to swap it *back* if the user has already done this (or if there are two back-to-back calls to this function with the same `ax`)

eteq:
A great start @jberlanga - I went through and left a bunch of comments for `plot_airmass`.  Many of them also apply to ``plot_parallactic`` but I didn't copy any of the redundant items, so just remember to reproduce whatever is relevant on both.

One style suggestion on the plots that didn't have an obvious place inline: I think it would be better if these plots defaulted to a ``pyplot.plot``-like style.  That is, *line* plots, not points. The way to do this would be to have the ``fmt`` argument of ``plot_date`` default to "-" instead of "o".  You could just do this with ``style_kwargs.setdefault('fmt','-') if you follow my suggestion inline, @jberlanga .

And one general question for consideration later: how do we want to combine  `dark_plot` and `style_kwargs`? Should `dark_plot` just set a different set of defaults that the user can override with `style_kwargs`?

astroplan/plots/plots.py
jberlanga: Thanks for pointing this out.  I based this file off of `core.py`, which I in turn based off some `.py` files in `astropy/cosmology`--it seemed cumbersome, but I just assumed that's how it was done.

astroplan/plots/plots.py
jberlanga: @eteq I think letting the user call 'ax.legend()` is simpler.

jberlanga:
@eteq I imagined `dark_plot` as triggering a dark background color, light grid line and font label colors, and perhaps disallowing dark plot colors (or at least throwing up a warning to the user if they try to use black plot lines).

eteq:
@jberlanga - ok, that makes sense, but I think you want to give the user the *option* to override that.  In that case it's pretty trivial to do if you follow the `setdefault` pattern I described inline:  you just add more `style_kwargs.setdefault(...)` calls if `dark_plot` is True (and also set the background, which you can just do directly because the user can always change that by messing with `ax`)

jberlanga:

Hi all,

I've included some updates in response to @eteq's suggestions.
Please see updated `plots.py` and example notebooks (in `astroplan/examples`).

examples/plotting_examples--airmass.ipynb
```diff
@@ -0,0 +1,1240 @@
...
+       "6+PdgStsT7f9MHC/pFVr0i0pk9siImKENKXmUK/2rwRuhlc7jQ8AVgCwfbSkjYDj6/ZbgU92dTxL\n",
+       "Whs4htLEdBewa3+d0hERMeua2ucwHCTNZvuVVuejHaQsuqUsuqUsuqUsug21LEbFDGlJ80v6nKSV\n",
+       "gbnra312oox1KYtuKYtuKYtuKYtus1IWbV9zkPR24GeUpqbHgRdsf661uWqNlEW3lEW3lEW3lEW3\n",
+       "WS2L0VBzWAY4xfYHga8Bm0j6JJRqUktz1nwpi24pi24pi24pi26zVBZtV1iSVpC0XsNLqwPPAth+\n",
+       "FPgy8O36fEy3JaYsuqUsuqUsuqUsug17Wdhumx/gIOB+4BLgUGBhYBPg7h7p/gB8rdX5TVmkLFIW\n",
+       "KYt2+BmJsmibmoOkxSn3YFgF+AjwEvAN29cAUyV9tyH5r4Cl6sS6MSdl0S1l0S1l0S1l0W2kyqJt\n",
+       "ggNl5dYNgSVsPwmcCiDp48AewE6S3lrTrgY8aPvFluR05KUsuqUsuqUsuqUsuo1IWbQsOHQNp5I0\n",
+       "uyS5TGo7FdipJrmFsqbSRsAjwDeBHSVdWdNc2/xcj4yURbeURbeURbeURbemlUUL2sY+BawDLNjH\n",
+       "tvdRltl4U32+PnA2sHB9Pg+wdavb91IWKYuURcpirJdF02oOkiZImgJsUz/kjxu2nSjpzcB1lOUx\n",
+       "Pg9g+zpgMWDp+vw52+c1K88jJWXRLWXRLWXRLWXRrWVl0cSo9zbgZ/Xx/JQVWg+
```
eteq: I think you either want this to be ``plt.legend`` instead of ``ax.legend`` (`plt.legend()` will use the current axis, which in this case is the one you want), or you want to set the `ax` variable to what's returned by any of the `plot_airmass` call.

astroplan/plots/plots.py
```diff
@@ -0,0 +1,216 @@
...
+        4) Legend should not be transparent.
+        5) dark_plot option
+
+    Makes an airmass vs. time plot.
+
+    If an ax object is passed in, plots an additional airmass plot on top.
+    Otherwise, creates a new ax object with an airmass plot.
+
+    When a `Time` object with a single instance in time is used (e.g., if
+    len(time) = 1), a new `Time` object is created internally.  This new object
```
eteq: One clarification here: it is *not* correct that a scalar time (which is what you mean here) is the same as one with ``len(time) == 1``.  For example:
```python
>>> Time(['2010-1-1']).isscalar
False
>>> len(Time(['2010-1-1']))
1
>>> Time('2010-1-1').isscalar
True
>>> len(Time('2010-1-1'))
TypeError: Scalar Time object has no len()
```
So in the docs you want to be very clear about this: it's not ``a Time object with a single instance`` but rather ``a Time object with a scalar value (e.g. if `time.isscalar` is true) ``

astroplan/plots/plots.py
```diff
@@ -0,0 +1,216 @@
...
+                 dark_plot=False):
+    """
+    TODO:
+        1) Timezones?
+        2) Limit airmass <= 3.
+        3) Plot lines instead of points.
+        4) Legend should not be transparent.
+        5) dark_plot option
+
+    Makes an airmass vs. time plot.
```
adrn: Nitpicky, but a general rule: it's good to be *a little* more verbose here, e.g., "Plot the airmass as a function of time for a given target."

astroplan/plots/plots.py
adrn: Also, the "sampling of times" are +/- 12 hours right? or is it 24 hours? Whatever it is, specify the size of the window.

astroplan/plots/plots.py
```diff
@@ -0,0 +1,216 @@
...
+    Parameters
+    ----------
+    target : `~astroplan.FixedTarget`
+        The celestial body of interest.
+
+    observer : `~astroplan.Observer`
+        The person, telescope, observatory, etc. doing the observing.
+
+    time : `~astropy.time.Time`
+        Used in generating data to plot.
```
adrn: This is a fluff line, delete!

astroplan/plots/plots.py
```diff
@@ -0,0 +1,216 @@
...
+        Can be scalar (len(time)=1) or contain multiple times.
+
+    ax : `~matplotlib.axes.Axes` or None, optional.
+        The axes this plot will be added to. If none are passed in, new axes
```
adrn: This is not quite right -- how about:

The axes object to draw to. If `None`, use the current axes (`matplotlib.pyplot.gca`).

astroplan/plots/plots.py
```diff
@@ -0,0 +1,216 @@
...
+        will be created.
+
+    style_kwargs : Dictionary or Empty, optional.
```
adrn: Generally specify the class of the objects passed in, so I would make this: `dict or None (optional)`

astroplan/plots/plots.py
```diff
@@ -0,0 +1,216 @@
...
+        A dictionary of keywords passed into `matplotlib.pyplot.plot_date`
+        to set plotting styles.
+
+    dark_plot : Boolean, optional.
```
adrn: `bool (optional)`

astroplan/plots/plots.py
```diff
@@ -0,0 +1,216 @@
...
+    Returns
+    -------
+    ax :  `~matplotlib.axes.Axes`
+        An axes object with added airmass vs. time plot.
+    """
+
+    # Set up plot axes and style if needed.
+    if ax is None:
+        ax = plt.gca()
+    if dark_plot is True:
```
adrn: Changing my mind about how dark plots should be handled -- how about getting rid of this boolean keyword, and (from an API standpoint) having something like:

```python
from astroplan.plots.style import dark_mode
dark_mode.enable()
```

Someone else suggested doing:
```python
from astroplan.plots.style import dark_mode
plot_airmass(..., style_kwargs=dark_mode)
```
but the problem is that `style_kwargs` only gets passed in to the `plot` call, whereas for dark mode we probably want to change the axis line colors, tick colors, figure background, etc.

astroplan/plots/plots.py
```diff
@@ -0,0 +1,216 @@
...
+    if time.isscalar:
+        time = time + np.linspace(-12, 12, 100)*u.hour
+
+    # Calculate airmass
+    airmass = observer.altaz(time, target).secz
+
+    # Some checks & info for labels.
+    assert len(time) == len(airmass)
+
+    if hasattr(observer, 'name') is False:
```
adrn: Minor Python style comment: it's a little more common to see `if not hasattr(observer, 'name')`

astroplan/plots/plots.py
```diff
@@ -0,0 +1,216 @@
...
+        observer_name = ''
+    else:
+        observer_name = observer.name
+
+    if hasattr(target, 'name') is False:
+        target_name = str('%.3f' % (random.random()))
```
adrn: Hm, this probably shouldn't be the default behavior...no solid ideas on the best thing to do here, so for now make it `target_name = ""`

Also `if not hasattr(target, 'name')`

astroplan/plots/plots.py
adrn: Also also, @eteq the user shouldn't really have to care about this though, right? If they pass in `Time(['2010-1-1'])` shouldn't we treat it as a scalar time?

astroplan/plots/plots.py
```diff
@@ -0,0 +1,216 @@
...
+        target_name = target.name
+    style_kwargs.setdefault('label', target_name)
+
+    observe_date = time[0].datetime.strftime('%Y-%m-%d')
+    observe_timezone = 'UTC'
+
+    # Plot data.
+    ax.plot_date(time.plot_date, airmass, **style_kwargs)
+    ax.figure.autofmt_xdate()
+    if plt.ylim()[1] > plt.ylim()[0]:
```
adrn: Probably want to use `ax.get_ylim()`

astroplan/plots/plots.py
```diff
@@ -0,0 +1,216 @@
...
+        ax.invert_yaxis()
+
+    # Set labels, title, legend, etc.
+    ax.set_ylabel("Airmass")
+    ax.set_xlabel("Time - "+observe_timezone)
+    ax.set_title("Airmass vs Time | " + observer_name + " | " + observe_date +
```
adrn: Not sure if we should assume a title or not -- my vote is no (don't set a title). What do others think? @astroplanners/astroplanners-contributors

astroplan/plots/plots.py
```diff
@@ -0,0 +1,216 @@
...
+    # Set up plot axes and style if needed.
+    if ax is None:
+        ax = plt.gca()
+    if dark_plot is True:
+        raise NotImplementedError
+    if style_kwargs is None:
+        style_kwargs = {}
+    style_kwargs = dict(style_kwargs)
+    style_kwargs.setdefault('linestyle', '-')
+    style_kwargs.setdefault('color', 'b')
```
adrn: default color should be black (`k`)


like my soul

astroplan/plots/plots.py
cdeil: +1 to not put a title here ... let users add it if they like.

examples/plotting_examples--airmass.ipynb
jberlanga: `plt.legend` worked perfectly!

astroplan/plots/plots.py
jberlanga: @adrn  In talking with @eteq, we decided for now that something like `Time(['2010-1-1'])` would be not be treated as a scalar.  If the user wants a time window centered on `2010-1-1`, then they must pass in `Time('2010-1-1')`.  If they pass in `Time(['2010-1-1'])`, they will get an airmass plot with only 1 data point.

astroplan/plots/plots.py
jberlanga: @adrn :  @eteq  and I decided to leave dark plots for a future PR, but I agree that having `style_kwargs=dark_mode` doesn't seem appropriate.

cdeil:
@jberlanga – Please rename `astroplan/plots/plots.py` to `astroplan/plots/core.py`. Experience has shown that giving the user the option to do
```python
from astroplan import plots
```
and
```python
from astroplan.plots import plots
```
with eventually different namespace content will lead to confusion, and calling it `core.py` has been the solution chosen in `astropy`. Any other name will do just as well if you prefer, but please not `plots/plots.py`.

astroplan/plots/plots.py
```diff
@@ -0,0 +1,216 @@
...
+
+    time : `~astropy.time.Time`
+        Used in generating data to plot.
+        Can be scalar (or len(time)=1) or contain multiple times.
+
+    ax : `~matplotlib.axes.Axes` or None, optional.
+        The axes this plot will be added to. If none are passed in, new axes
+        will be created.
+
+    style_kwargs : Dictionary or Empty, optional.
```
cdeil: `Dictionary` -> `dict`.
`Empty` -> `None` ?

astroplan/plots/plots.py
```diff
@@ -0,0 +1,216 @@
...
+        A dictionary of keywords passed into `matplotlib.pyplot.plot_date`
+        to set plotting styles.
+
+    dark_plot: Boolean.
```
cdeil: `Boolean` -> `bool` (for Python built-in types, always use the class name, like `dict`, `bool`, `None`, `list`, ...

There should be a space between parameter name and colon:
```python
dark_plot : bool
```
I think without it the HTML docs will not be rendered 100% correctly, but I'm not sure.

astroplan/plots/plots.py
```diff
@@ -0,0 +1,216 @@
...
+    latitude = observer.location.latitude.to(u.rad)
+    azimuth = observer.altaz(time, target).az.to(u.rad)
+
+    # Replace with Astroplan parallactic angle function call.
+    longitude = observer.location.longitude.to(u.rad).value
+    local_sidereal_time = time.sidereal_time('mean', longitude).to(u.rad).value
+    hour_angle = local_sidereal_time - target.ra.to(u.rad).value
+
+    numerator = np.sin(azimuth) * np.cos(latitude)
+    denominator = np.cos(declination) * (np.cos(azimuth)*np.cos(hour_angle) + np.sin(azimuth)*np.sin(hour_angle)*np.sin(latitude))
```
cdeil: This line is too long (I can't read it without scrolling left / right).
Please wrap it, putting the whole expression in parentheses, or use temp variables for sub-expressions (that's what I do, it's just as efficient in Python).

cdeil:
I see you're starting to add IPython notebooks to `examples`.
Please link to them from the Sphinx docs, so that users can easily find / browse them.

jberlanga:
@cdeil  :+1:  to linking to notebooks from Sphinx docs.  It seems straightforward, but since Sphinx never really is for me, I'll make that a separate PR.

jberlanga:

Fresh commit.  Please note that `astroplan/plots/plots.py` has been moved to `astroplan/plots/core.py`, and the IPython example notebooks have been moved to `astroplan/examples/notebooks/`.

astroplan/plots/core.py
```diff
@@ -0,0 +1,223 @@
...
+
+    ax : `~matplotlib.axes.Axes` or None, optional.
+        The axes object to be drawn on.
+        If None, use the current axes (matplotlib.pyplot.gca).
+
+    style_kwargs : dict or Empty, optional.
+        A dictionary of keywords passed into `matplotlib.pyplot.plot_date`
+        to set plotting styles.
+
+    dark_plot : bool, optional.
```
adrn: If we're not going to tackle this in this PR (which I am totally fine with!), please remove the `dark_plot` kwarg since we're not totally sure this will even exist when it is implemented.

astroplan/plots/core.py
```diff
@@ -0,0 +1,223 @@
...
+import matplotlib.pyplot as plt
+import numpy as np
+import astropy.units as u
+import random
+
+
+def plot_airmass(target, observer, time, ax=None, style_kwargs=None,
+                 dark_plot=False):
+    """
+    TODO:
```
adrn: I don't mind having a TODO in the docstring (others might disagree..?) but please put it below the summary and description (just before `Parameters`)

adrn:
@jberlanga I'm pretty happy with this! After you implement the last few (nitpicky) suggestions I had, I think this is ready for final review and merge!

@astroplanners/astroplanners-contributors: please take a look at this soon!

astroplan/plots/core.py
cdeil: +1 (leaving TODOs is OK, but move them down).

cdeil:
@adrn @jberlanga Are the IPython notebooks a temp solution until you get around to writing docs, or do you want them to be part of the docs long-term? Is long-term, we should set up automatic tests somehow, at the very least run the notebooks and check that they don't raise exceptions on travis-ci, maybe there's better tooling by now. Most likely out of scope for this PR, but could you please file a new issue with your thinking how to handle this for `astroplan`?

cdeil:
In https://github.com/jberlanga/astroplan/blob/plots/examples/notebooks/plotting_examples--parallactic_angle.ipynb there's these warnings:
```
/home/jazmin/da_codes/gsoc2015/astroplan-jaz/astroplan/build/lib.linux-x86_64-3.4/astroplan/plots/core.py:201: RuntimeWarning: invalid value encountered in arcsin
  parallactic_angle = np.arcsin(numerator/denominator)
```
I guess they occur all the time in normal usage and should be suppressed?
Probably not to be addressed in this plotting PR, maybe just make a note or file a separate issue?

Otherwise: :+1: to merge!

PS: I'll be pretty much +1 to merge all the time from now on ... I prefer to assemble a full, working `astroplan` by next week or two and then iterate in the remaining two GSoC weeks.

jberlanga:
@cdeil 

>@adrn @jberlanga Are the IPython notebooks a temp solution until you get around to writing docs, or do you want them to be part of the docs long-term? Is long-term, we should set up automatic tests somehow, at the very least run the notebooks and check that they don't raise exceptions on travis-ci, maybe there's better tooling by now. Most likely out of scope for this PR, but could you please file a new issue with your thinking how to handle this for astroplan?

IPython notebooks came up during discussion early on in the summer as a way to show users how to utilize the plotting functions, and they've certainly helped me find some bugs.  I can take the content and put them in the Sphinx docs in another PR if everyone thinks that's better than notebooks long-term.

>In https://github.com/jberlanga/astroplan/blob/plots/examples/notebooks/plotting_examples--parallactic_angle.ipynb there's these warnings:

The warnings *may* be coming from the particular calculation I'm using for parallactic angle.  The plan is to replace mine with the parallactic angle implementation as soon as it's merged, so maybe @bmorris3 can chime in here?

ejeschke:
On the notebook example for ["Plotting airmass for multiple targets"](https://github.com/jberlanga/astroplan/blob/plots/examples/notebooks/plotting_examples--airmass.ipynb) it looks like there is only one plot visible.  This could be because the other target is not visible, but it would be nice (I think) to have an example with 2+ curves visible.

jberlanga:
@ejeschke Thanks for pointing that out!

cdeil:
I'm +1 for mostly Sphinx-based docs and have one IPython notebook showcasing `astroplan` and duplicating some of the most-used functionality at http://www.astropy.org/astropy-tutorials/ and at the end linking to the Sphinx docs.

jberlanga:

I addressed comments after third commit.  

Please note:
* Parallactic angle function that was just merged is NOT used yet.
* IPython example notebooks have been left in for now.

eteq:
@jberlanga - note that it's pretty straightforward to use the new parallactic stuff: just do this:
```
git fetch upstream <instead of "upstream", this may be "origin" or "astroplanners", depending on how you did the initial cloning - same below>
git checkout plots <you might already be on that branch if you're working on it now>
git rebase upstream/master
```
after that your branch should look like the latest version of the astroplan repo, but with your plot stuff added.  So you can then just add another commit that switches to using the new method.  Once you've done that you can just ``git push -f jberlanga plots`` (the ``-f`` means it's ok to overwrite the old stuff, even though you've changed history with the rebase).

eteq:
And re: notebooks, I think it's ok to leave these as a temporary measure, and once you've got the sphinx-gallery stuff in you can move it to there.  They aren't as good of a long-term solution as straight-up examples like that.

One thing to note: there's an ``ipython nbconvert`` tool  (http://ipython.org/ipython-doc/stable/notebook/nbconvert.html) that lets you convert ipython notebooks into python or `.rst` files.  You might find it useful to use that to do the switchover later.

adrn:
Agreed on notebooks. This is looking good to me!

jberlanga:

PR now uses parallactic angle function.  Good to merge?

eteq:
@jberlanga looks good to me except for one small thing:  right now you're using ``ax.set_*`` to do things like setting plot limits, which is good.  The one problem is that (as we discussed yesterday), those do *not* update the plot if you're showing it live (e.g. in a GUI window or the `nbagg` backend).  To do that you need to use the `draw` function.  The best way to do that in an internally consistent manner is to do ``ax.figure.canvas.draw()`` right before you ``return ax`` at the end of each function.

eteq:
Passed and merged.  Yay!!

jberlanga:
sweet!

PR #17, opened 2015-07-09
Parallactic angle implementation

bmorris3:
@astroplanners/astroplanners-contributors I've made a small, separate implementation of parallactic angle, which @eteq realized was missing from the current API.

### DIFF
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -170,9 +170,9 @@ class Observer(object):
         Parameters
         ----------
         time : `~astropy.time.Time`
-            The time at which the observation is taking place.
+            Observation time.
 
-        target : {`~astroplan.FixedTarget`, `~astropy.coordinates.SkyCoord`}
+        target : `~astroplan.FixedTarget` or `~astropy.coordinates.SkyCoord`
             Celestial object of interest.
 
         Returns
@@ -192,7 +192,7 @@ class Observer(object):
             coordinate = target
 
         # Eqn (14.1) of Meeus' Astronomical Algorithms
-        print(self.location.longitude, type(self.location.longitude))
+
         H = (time.sidereal_time('mean', longitude=self.location.longitude)
              - coordinate.ra).radian
         q = np.arctan(np.sin(H) /
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -192,9 +192,8 @@ class Observer(object):
             coordinate = target
 
         # Eqn (14.1) of Meeus' Astronomical Algorithms
-
-        H = (time.sidereal_time('mean', longitude=self.location.longitude)
-             - coordinate.ra).radian
+        LST = time.sidereal_time('mean', longitude=self.location.longitude)
+        H = (LST - coordinate.ra).radian
         q = np.arctan(np.sin(H) /
                       (np.tan(self.location.latitude.radian)*
                        np.cos(coordinate.dec.radian) -
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -192,12 +192,16 @@ class Observer(object):
             coordinate = target
 
         # Eqn (14.1) of Meeus' Astronomical Algorithms
+        print(self.location.longitude, type(self.location.longitude))
         H = (time.sidereal_time('mean', longitude=self.location.longitude)
              - coordinate.ra).radian
-        phi = self.location.latitude.radian
-        delta = coordinate.dec.radian
-        q = np.arctan(np.sin(H)/
-                      (np.tan(phi)*np.cos(delta) - np.sin(delta)*np.cos(H)))
+        q = np.arctan(np.sin(H) /
+                      (np.tan(self.location.latitude.radian)*
+                       np.cos(coordinate.dec.radian) -
+                       np.sin(coordinate.dec.radian)*np.cos(H)))*u.rad
+
+        if q < 0*u.rad:
+            q += np.pi*u.rad
         return Angle(q)
     # Sun-related methods.
 
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -3,11 +3,13 @@ from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
 from astropy.coordinates import (EarthLocation, Latitude, Longitude, SkyCoord,
-                                 AltAz)
+                                 AltAz, Angle)
 import astropy.units as u
 from astropy.units import Quantity
 
 import pytz
+import numpy as np
+from astropy.time import Time
 
 ################################################################################
 # TODO: Temporary solution to IERS tables problems
@@ -161,6 +163,42 @@ class Observer(object):
                 coordinate = target
             return coordinate.transform_to(altaz_frame)
 
+    def parallactic_angle(self, time, target):
+        '''
+        Calculate the parallactic angle.
+
+        Parameters
+        ----------
+        time : `~astropy.time.Time`
+            The time at which the observation is taking place.
+
+        target : {`~astroplan.FixedTarget`, `~astropy.coordinates.SkyCoord`}
+            Celestial object of interest.
+
+        Returns
+        -------
+        `~astropy.coordinates.Angle`
+            Parallactic angle
+        '''
+
+        if not (hasattr(target, 'ra') or hasattr(target, 'dec') or
+                hasattr(target, 'coord')):
+            raise TypeError('The target must be a coordinate (i.e. a '
+                            'FixedTarget or SkyCoord).')
+
+        if hasattr(target, 'coord'):
+            coordinate = target.coord
+        else:
+            coordinate = target
+
+        # Eqn (14.1) of Meeus' Astronomical Algorithms
+        H = (time.sidereal_time('mean', longitude=self.location.longitude)
+             - coordinate.ra).radian
+        phi = self.location.latitude.radian
+        delta = coordinate.dec.radian
+        q = np.arctan(np.sin(H)/
+                      (np.tan(phi)*np.cos(delta) - np.sin(delta)*np.cos(H)))
+        return Angle(q)
     # Sun-related methods.
 
     def noon(self, time):
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -180,6 +180,8 @@ class Observer(object):
         `~astropy.coordinates.Angle`
             Parallactic angle
         '''
+        if not isinstance(time, Time):
+            time = Time(time)
 
         if not (hasattr(target, 'ra') or hasattr(target, 'dec') or
                 hasattr(target, 'coord')):
@@ -199,8 +201,6 @@ class Observer(object):
                        np.cos(coordinate.dec.radian) -
                        np.sin(coordinate.dec.radian)*np.cos(H)))*u.rad
 
-        if q < 0*u.rad:
-            q += np.pi*u.rad
         return Angle(q)
     # Sun-related methods.
 
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -225,7 +225,16 @@ class Observer(object):
         Returns
         -------
         `~astropy.coordinates.Angle`
-            Parallactic angle
+            Parallactic angle.
+
+        Notes
+        -----
+        The parallactic angle is the angle between the great circle that
+        intersects a celestial object and the zenith, and the object's hour
+        circle [1]_.
+
+        .. [1] https://en.wikipedia.org/wiki/Parallactic_angle
+
         '''
         if not isinstance(time, Time):
             time = Time(time)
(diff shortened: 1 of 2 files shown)


astroplan/core.py
```diff
@@ -161,6 +163,46 @@ def altaz(self, time, target=None, obswl=None):
...
+            raise TypeError('The target must be a coordinate (i.e. a '
+                            'FixedTarget or SkyCoord).')
+
+        if hasattr(target, 'coord'):
+            coordinate = target.coord
+        else:
+            coordinate = target
+
+        # Eqn (14.1) of Meeus' Astronomical Algorithms
+        print(self.location.longitude, type(self.location.longitude))
```
cdeil: Remove print

astroplan/tests/test_core.py
```diff
@@ -155,3 +155,34 @@ def test_FixedTarget_ra_dec():
...
+
+    import ephem
+    obs = ephem.Observer()
+    obs.lat = lat
+    obs.lon = lon
+    obs.elevation = elevation
+    obs.date = time.datetime
+    vega = ephem.star('Vega')
+    vega.compute(obs)
+    print(float(vega.parallactic_angle()))
```
cdeil: Add newline at end of file.

astroplan/tests/test_core.py
```diff
@@ -155,3 +155,34 @@ def test_FixedTarget_ra_dec():
                                                         'SkyCoord')
     assert vega.coord.dec == vega_coords.dec == vega.dec, ('Retrieve Dec from '
                                                            'SkyCoord')
+
+def test_pang():
```
cdeil: I'd prefer `test_parallactic_angle` over `test_pang`.
Easier to read and clearer 1:1 correspondence with the function being tested by name.

astroplan/core.py
```diff
@@ -161,6 +163,46 @@ def altaz(self, time, target=None, obswl=None):
...
             return coordinate.transform_to(altaz_frame)
 
+    def parallactic_angle(self, time, target):
+        '''
+        Calculate the parallactic angle.
+
+        Parameters
+        ----------
+        time : `~astropy.time.Time`
+            The time at which the observation is taking place.
```
cdeil: I'd prefer a shorter "Observation time" ... quicker to read and says the same thing. OK?

astroplan/core.py
```diff
@@ -161,6 +163,46 @@ def altaz(self, time, target=None, obswl=None):
...
+
+        target : {`~astroplan.FixedTarget`, `~astropy.coordinates.SkyCoord`}
```
cdeil: The braces `{` are for multiple-choice lists, usually strings like `{"fast", "slow", "turtle"}`
If multiple input types are accepted, remove the braches and add an or between the types (I think that's the common format ... check a few Astropy docstrings or the Numpy docstring guide.)

astroplan/core.py
```diff
@@ -161,6 +163,46 @@ def altaz(self, time, target=None, obswl=None):
...
+                hasattr(target, 'coord')):
+            raise TypeError('The target must be a coordinate (i.e. a '
+                            'FixedTarget or SkyCoord).')
+
+        if hasattr(target, 'coord'):
+            coordinate = target.coord
+        else:
+            coordinate = target
+
+        # Eqn (14.1) of Meeus' Astronomical Algorithms
```
cdeil: Is this reference available freely on the internet? If yes, please link to it so that it's easy to check your implementation.
If no, I guess that's the best you can do.

astroplan/core.py
```diff
@@ -161,6 +163,46 @@ def altaz(self, time, target=None, obswl=None):
                 coordinate = target
             return coordinate.transform_to(altaz_frame)
 
+    def parallactic_angle(self, time, target):
+        '''
+        Calculate the parallactic angle.
```
cdeil: You'll probably laugh at me, but my day job is data analysis and statistics and even though I've learned this in astronomy classes 10 years ago, by now I don't remember what the parallactic angle is.

So I'd be +1 to add a link to wikipedia or something.

astroplan/core.py
bmorris3: D'oh!

astroplan/core.py
bmorris3: It's not freely available. @eteq recommended I make a citation something like this.

cdeil:
The travis-ci build errors are timeouts when conda installing things ... ignore them for now and let's hope they just go away soon with the next commit you push here.

astroplan/core.py
bmorris3: I wouldn't laugh at you – I had to look it up! Is there a common format for listing wikipedia references? Maybe [like this](https://github.com/astropy/astropy/blob/6d0e5b050c5b71be9049ec662143f5b55b2a9a41/astropy/coordinates/baseframe.py#L1094)?

bmorris3:
@cdeil all passing!

astroplan/core.py
cdeil: Looks good to me. And one of the Astropy project leads is [to blame](https://github.com/astropy/astropy/blame/6d0e5b050c5b71be9049ec662143f5b55b2a9a41/astropy/coordinates/baseframe.py#L1094) for that, so it's probably the recommended format.

astroplan/core.py
```diff
@@ -161,6 +163,46 @@ def altaz(self, time, target=None, obswl=None):
...
+        # Eqn (14.1) of Meeus' Astronomical Algorithms
+
+        H = (time.sidereal_time('mean', longitude=self.location.longitude)
+             - coordinate.ra).radian
+        q = np.arctan(np.sin(H) /
+                      (np.tan(self.location.latitude.radian)*
+                       np.cos(coordinate.dec.radian) -
+                       np.sin(coordinate.dec.radian)*np.cos(H)))*u.rad
+
+        if q < 0*u.rad:
```
cdeil: `Angle` has a method to wrap ... I think that's what you're doing here, so calling `wrap` would be more readable code.

astroplan/core.py
```diff
@@ -161,6 +163,46 @@ def altaz(self, time, target=None, obswl=None):
...
+                            'FixedTarget or SkyCoord).')
+
+        if hasattr(target, 'coord'):
+            coordinate = target.coord
+        else:
+            coordinate = target
+
+        # Eqn (14.1) of Meeus' Astronomical Algorithms
+
+        H = (time.sidereal_time('mean', longitude=self.location.longitude)
```
cdeil: Can you split this out in a separate line?
```python
t = time.sidereal_time('mean', longitude=self.location.longitude)
```
I think it's more readable.
(if you disagree, no need to argue, OK to keep as is)

astroplan/core.py
```diff
@@ -161,6 +163,46 @@ def altaz(self, time, target=None, obswl=None):
                 coordinate = target
             return coordinate.transform_to(altaz_frame)
 
+    def parallactic_angle(self, time, target):
```
cdeil: Is it worth splitting out the formula in a standalone function that doesn't depend on `astroplan` objects?
Then call it from this method.

This is done in Astropy e.g. for `separation` and a few other places ... makes re-use easier.
Although if there's no second caller right now, it's not really needed and maybe isn't a good idea.
Just a thought.

astroplan/tests/test_core.py
```diff
@@ -155,3 +155,34 @@ def test_FixedTarget_ra_dec():
...
+def test_parallactic_angle():
+    lat = '+19:00:00'
+    lon = '-155:00:00'
+    elevation = 0.0 * u.m
+    location = EarthLocation.from_geodetic(lon, lat, elevation)
+    time = Time('2015-01-01')
+    vega = SkyCoord('18h36m56.33635s', '+38d47m01.2802s')
+    obs = Observer(name='Observatory', location=location)
+    astroplan_q = obs.parallactic_angle(time, vega)
+    threshold_angle = 2*u.deg
```
cdeil: Is this the precision you expect, or would you expect much smaller difference? (I have no idea here, just a question if this is understood.)

bmorris3: I don't know what to expect, but that seemed good enough for me. This is probably a question for @eteq .

astroplan/core.py
bmorris3: Should this function be a module-level function called `parallactic_angle` also?

astroplan/core.py
bmorris3: I looked into this before I posted the PR and I think what we need to do is subtler than a wrap, because the definition (according to @adrn?) of parallactic angle is bounded on (0,π). `Angle().wrap_at()` [wraps things within 360 deg](https://astropy.readthedocs.org/en/v0.3/api/astropy.coordinates.angles.Angle.html#astropy.coordinates.angles.Angle.wrap_at) intervals, and what I'm doing is flipping things if they're in the outside of (0, 180 degrees). I might be missing something obvious, but I don't think wrap_at can be used to solve this wrap issue.

astroplan/core.py
bmorris3: :+1:

astroplan/core.py
eteq: For e.g. `separation` the functions exsit because sometimes you want to do that stuff on a low-level class *or* a `SkyCoord`.  That is, the same function can be used on multiple types of objects.  Here I think the only thing you'd use it on is `Observer`, right?  So I'd say stick with the principal that "there should be only one obvious way to do it" and just leave it as a method.

astroplan/core.py
eteq: Yeah, I agree it would be better if there was something more easily accessed, but unfortunately I don't know of any other equally complete online text comparable to Meeus.  Meeus is one of the more standard texts on spherical trig, though, so I think it's ok.  You could probably also point to http://www.petermeadows.com/html/parallactic.html but I don't think it's as clear as Meeus (and not quite in the same mathematical form)

astroplan/tests/test_core.py
eteq: Honestly I don't know either, but I'm not sure I would trust PyEphem as an authoritative source, though - less clear to me that that's well-tested/used (it's based on XEphem, and my subjective impression is that that is less likely to be used by people who care about the pang).  Maybe instead one of the observatory-based calculators?  If you google "parallactic angle calculator" you get things from ESO, Keck, and the like (although with various levels of inputs and usability).  Or if there's an IDL code in one of the well-used libraries, that might be good too.

eteq:
This is related to an inline comment but I think it's a bit more general, so I'm leaving it as a comment: in my experience parallactic angle is normally on (-90, 90).  Isn't that the typical range?  It appears this function instead pushes it around to (0, 180)...?

adrn:
@eteq oh duh, 0 crossing is crossing the meridian, right? So why the heck does PyEphem bound at (0,π)???

eteq:
@adrn - well, it's an arbitrary convention really: it's the angle between two great circles so you can either think of it as on (0, 180) or (-90, 90), but I've usually seen the (-90,90) convention (presumably because it's more convenient to have a discontinuity near the ground rather than near the meridian).

bmorris3:
@astroplanners/astroplanners-contributors I've updated the parallactic angle calculator to produce consistent results with two sources: PyEphem (for the convenience) and the [SpeX Parallactic Angle Calculator](http://irtfweb.ifa.hawaii.edu/cgi-bin/spex/parangle.cgi) (for corroboration), in the most recent commit 2f691a8. 

It turns out that I didn't need that [little hard-coded wrap hack](https://github.com/bmorris3/astroplan/commit/2f691a8e5ff53ec029e84cbbb853ce4eee3a7d1b#diff-8237cb1c7bdb65945cb3e8f60519cf85L202) – the Meeus calculation reproduces the results from PyEphem and the SpeX calculator as-is.

@cdeil / @ejeschke – when you're happy merging, so am I (though it looks like I'll have to take care of some merge conflicts. Halp?).

astroplan/core.py
cdeil: OK. If a second use case arises besides as an observer method, we can always split it out into a separate function then.

astroplan/core.py
cdeil: @bmorris3 – Please add the reference explaining what parallactic angle is to the docstring.

cdeil:
Since there seems to have been confusion about the range of angle values returned, would it be useful adding a sentence to the docstring about the range returned by this function?
Or was that just misunderstandings and it's obvious or irrelevant, and simply returning an `Angle` is OK?

bmorris3:
> adding a sentence to the docstring about the range returned by this function

@eteq @adrn would this be useful? I'm not sure that there's actually legitimate confusion apart from our own.

bmorris3:
@cdeil I've added the definition and wikipedia reference in fb07e4d.

adrn:
@bmorris3 I think it was just me who was confused, so up to you :)

bmorris3:
@cdeil I think we're good to go then.

cdeil:
@bmorris3 - please rebase on master

bmorris3:
I think I took care of that in a3c2a85, @cdeil ?

cdeil:
I'm :+1: to merge. But since I don't use or care about parallactic angles, and @eteq and @adrn reviewed this PR, I'll leave the final review / merge to them.

PR #20, opened 2015-07-10
Introducing custom exceptions

bmorris3:
Here are some custom exceptions for astroplan. These first two, `AlwaysUpError` and `NeverUpError` are used when a rise or set calculation is called on a target that does not rise or set.


bmorris3:
@astroplanners/astroplanners-contributors I had a quick discussion with @adrn about an option I was considering, and though I think we came to an agreement, I'll document it here and ask for your opinion.

If I was doing a quick calculation for the rise/set times of a long list of targets, I'd want to be able to get an answer for rise/set times even if some of the targets turned out to be circumpolar or never rise (without an exception killing the script). I know it's dangerous and maybe poor form to add a kwarg that prevents `calc_rise`/`calc_set` from raising an exception in those circumstances, but it's a feature I'd be thankful for personally. 

I think the reason this conversation is null is because our constraints submodule should allow users to determine which targets will rise/set to clean their target list _before_ they call `calc_rise`/`calc_set`. So I haven't implemented this exception killer.

astroplan/exceptions.py
```diff
@@ -0,0 +1,11 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+
+class AlwaysUpError(ValueError):
+    '''Target is circumpolar'''
```
cdeil: @bmorris3 Nitpick comment – always used triple double quotes for docstrings.

This is the Python coding style. It's mentioned in [pep8](https://www.python.org/dev/peps/pep-0008/#documentation-strings), but the command line `pep8` tool doesn't complain about triple single quotes as you use them everywhere.

There's [PEP257](https://www.python.org/dev/peps/pep-0257/) which says 
> For consistency, always use """triple double quotes""" around docstrings. Use r"""raw triple double quotes""" if you use any backslashes in your docstrings. For Unicode docstrings, use u"""Unicode triple-quoted strings""" .

IMO it's nice and important to have a coding style for Python, even if it seems pedantic sometimes.
It means that you don't have to learn a new one for every project you work on, and that devs don't have to argue.

cdeil:
I think I've said this in a comment somewhere else already, but I'm not sure any more.

Numpy / scipy uses warnings and NaN, not exceptions to handle computations for arrays that fail for some elements:
```
>>> import numpy as np
>>> np.sqrt([10, -10, 100])
__main__:1: RuntimeWarning: invalid value encountered in sqrt
array([  3.16227766,          nan,  10.        ])
>>> np.sqrt([10, -10, 100])
array([  3.16227766,          nan,  10.        ])
```
I think that's probably the best solution for us here as well?

adrn:
Right -- forgot about Numpy's convention. Agreed that that's the way to go here!

cdeil:
So if @bmorris3 wants to explore the solution using warnings, how should it be implemented?
Should he subclass `AstropyUserWarning` or look to some example in Scipy / Numpy for how it's done?
http://astropy.readthedocs.org/en/latest/warnings.html#python-warnings-system

bmorris3:
I've corrected the triple quotes in this latest commit. 

@eteq @ejeschke Do you have a preference about how the warnings/exceptions should be done?

ejeschke:
@bmorris3, strongly agree that in an array of targets an exception should not be raised for targets that never rise.  `nan` is a kind of exception placeholder (in some functional programming languages I've heard them called "errons") -- basically an "error value".  There is nothing that says that you have to use 'nan' for this, you can use a unique error value, so long as it can be efficiently tested for with a vector test (`if` involving an array test against an erron scalar) and so easily removed or masked out from the result set.

eteq:
I see the use case now for using `NaN`s instead of exceptions here.  So I'm ok with the solution of returning NaNs, but with warnings.  I worry a bit because in my experience users rarely dead with NaNs correctly, but I don't see any better option here.

@bmorris3 - To implement warnings @cdeil had the right idea - basically what you did here, but subclass from `AstropyWarning` (*not* `AstropyUserWarning - I think the docs are a bit misleading there by implying `AstropyWarning` doesn't exist).  Also note that there's a useful feature of py.test for testing if warnings are raised: https://pytest.org/latest/recwarn.html .  That's by far the easiest way to check warnings within tests, in my experience.

bmorris3:
@astroplanners/astroplanners-contributors The most recent update returns `None` and a warning instead of raising an error, and uses `AlwaysUpWarning` and `NeverUpWarning`, which are subclasses of `AstropyWarning`.

adrn:
Cool! This is ok, but I would instead expect it to return `np.nan`'s wherever the search fails instead. The advantage is that `nan` is a valid numeric value, so it should be trivial then to extend this to work for numpy arrays / quantity arrays. For example, imagine you want to know the rise times for 128 objects, but the first 4 of them never rise -- to me it'd make sense to get back a `Time` array that looks like:

```python
Time([nan, nan, nan, nan, <valid time>, <valid time>, ...etc])
```

buuuttt it looks like `astropy.time.Time` barfs if you give it a `nan` value...so not sure what to do about that:

```python
>>> from astropy.time import Time
>>> Time([2450324.232, np.nan], format='jd')
...
ValueError: Input values did not match the format class jd
```

bmorris3:
@adrn: got it! Implemented in 2622dc9.

astroplan/exceptions.py
```diff
@@ -0,0 +1,13 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+
+from astropy.utils.exceptions import AstropyWarning
+
+class AlwaysUpWarning(AstropyWarning):
```
cdeil: I'd prefer something more verbose than `AlwaysUpWarning` that includes the word `Target`.
E.g. `TargetAlwaysUpWarning` or `TargetNeverSetsWarning`, because it occurs solely when the user asks when the target sets.

astroplan/exceptions.py
```diff
@@ -0,0 +1,13 @@
...
+    """Target is circumpolar"""
+    pass
+
+class NeverUpWarning(AstropyWarning):
+    """Target never rises above horizon"""
+    pass
```
cdeil: Add newline at end of file.

cdeil:
I think we said that we'd start with a flat namespace for `astroplan`.
So should these warnings be imported into the top-level namespace, i.e.
```python
from astroplan import TargetNeverSetsWarning
```

At the moment the user has to say
```python
from astroplan.exceptions import TargetNeverSetsWarning
```
and we have to add a separate "Reference / API" section to the Sphinx docs for these to show up in the docs.

I'm +1 to stick with the flat namespace, including these warnings.

bmorris3:
@cdeil In 7109eba I've added `Target*` to the warning names and flatten the namespace.

astroplan/__init__.py
```diff
@@ -16,3 +16,6 @@
 
     # Import everything in astroplan.core
     from .core import *
+
+    # Import everything in astroplan.exceptions
```
cdeil: Could you please remove this comment and the similar one above that was there before?
(It just repeats what the code says (and is not 100% accurate as-is))

cdeil:
I think this is ready to merge.

Anyone else wants to do a final review?

adrn:
All looks good to me

ejeschke:
I'm fine.

PR #22, opened 2015-07-13
Timezone convenience methods

bmorris3:
Here are some convenience methods for switching between timezones when using the `Observer` class, requested by @eteq and @ejeschke. On Slack, @eteq gave the following mini example API:

```python
>>> obs = Observer(…,timezone=pytz.timezone('US/Eastern'),…)
>>> t = Time(57100.3, format='mjd')
>>> obs.time_to_local_time(t)
datetime.datetime(2015, 3, 19, 3, 12, tzinfo=<DstTzInfo 'US/Eastern' EDT-1 day, 20:00:00 DST>)
>>> dt = datetime.datetime(2015, 3, 19, 3, 12) #this is naive
>>> obs.local_time_to_time(dt)
<Time object: scale='utc' format='datetime' value=2015-03-19 07:12:00>
>>> dt = datetime.datetime(2015, 3, 19, 3, 12, tzinfo=pytz.timezone('US/Pacific'))
>>> obs.local_time_to_time(dt)
ValueError: timezone of input and Observer.timezone don’t match!
```
which was the template for this PR.


ejeschke:
@bmorris3, :+1: for adding something like this back in.  Looking at your example, the `time_to_local_time` and `local_time_to_time` are not perfect inverses, however, in that `time_to_local_time` returns a `datetime.datetime` object while `local_time_to_time` returns an astropy `Time` object.  Won't this be confusing to users?  Since most of our API that accepts "date/time" objects accept `datetime` compatible ones, how about just returning that instead?

I also think the name could be shorter, because I think this will be used frequently, but this is not a big issue.

bmorris3:
@ejeschke I think the names that I actually implemented in the PR – `astropy_to_local_time` and `local_to_astropy_time` – are a little clearer (to me) than the ones in the example above, and make the type of the returned time more intuitive. 

I agree that the names are long though. If you have more succinct ideas let me know!

bmorris3:
@ejeschke @eteq @adrn I've made some updates to these methods and their docs, and wrote up [this iPython notebook describing their intended functionality](https://github.com/bmorris3/gsoc2015/blob/master/timezones.ipynb) to give you a sample of what they are and how they can be used. I'd love your feedback!

cdeil:
At the moment we have zero high-level docs for astroplan and GSoC will end in ~ 1 month.
I'm +1 for @bmorris3 and @jberlanga to start writing high-level docs and add `Examples` sections to the docstrings. I don't mean a lot of text, mainly examples that show how to do common things, like the IPython notebook @bmorris3  linked to in the last comment.

@astroplanners/astroplanners-contributors Start writing docs iteratively now or reserve the last one or two weeks exclusively for this?

ejeschke:
I think it depends on how you view notebooks as "high level docs".  I think if we consider the notebooks then there is more than it would appear otherwise.

I think I'd rather see a minimally complete toolbox with notebooks than use any of the time for high level docs.  We still have no constraints implemented and vectorization is not yet complete.  Also, I'd like to see the polar plot completed, with examples.  My vote would be to make an agreement at the next large group hangout on the items to complete, and any ideas or plans post-GSOC.

astroplan/core.py
```diff
@@ -158,6 +157,52 @@ def __init__(self, location=None, timezone='UTC', name=None, latitude=None,
...
                             'instance of datetime.tzinfo')
 
+    def astropy_to_local_time(self, astropy_time):
+        '''
+        Convert the `~astropy.time.Time` object ``astropy_time`` to a
+        localized `~datetime.datetime` object.
+
+        Parameters
+        ----------
+        time : `~astropy.time.Time`
```
eteq: I think this name is supposed to be `astropy_time`, right?

astroplan/core.py
```diff
@@ -158,6 +157,52 @@ def __init__(self, location=None, timezone='UTC', name=None, latitude=None,
             raise TypeError('timezone keyword should be a string, or an '
                             'instance of datetime.tzinfo')
 
+    def astropy_to_local_time(self, astropy_time):
+        '''
+        Convert the `~astropy.time.Time` object ``astropy_time`` to a
```
eteq: You might want to add somewhere in this docstring that it's localized via `pytz`. Although it might fit better as a high-level statement somewhere in the narrative docs (although they don't exist yet...)

astroplan/core.py
```diff
@@ -158,6 +157,52 @@ def __init__(self, location=None, timezone='UTC', name=None, latitude=None,
...
+        Convert the `~datetime.datetime` object ``date_time`` to a
+        `~astropy.time.Time` object.
+
+        If the ``date_time`` is naive, the implied timezone is the one set by
+        the ``timezone`` keyword argument of the `~astroplan.Observer`
+        constructor.
+
+        Parameters
+        ----------
+        time : `~datetime.datetime`
```
eteq: Similarly, I think you want `date_time` here.

astroplan/core.py
```diff
@@ -158,6 +157,52 @@ def __init__(self, location=None, timezone='UTC', name=None, latitude=None,
...
+        # Convert UTC to local timezone
+        return self.timezone.normalize(utc_datetime)
+
+    def local_to_astropy_time(self, date_time):
+        '''
+        Convert the `~datetime.datetime` object ``date_time`` to a
+        `~astropy.time.Time` object.
+
+        If the ``date_time`` is naive, the implied timezone is the one set by
+        the ``timezone`` keyword argument of the `~astroplan.Observer`
```
eteq: To be clear here you might instead say "the `timezone` structure of `self`" or something like that. That is, the user might have changed it since they constructed the object, so what really matters is the *current* value.

eteq:
Left some clarification comments, but I like this overall.

Two larger scale items:
* I see you removed the error for when the user passes in datetime with a different timezone  than the `Observer`. Can you clarify why? 
* Did you test this on `Time` objects that are not scalar? I would think we'd want to return a list of datetime objects in that case. The simple way to do this is to just do something like ``if not astropy_time.isscalar:;    return [self.astropy_to_local_time(t) for t in astropy_time]``. Similar if a list of `datetime` are passed into the other method

eteq:
Oh and I agree with @bmorris3 that the current names make it fairly clear that the purpose of explicitly for switching between astropy `Time` objects and localized `datetime`s. 

On the docs question, I mostly agree with @ejeschke, but also think it makes more sense to discuss this next week in our meeting. (Keep in mind also that IPython includes some neat tools to convert notebooks into rst docs.)

astroplan/core.py
bmorris3: Would you like that? I was using `time` only once because it's already a long name, but I can stick it in.

astroplan/core.py
bmorris3: How about `astropy_time_to_datetime` and `datetime_to_astropy_time`?

astroplan/core.py
eteq: Oh, I meant in the parameters part of the docstring, not the method name. I like the name!

astroplan/core.py
eteq: As I said above I was talking about the docstring... But now that you mention it, I like this suggestion!

bmorris3:
@eteq :

> I see you removed the error for when the user passes in datetime with a different timezone than the Observer.

I removed that error because I wanted the user to be able to pass in aware datetimes from other timezones for calculations as well. In my 
[demo notebook](https://github.com/bmorris3/gsoc2015/blob/master/timezones.ipynb) I use this feature for a remote observer using Subaru from the East Coast, enabling the user to nimbly convert between timezones with whatever timezone matters to them.

> Did you test this on Time objects that are not scalar?

Nope! I should do that.

astroplan/core.py
bmorris3: Ah, got it.

bmorris3:
@eteq : List-like input now accepted!

bmorris3:
Question from the Hangout on July 21: @ejeschke would you like an error to be raised if users run `obs.datetime_to_astropy_time(date_time)` on a timezone-aware datetime?

cdeil:
@bmorris3 – I don't know if you've seen this, just in case not, there's a failing test on this PR:
https://travis-ci.org/astroplanners/astroplan/jobs/71426909#L552

eteq:
The error in question is a 3.x vs 2.x error

astroplan/tests/test_core.py
```diff
@@ -786,6 +744,29 @@ def test_TargetNeverUpWarning(recwarn):
...
+    assert (obs.datetime_to_astropy_time(dt).datetime ==
+            datetime.datetime(2015, 3, 19, 7, 12))
+
+    assert (obs.astropy_time_to_datetime(obs.datetime_to_astropy_time(dt)).replace(
+            tzinfo=None) == dt)
+
+    # Test vectors of times:
+    times = t + np.linspace(0, 24, 10)*u.hour
+    times_dt = times.datetime
+    assert all((obs.datetime_to_astropy_time(times_dt)).jd ==
```
eteq: Very minor: you *might* want to also try it with ``list(times_dt)`` - in this test right now ``times_dt`` is an array of datetimes, but users may also pass in python lists. And as we've discussed, there are occasional subtle differences bewteen the two...

eteq:
Other than the one minor comment above (which you can address or ignore as you see fit, @bmorris3), and a vote from @ejeschke on the question of raising an error for a non-naive `datetime`, I think this is good to go.

ejeschke:
I think the purpose of this PR has veered a bit away from what I was thinking originally, which was to have a succinct way to create astroplan-compatible time objects in the observer's timezone.  So I don't care so much whether it raises an exception in this case or not. @bmorris3, why don't you decide this.

eteq:
Ok, @bmmoris3, feel free to choose whichever way you want now, and either address my comment above or choose to ignore it.


 Either way just leave s note here when you're done, and if the tests are passing after that I'll go ahead and merge this.

bmorris3:
@eteq :  We're good to go!

eteq:
Awesome - merging!

bmorris3:
@ejeschke / @eteq : I'm scanning through the 0.1 milestone blockers and came upon this. I think this PR satisfied Issue #46. If it did, let's close #46!

ejeschke:
@bmorris3, no it doesn't.  That's why I filed a new issue #46!

bmorris3:
@ejeschke Gah, I thought this was an older issue. Sorry.

Issue #12, opened 2015-06-23
Add a Sphinx plot gallery


cdeil:
@jberlanga, @adrn, all – What do you think about adding a sphinx plot gallery to the `astroplan` docs.
I haven't used this, but I think it's very easy to do:
http://sphinx-gallery.readthedocs.org/en/latest/auto_examples/index.html#examples-index

I wasn't sure if it's a good idea at all, and if so it's probably for later (although it could just as well be set up now), so I made a separate issue from #7 for discussion ...

jberlanga:
@astroplanners/astroplanners-contributors 

I *really* like this idea. :+1:

bmorris3:
:clap: :+1:

adrn:
:+1:

cdeil:
Sounds like we'll add a Sphinx gallery to `astroplan`.
Setting this up should be done in an independent PR from any PR implementing actual plots.
It should just contain one dummy plot for now and then we should see that this works locally, on travis-ci and readthedocs.

@jberlanga – Is this something you're interested in doing in the coming weeks or should I give it a try?

eteq:
:+1: from me too, and I think it fits well with the other stuff @jberlanga will be working on (but it's her call).

jberlanga:
@cdeil  Yeah, I can do this after I get the plotting code all straightened out (hopefully w/in the next week).

cdeil:
@jberlanga – Thanks!
I've assigned this issue to you.

PR #23, opened 2015-07-14
Add Sphinx Plot Gallery

jberlanga:
This PR addresses Issue #12.


cdeil:
@jberlanga – Thanks!

I see you have `astroplan/plots/plots.py` in this PR. Is that on purpose? Do you think code review for that will be quick (~ 1 day)? If no, I'd suggest you split the addition of the gallery (maybe with a dummy plot) from the PR where your plotting code is discussed.

cdeil:
https://travis-ci.org/astroplanners/astroplan/jobs/70927747#L870
```
ImportError: No module named sphinxgallery
```
You probably have to add a line after this one:
https://github.com/astroplanners/astroplan/blob/master/.travis.yml#L101
```
 - if [[ $SETUP_CMD == build_sphinx* ]]; then $PIP_INSTALL sphinxgallery; fi
```

cdeil:
@jberlanga – Is this working locally for you? Do you have a place where you can upload the generated html docs for easy review / discussion? (if no, maybe make a screenshot and paste it here?)

jberlanga:
@cdeil 

1) I did add plots.py here as a way to use the plotting function directly (since the plotting PR isn't merged yet), but I can delete it and replace the (non-functioning) plot with a dummy. 
2) Ah, yes, I had to install sphinx-gallery, good point!
3) It's working locally for me in the sense that a gallery is being added (even though the plot itself isn't showing up).   

Screenshots:

Astroplan RTD Homepage (see "General Documentation")
screenshot from 2015-07-14 13 12 54

Astroplan Plotting Examples
screenshot from 2015-07-14 13 12 34

jberlanga:
@cdeil   I addressed the sphinx-gallery pip install, as well as another dependency (pillow).  The Travis CI build is still failing on build_sphinx.

cdeil:
@jberlanga – Can you reproduce (and fix) the Sphinx warning locally?
https://travis-ci.org/astroplanners/astroplan/jobs/70949409#L877
Use `python setup.py build_sphinx --clean-docs` to make sure you do a clean docs build. 

>  plot itself isn't showing up

Can you try and investigate / fix this?
If there's no warning / error that gives an indication, one thing you could try is starting from a working example (https://github.com/sphinx-gallery/sphinx-gallery) and then trying to spot the difference or make our docs build more like theirs, or theirs more like ours.
Leave a comment if you're not getting anywhere with this and I'll have a look tomorrow morning...

cdeil:
@jberlanga – There's one unrelated issue: the API entries (like e.g. `Target`) are showing up at the top level on the navigation bar on the left (see your screenshot above). Instead there should just be one "Reference / API" entry as it's the case e.g. at http://wcsaxes.readthedocs.org/en/latest/ .

Could you please try to fix this here (e.g. by comparing out `index.rst` file to the one from wcsaxes) and if it's not clear how to fix it, file a separate issue for that?

eteq:
@jberlanga - one thing I noticed about why you're not getting the plot output: if you read http://sphinx-gallery.readthedocs.org/en/latest/getting_started.html#setting-up-your-project very closely, you'll see it says it only actually runs code that starts with `plot_`.  So If I grab this branch and *locally* build it, it actually tries to execute the plot if you rename `airmass.py` to `plot_airmass.py`.

jberlanga:
@cdeil 

>@jberlanga – There's one unrelated issue: the API entries (like e.g. Target) are showing up at the top level on the navigation bar on the left (see your screenshot above). Instead there should just be one "Reference / API" entry as it's the case e.g. at http://wcsaxes.readthedocs.org/en/latest/ .

>Could you please try to fix this here (e.g. by comparing out index.rst file to the one from wcsaxes) and if it's not clear how to fix it, file a separate issue for that?

I have a good idea of how to (start to) fix it, but I'd rather keep that separate from the Sphinx-Gallery stuff, so I'll make an issue/separate PR for this.

@eteq

>@jberlanga - one thing I noticed about why you're not getting the plot output: if you read http://sphinx-gallery.readthedocs.org/en/latest/getting_started.html#setting-up-your-project very closely, you'll see it says it only actually runs code that starts with plot_. So If I grab this branch and locally build it, it actually tries to execute the plot if you rename airmass.py to plot_airmass.py.

I do believe you are correct!

jberlanga:
Plot is now showing up, but is getting cut off on the right.

cdeil:
@jberlanga – I know I said I'd try to help with this today, but I'm very busy, so I'll leave this to you for now to try and resolve the Sphinx warnings / error. But if one of them won't budge, please leave a comment here and I'll have a look.

jberlanga:
@cdeil  After playing around with Sphinx Gallery some more, I'm leaning towards using Matplotlib's [sphinxext](http://matplotlib.org/sampledoc/extensions.html) instead.  Sphinx Gallery makes it harder to control exactly how your documentation looks, and doesn't link the gallery page to the navigation bar (you could add a page listed in the navigation bar which contains a link to the gallery page, but that would be more confusing to users trying to find the gallery).

cdeil:
@jberlanga How do you want to use http://matplotlib.org/sampledoc/extensions.html ?
If you mean just use the `.. plot::` directive in various parts of the docs, then yes, we should do that.
I don't think it contains something to build a gallery, i.e. a page with thumbnail plots and short descriptions, that are clickable and lead to new pages with code and plots.

I think having a Gallery for astroplan would be nice. It's very popular, even for a more number-crunching-oriented package like http://scikit-learn.org/stable/auto_examples/index.html

I didn't fully understand the issue you described linking the navigation bar and the gallery page.
What I would have expected was a separate page in our docs called "Gallery" that can be reached via an entry called "Gallery" in the navigation bar. Now if that is not easily possible, maybe you can look through the issues and feature requests athttps://github.com/sphinx-gallery/sphinx-gallery/issues and file a question / feature request for this and see if they respond?

eteq:
@cdeil - just to clarify, the point here is that `sphinx-gallery` additional complexity which we don't need if there are a fairly limited number of plots - it's straightforward to just hand-create a gallery.  And then contributors don't need to install sphinx-gallery to be able to generate the docs (not a critical thing, but nice).

Part of the issue we encountered is that it's not that trivial to annotate the plots with sphinx-gallery.  You have to embed it in with the examples.  Whereas using the `matplotlib` stuff you can just have a regular ``rst`` file and then link to it and the generated images from an index page.

cdeil:
OK, fine, let's make it a normal RST page with `.. plot::` directives, then.

I still would like a slick thumbnail gallery with some Javascript action when hovering over each image. But not enough that I'd actually put in time to get it to work the way we like. So fair enough, thanks for trying!

Do you think it's worth filing a feature request with `sphinx-gallery` ... maybe they are just not aware what people want and would be willing to implement it.

eteq:
Yeah, that might be a good idea... It may turn out that the main problem is that they just don't have time/effort to document it fully, but it can't hurt.

It shouldn't be very hard to switch from what @jberlanga is doing back to sphinx-gallery if we want, though: she can just put the RST content as the module docstring in the example, and move the ``.. plot::`` directive out as the actual content.  So there's an easy path if there's more time later.

jberlanga:
@cdeil / @eteq  It looks like they will support multiple galleries AND include those in the navigation bar in their [newest dev version](https://github.com/sphinx-gallery/sphinx-gallery/pull/36#issuecomment-121318328).

eteq:
oh wow, that version looks a ton nicer.  It also seems like you can get what you want with inline commentary using the [notebook styled examples](http://sphinx-gallery-local.readthedocs.org/en/refactor_split/tutorials/plot_notebook.html#sphx-glr-tutorials-plot-notebook-py).  So that suggests to me that you should use the matplotlib approach now but consider switching to sphinx-gallery once they've merged that and done a release.

jberlanga:
Sounds like a plan!

eteq:
This was superceded by #42

PR #25, opened 2015-07-15
Re-ordering Observer kwargs

bmorris3:
Here's a really tiny PR to re-order the kwargs in `Observer` so you can initialize `Observer`s faster. In all of my examples I've had to use the `location` kwarg whenever initializing an example observatory, like this:
```python
obs = Observer(location=EarthLocation(...))
```
Often the observatory name isn't important, so I've reordered the kwargs to reflect the order of importance in the calculations (location, timezone, name), so you don't need to write the kwarg names each time, like this:
```python
obs = Observer(EarthLocation(...))
obs = Observer(EarthLocation(...), 'US/Hawaii')
obs = Observer(EarthLocation(...), 'US/Hawaii', 'Subaru')
```

### DIFF
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -81,7 +81,7 @@ class Observer(object):
     TODO: write this docstring
     """
     @u.quantity_input(elevation=u.m)
-    def __init__(self, location=None, name=None, timezone='UTC', latitude=None,
+    def __init__(self, location=None, timezone='UTC', name=None, latitude=None,
                  longitude=None, elevation=0*u.m, pressure=None,
                  relative_humidity=None, temperature=None, description=None):
         """
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -81,8 +81,8 @@ class Observer(object):
     TODO: write this docstring
     """
     @u.quantity_input(elevation=u.m)
-    def __init__(self, name=None, location=None, latitude=None, longitude=None,
-                 elevation=0*u.m, timezone='UTC', pressure=None,
+    def __init__(self, location=None, name=None, timezone='UTC', latitude=None,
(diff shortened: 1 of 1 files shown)


cdeil:
I'm 50:50 on this change, I don't think it's a real improvement, but also am not attached to the old order.

The real issue is that we have created a very flexible (others might say complex) constructor with a lot of parameters. In Python 3, I'd suggest making the arguments [keyword-only](https://www.python.org/dev/peps/pep-3102/) to force callers to not rely on the order.
Since we support Python 2, I don't think we can do better, except to always call such functions with many arguments with explicit keyword arguments in our code and docs. I guess we're already doing this or tests would have failed? Or is this not well-covered by tests?

So @astroplanners/astroplanners-contributors, everyone please comment :+1:  or :-1:  here soon or say what you think should be done, so that we can merge or reject this and move on ...

adrn:
I'm fine with this :+1:

ejeschke:
:+1: on eventual keyword only parameters.  Not only does it make code at the calling point of a method easier to understand, but it is resistant to breakage when new positional parameters are inserted, deleted or reordered.

The only real problem is that keyword parameters in python 2 have an implied "optional" status, which means that it is not clear by the API that they are required.  I think in python 3 if more code starts using keyword-only this idea will change.

Regarding the current reordering, don't much care, but we should avoid needless reordering in the future to avoid confusing users and/or conflicting with external documentation. Also, there should be a test to catch this.

bmorris3:
> we should avoid needless reordering in the future to avoid confusing users and/or conflicting with external documentation

Agreed, @ejeschke / @cdeil! That's why I wanted to get this PR in early (now) and before 0.1.

PR #26, opened 2015-07-15
Sites

bmorris3:
I've implemented here a simple site importing tool, as the basis for discussion amongst @astroplanners/astroplanners-contributors on what we want our site tool to look like. A user can get an `EarthLocation` object for known observatories with the following syntax: 
```python
>>> from astroplan import get_site, Observer
>>> keck = get_site('Keck')
>>> obs = Observer(location=keck)
>>> print(keck.to_geodetic())
(<Longitude -155.47833333333332 deg>, <Latitude 19.828333333333326 deg>, 
 <Quantity 4160.000000000752 m>)
```
There are several aliases for each observatory, and if the queried name is not recognized, a `KeyError` is raised with an informative error message that suggests fuzzy matches (using `difflib`, thanks @adrn!). For example, searching for `'keck obs'` raises the following error:

```python
In [1]: import astroplan; astroplan.get_site(str('keck obs'))
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-1-7986be699dec> in <module>()
----> 1 import astroplan; astroplan.get_site(str('keck obs'))

astroplan/astroplan/core.py in get_site(site_name)
    126             errmsg = ('Site not in database. `use get_site_names()` to see '
    127                       'available sites.')
--> 128         raise KeyError(errmsg)
    129 
    130     return _site_db[site_name.lower()]

KeyError: u"Site not in database. `use get_site_names()` to see available sites. 
Did you mean: 'keck observatory', 'keck'?"
```


cdeil:
Please rebase against master ...

adrn:
Nice @bmorris3! I have 2 main comments:

1) In the JSON file, can we use numbers to represent the latitude/longitude? That way, if other people want to use the same file they don't have to parse the strings...We can just make it clear that all values are in decimal degrees.
2) This one is pretty dumb, but maybe the suggested, fuzzy-matched names in the exception could be sorted on length? That way lazy people like me don't have to search for the shortest thing to type :) For example, just do `sorted(suggested_site_names, key=lambda x: len(x))`

astroplan/core.py
```diff
@@ -73,6 +75,80 @@ def _generate_24hr_grid(t0, start, end, N, for_deriv=False):
 
     return t0 + time_grid
 
+# Observatory database and list of names:
+_site_db = None
+_site_names = []
+
+def _load_sites():
+    '''
```
cdeil: Use `"""` instead of `'''`.

adrn: (That was going to be my third comment, but you beat me to it :))

cdeil:
Looks good to me.

Can you add some info (maybe to docstrings, maybe to a separate `README` file, maybe the high-level docs) where the list comes from. Is it hand-edited? Fixed? Are pull requests with additions welcome?
If we don't expect it to grow much larger, maybe it would be useful to generate an RST table in the Sphinx build and include that in the docs?
(so that users don't have to call the function that lists the names before knowing what's available)

astroplan/core.py
```diff
@@ -73,6 +75,80 @@ def _generate_24hr_grid(t0, start, end, N, for_deriv=False):
...
+                                   db[site]['latitude'],
+                                   db[site]['elevation'])
+        _site_db[db[site]['name']] = location
+        for alias in db[site]['aliases']:
+            _site_db[alias.lower()] = location
+
+def get_site(site_name):
+    '''
+    Construct an `~astropy.coordinates.EarthLocation` object for a known
+    observatory.
```
eteq: "a known observatory" -> "an observatory in the builtin list."?   

Also, perhaps "construct"->"return"?  "construct" implies to me it's a new object every time, but in this case it's actually the *same* object if you ask for the same name twice.  A subtle distinction, but important if someone wants to do ``is`` comparison or the like.

eteq:
I like it aside from that one minor doc comment.  I didn't know about that `difflib.get_closest_matches` function - that's really useful!

Re: @adrn's comment  about decimal degrees: I don't really mind that these can be provided in whatever form astropy can parse.  That way if someone wants to contribute a new observatory they can just put it in whatever form they like.  Maybe as a middle ground they should all be strings like '12.3456d'?

Along similar lines, perhaps the elevation should be `elevation_meters`?  It's pretty clear in context, but it can sometimes catch future silly mistakes to be explicit (someone might assume feet if they're not thinking/reading carefully...)

One final thought: should there be an `add_site_name` function?  This would just take a name and an `EarthLocation`, check to make sure it's actually an`EarthLocation`, and then set it in the `_site_db`.  That gives users a way to provide custom named sites.  In generally I think it's better to do this so that the builtin sites don't become "special".

adrn:
@eteq Can we please store numbers as numbers? If others want to use a similar file for a web interface, for example, then they don't have to worry about parsing strings in to numbers...

```json
"Palomar": {
    "aliases": [
        "Palomar", 
        "Hale Telescope"
    ], 
    "elevation": 1706, 
    "elevation_unit": "meter", # <-- pass whatever is here directly in to u.Unit(), let it parse
    "latitude": 33.356,
    "latitude_unit": "degree", 
    "longitude": "243.137", 
    "longitude_unit": "degree",
    "name": "The Hale Telescope"
}, 
````

**Comment by bmorris3** [comment]
> One final thought: should there be an add_site_name function?

@eteq : Would this actually change the json file, or just store the site in the current instance of the dictionary?

> can we use numbers to represent the latitude/longitude?

@adrn : I thought about this and chose against it because most of the lists/webpages that I see that share lat/lon of observatories do so on dd:mm:ss format, and that leaves a natural cutoff at realistic precision. I wouldn't know where to truncate the decimal equivalents in all cases to preserve their true accuracy – is there a natural way to do that?

> sorted on length

@adrn I like this idea!

> maybe it would be useful to generate an RST table

@cdeil I think that's a great idea but I'm afraid it will grow.

**Comment by bmorris3** [comment]
(I think I just did my first successful rebase. I may be very wrong, but if I'm right, I'm throwing a party here. :fireworks: :balloon: :tada: )

**Comment by eteq** [comment]
Hooray re: rebase :fireworks: :100: 

> @eteq : Would this actually change the json file, or just store the site in the current instance of the dictionary?

No change to JSON, just a change to the in-python state.  I can imagine future applications where e.g. someone designs a drop-down GUI or something with all the sites where they get names from `get_site_names`.  So they would have their own local storage for the added sites if they wanted to save them, but they'd want to start from our list and add to it.

**Comment by bmorris3** [comment]
Of course, since I celebrated, the rebase was botched somewhere. Trying to recreate some functiontality that was lost now (like the `difflib` magic).

File: astroplan/tests/test_core.py [hunk 3]
```diff
@@ -824,3 +804,7 @@ def test_Observer_altaz(self):
         with self.assertRaises(TypeError):
             obs = Observer(location=EarthLocation(0, 0, 0))
             _ = obs.altaz(Time('2000-01-01 00:00:00'), ['00:00:00','00:00:00'])
+
+    def test_bad_site(self):
+        with self.assertRaises(KeyError):
+            _ = get_site('nonexistent site')
```
cdeil: Add newline at end of file.
What's the point in assigning to `_ =`? Can't you just remove that?
(if yes, do it everywhere)

astroplan/tests/test_core.py
```diff
@@ -786,6 +741,31 @@ def test_TargetNeverUpWarning(recwarn):
...
+    keck = get_site('keck')
+    lon, lat, el = keck.to_geodetic()
+    assert_allclose(lon, -1*Longitude('155:28.7', unit=u.deg).value, atol=0.001)
+    assert_allclose(lat, Latitude('19:49.7', unit=u.deg).to(u.deg).value,
+                    atol=0.001)
+    assert_allclose(el, 4160, atol=1)
+
+    keck = get_site('ctio')
+    lon, lat, el = keck.to_geodetic()
+    assert_allclose(lon, -1*Longitude('70.815', unit=u.deg).value, atol=0.001)
```
cdeil: Is `lon` a `Longitude` or `Angle` object?
If yes, I'd be +1 to call [assert_quantity_allclose](http://astropy.readthedocs.org/en/latest/api/astropy.tests.helper.assert_quantity_allclose.html) here (and in similar cases where you do asserts on quantities.

astroplan/data/observatories.json
```diff
@@ -0,0 +1,277 @@
...
+        "aliases": [
+            "Vainu Bappu Observatory", 
+            "VBO"
+        ], 
+        "elevation": 725, 
+        "latitude": "12d34m35.976s", 
+        "longitude": "78d49m35.76s", 
+        "name": "Vainu Bappu Observatory"
+    }
+}
```
cdeil: Add newline at end of file

eteq:
@bmorris3 -  I did another review, and I agree with @cdeil's comments and have two more:
1. can you change `elevation` to `elevation_meters` or similar? You could also have the elevation be given as a string like ``'725 m'`` and then parse that by splitting into a number and a unit, but I think the `elevation_meters` approach is probably easier.  But we need *some* way of saying the units.
2. Can you move all of this from `core.py` to `sites.py`?  The motivation for this is that it's not exactly "core" functionality, it's better to keep global state like the `_site_db` in a well-focused module, and it will make it easier to move over to astropy later. You can just move the functions and imports over (and the tests), add an `__all__` with these function, and then in `__init__.py` add ``from .sites import *``.

(Also, if we don't go with # 2 for whatever reason, you'll need to update the ``__all__`` in `core.py` to include these functions.)

cdeil:
+1 to splitting this out to a separate `sites.py`.

Looks like just having the file `astroplan/data/observatories.json` is OK, i.e. it's installed!?

eteq:
@cdeil - yes, that's right, it gets installed along side the source code.  There's some magic in the setup helpers that does that, keying on the particular use of the name "data" for the directory.

One other thing I just noticed, though: ``astroplan/data/__init__.py`` does *not* need to exist: ``astroplan/data`` is not an actual source code package (i.e. there are no .py files), so just having the JSON file in it is sufficient.

astroplan/tests/test_core.py
bmorris3: Yes, I can remove that.

astroplan/data/observatories.json
bmorris3: This is standard json output – I'm generating the file with Python's json module, [see this notebook](https://github.com/bmorris3/gsoc2015/blob/master/IRAF-sites.ipynb).

bmorris3:
@cdeil / @eteq : it looks like the [travis build is failing on a sphinx command](https://travis-ci.org/astroplanners/astroplan/builds/72040639). Did reorganizing the sites submodule ruin something?

ejeschke:
If you take out the `__init__.py` then you may need to explicitly list the files to be copied in `setup.py`.  I ran across this problem before.  If there is only one or two files it's no big deal, but becomes ugly if there are a lot of such files.

adrn:
I still feel pretty strongly that latitude / longitude should be numbers, or that we should at least be consistent with how we treat units (e.g., why do we pack units in to the lat/lon definitions as strings, but then have `elevation_meters`?) but since no one else seems to care I'm willing to let it go...

astroplan/data/observatories.json
cdeil: OK then.

cdeil:
@adrn One argument for keeping strings for lon / lat could be that they are manually copied from some reference, and keeping them as-is makes it easier to check for errors in some values later.

But @bmorris3, I think you mentioned that you're generating the JSON file with a script?
In this case +1 to @adrn's request to store lon / lat as numbers.

And I also want some explanation (e.g. in a new README file besides the JSON file or anywhere, really) where this data came from. I'd also be +1 to store the script you're using to generate the JSON file in the astroplan repo somewhere and mention it in the README file. Makes it easier for the next person that wants to update that data file in the future.
There should also be some statement in the user-visible docs where the data came from and whether we're accepting feature or pull requests with additions if people want to add their favourite observatory.

eteq:
@bmorris3 - the travis/sphinx problem is because you don't have an ``__all__`` in `sites.py`.  Without that , ``from .sites import *`` pulls in everything, including all of the imports you have at the top of the module.  So then sphinx tries to document all of those, but of course it gets confused because they aren't part of `astroplan`.  So if you add an ``__all__ = ['get_site', 'get_site_names', 'add_site']``, it should work.

On @adrn's bit about lat/lon/consistency... I was persuaded by @bmorris3's point about correct levels of precision that we should at least allow string-like formatting.  But I see your point about the inconsistency with `elevation_meters`.  Perhaps the thing to do is to allow `elevation` to be a unitful string as well?  This would be fairly straightforward: something like:
```
numberstr, unit = db[site]['elevation'].split()
elevation = u.Quantity(float(numberstr), unit)
``` 
?

Re @cdeil's suggestion: I'm :+1: on including info about where the data comes from - a natural place might be the docstring for the ``sites`` module.  In addition, I suggest adding a key-value pair to each entry called ``"source"`` or something like that, which just takes a string describing where that entry came from.  This wouldn't be used at all in the code, but works as a form of documentation for anyone reading the JSON file.

But I'm -0 on including the generating script - if I understand right it would only be useful for the one particular data source (it's the IRAF database, right?), and JSON's a pretty trivial format to work with.

bmorris3:
Re @eteq : Would the "source" be a URL to the IRAF list for all of these entries? How would you like to cite it? 

Re @ejeschke : 

> If you take out the __init__.py then you may need to explicitly list the files to be copied in setup.py.

@eteq : Do the astropy helpers take care of this for the data file, or should I add it to setup.py?

Re @adrn and @eteq :  

>  But I see your point about the inconsistency with elevation_meters

My priority was to make a file that's human readable, that can be easily parsed by standard astropy routines, which can be copied from any observatory fact sheet with ease. Since elevation is normally listed in meters, it made sense to me to expect an integer and assume meters.

eteq:
FYI, @bmorris3, merging the timezone stuff apparently caused a conflict with this so you'll need to rebase this against master.  Feel free to ping me on slack if you want help with that.

> Re @eteq : Would the "source" be a URL to the IRAF list for all of these entries? How would you like to cite it?

 I'm fine with what you put here ("IRAF observatory database") - I think anyone looking at that would understand it.  

> @eteq : Do the astropy helpers take care of this for the data file, or should I add it to setup.py?

I'm 95% sure the helpers take care of this, but *only* for files in a directory named "data".  Anything else you have to manually list in the ``MANIFEST.in`` file at the root of the repo.  The way to check is you do ``python setup.py sdist``, which creates a ``dist/astroplan-<version>.tar.gz`` file.  You can look in that file and see if what you want made it in.  If I do that right now on your branch it includes the file, but I suggest you check yourself (it's always possible I have something weird set up on my computer...)

> My priority was to make a file that's human readable, that can be easily parsed by standard astropy routines, which can be copied from any observatory fact sheet with ease. Since elevation is normally listed in meters, it made sense to me to expect an integer and assume meters.

I think my solution above is still pretty close to "standard" and is very human readable.  But I have no strong opinion either way - I'm ok with the current solution (@adrn?).

bmorris3:
My `data` dir is not included after manipulating the `MANIFEST.in` file. Do I also need to edit setup.py?

adrn:
@eteq @bmorris3: +0.5 I'm fine with that solution

bmorris3:
@eteq `git clean -dfx` fixed it for me, I now see that `python setup.py sdist` builds the package with the observatories database.

bmorris3:
@eteq: Rebased, passed, and good to go, methinks.

eteq:
Alright, I'm going to go ahead and merge this.  I'm going to add an issue about documenting some of this a bit more in the narrative docs, though (including a table of the builtin observatories)

PR #30, opened 2015-07-16
Added plot_sky function and example notebook.

jberlanga:

Grid lines are still acting funky (see `astroplan/examples/notebooks/` for notebook), so input would be greatly appreciated.


astroplan/plots/core.py
```diff
@@ -0,0 +1,157 @@
...
+        '',
+        ]
+
+    # An alternate way to set ticks and labels.
+    ax.set_rgrids(range(1, 90, 15), r_labels, angle=-45)
+    ax.set_thetagrids(range(0, 360, 45), theta_labels)
+
+    #ax.rgrids(range(90, 0, -15), r_labels)
+    #ax.thetagrids(range(0, 360, 45), theta_labels)
+
```
eteq: need the `ax.figure.canvas` bit we discussed in #16

eteq:
I'm not going to have time to do a detailed read until I get back monday, but my first-level look says this is really good!

astroplan/plots/core.py
jberlanga: @eteq  Done.  Although it didn't help in this case.

jberlanga:

I've updated `plot_sky` to a point where I'm mostly happy.  A few issues still persist:

* I couldn't get the horizon to correspond to the black, outer edge of the plot--hopefully this isn't too confusing to users.
* Users may be confused as to where exactly 60 & 30 degrees are, but it's not clear to me how to easily add extra ticks--perhaps manually with `annotate`?
* Mapping over time doesn't label start/end positions, so users will have to print out the alt/az values OR label points themselves--it's not clear to me how we could automatically do this well.

On a related note (to the last point above)--right now I'm printing below horizon warnings plus alt/az values for checking that `plot_sky` indeed works.  I'd like to leave the horizon warnings in (at least as an option), but planning to take out the alt/az value printing for the final merge.

jberlanga:

Another note--Please see the last section in the IPython notebook example.  I'm getting a weird error when using `plt.legend()` (within the notebook, **not** within the plotting function itself).  I used the exact same command everywhere else, but I only got the error for the particular `Time` object used.  It would be a shame/confusing if users couldn't use a legend with the returned ax object for some weird subset of situations--any ideas what is going on?

jberlanga:
@eteq I forgot to mention yesterday--did you happen to see in the notebook that `annotate()` is only working for north? Any ideas as to why?

jberlanga:
@ejeschke  Could you expand on your comment from yesterday when you said that you wanted to be able to change the position/direction increase of theta & location of N/E/S/W?  I'm looking at your obsplan example [here](https://github.com/ejeschke/obsplan/blob/master/examples/ipython-notebook/obsplan%20example.ipynb) but since it doesn't make sense to me that North is at `180 deg.`, I'm not sure this is an example of something you'd like to be able to do with `plot_sky`.

jberlanga:

I've stripped the sky plots down to be more simple, added some coordinate options and fleshed out the examples in the `plot_sky` notebook.

ejeschke:
@jberlanga, the need for this orientation is probably restricted to a few special use cases, but some telescopes, such as Subaru, use a convention that 0 degrees is in the south, and increase going round to the east.  You can also find this kind of convention in ESO or SDSS FITS files (with sometimes the increase going to the west).  Fortunately this is easily handled with an option to the axes in matplotlib.

ejeschke:
@jberlanga, the orientation for this can be set by the following two calls:
```python
ax.set_theta_zero_location("S")
ax.set_theta_direction(1)
```
Perhaps most users would want "N" and -1, but it would be highly desirable if it can be configured.

jberlanga:
@ejeschke  I went ahead and added [these options](https://github.com/jberlanga/astroplan/blob/plots_part2/examples/notebooks/plotting_examples--sky_plot.ipynb), but it seems like it's not quite what you were thinking.

astroplan/plots/sky.py
```diff
@@ -0,0 +1,179 @@
...
+        'S' + '\n' + '180' + degree_sign,
+        '225' + degree_sign,
+        'W' + '\n' + '270' + degree_sign,
+        '',
+        ]
+
+    # Set ticks and labels.
+    ax.set_rgrids(range(1, 106, 15), r_labels, angle=-45)
+    ax.set_thetagrids(range(0, 360, 45), theta_labels)
+
```
ejeschke: I think you will need to rotate the `theta_labels` for alternative plots (S=0, E/W flipped) if you are going to use this approach, because it is distributing them from 0->360 in +45 increments.  Alternatively, you could specify a different range and/or flip the sign of the intermediate value.

jberlanga: @ejeschke  So my question is (bringing this here from our Slack discussion): I can give users this functionality by splitting up what is now the optional `north` argument into a `north` and an `az_zero` argument, but is this the best way to go about it?

astroplan/plots/sky.py
```diff
@@ -0,0 +1,179 @@
...
+    style_kwargs : dict or Empty, optional.
+        A dictionary of keywords passed into `matplotlib.pyplot.plot_date`
+        to set plotting styles.
+
+    north: str, optional.
+        A string that indicates where to place North on the sky ('top', 'left',
+        'bottom' or 'right').
+
+    inc_az_ccw: bool, optional.
```
eteq: can we rename `inc_az_ccw` to `north_to_east_ccw` or something like that?  I think that's easier to understand at a glance because it's easier to think in cardinal directions than remember which way is positive/negative azimuth (although I could be in the minority here...).

astroplan/plots/sky.py
```diff
@@ -0,0 +1,179 @@
...
+
+    ax : `~matplotlib.axes.Axes` or None, optional.
+        The axes object to be drawn on.
+        If None, use the current axes (`matplotlib.pyplot.gca`).
+
+    style_kwargs : dict or Empty, optional.
+        A dictionary of keywords passed into `matplotlib.pyplot.plot_date`
+        to set plotting styles.
+
+    north: str, optional.
```
eteq: maybe this should also be allowed to be an angle in astropy Quantity for?  I.e., `45*u.deg`?

jberlanga: If I use `[set_theta_offset](http://matplotlib.org/devel/add_new_projection.html#matplotlib.projections.polar.PolarAxes.set_theta_offset)` (which would allow any arbitrary angle), the offset is from the positive x-axis.  As for direction, not sure.

astroplan/plots/sky.py
jberlanga: That's fine with me.

jberlanga:
@ejeschke @eteq   I've updated to include an option for rotating the theta labels (but not the plotting points/theta themselves).   Please see the example notebook.

adrn:
This is looking good! My one comment would be perhaps we want to allow the user to control whether a grid is drawn or not? I'm not sure what the best way to do this is, but the 0th order thing we could do is add a new kwarg, e.g., `grid=True`.

jberlanga:
@adrn  Did just as you suggested.

ejeschke:
@jberlanga, thanks for adding the hooks for the customized azimuth orientation.  Looks good!

jberlanga:
@ejeschke: glad that works for you!

@astroplanners/astroplanners-contributors: I'm going to go ahead and delete the color defaults for `plot_airmass` and `plot_parallactic` in this branch.  After that I think it will be good to merge.

jberlanga:

Have removed default colors in all plots--should be ready to merge.

astroplan/plots/sky.py
```diff
@@ -0,0 +1,195 @@
...
+    time : `~astropy.time.Time`
+        Can be scalar (e.g., Time('2000-1-1')) or not
+        (e.g., Time(['2000-1-1'])).
+
+    ax : `~matplotlib.axes.Axes` or None, optional.
+        The axes object to be drawn on.
+        If None, use the current axes (`matplotlib.pyplot.gca`).
+
+    style_kwargs : dict or Empty, optional.
+        A dictionary of keywords passed into `matplotlib.pyplot.plot_date`
```
eteq: I think this is a typo in that it is passed into `scatter`, not `plot_date`, right?

astroplan/plots/sky.py
```diff
@@ -0,0 +1,195 @@
...
+    # Direction of azimuth increase. Clockwise is -1
+    if north_to_east_ccw is False:
+        ax.set_theta_direction(-1)
+
+    # Plot target coordinates.
+    # ax.plot connects dots.
+    #ax.plot(az_plot, alt_plot, **style_kwargs)
+    # ax.scatter does not connect dots.
+    ax.scatter(az_plot, alt_plot, **style_kwargs)
+    #ax.polar(az_plot, alt_plot, **style_kwargs)
```
eteq: This comment looks like leftover testing code (and also the ax.plot call above), right?  May as well remove it.

astroplan/plots/sky.py
jberlanga: Ah, yes, good catch!

eteq:
@jberlanga - I was trying this out and noticed some odd behavior: ``plots.plot_sky(targ, obs ,[Time('2010-1-1')])`` works just fine, plotting a single point on the sky.  But ``plots.plot_sky(targ, obs ,Time('2010-1-1'))`` fails, with a ``TypeError: 'Quantity' object with a scalar value has no len()``.

I think it looks like you haven't accounted for the scalar-time case.  Perhaps the thing is to do something similar to what you did for `plot_airmass` and `plots.plot_parallactic`?  That is, if ``time.isscalar`` is True, you replace it with +/- 12 hours from that time, with 1-hour spacings?

Other than that and my minor comments above, looks good to me!

astroplan/plots/sky.py
```diff
@@ -0,0 +1,195 @@
...
+        elif chunk == 6:
+            theta_labels.append('W' + '\n' + str(label_angle) + degree_sign)
+        else:
+            theta_labels.append(str(label_angle) + degree_sign)
+
+    # Set ticks and labels.
+    ax.set_rgrids(range(1, 106, 15), r_labels, angle=-45)
+    ax.set_thetagrids(range(0, 360, 45), theta_labels, frac=1.2)
+
+    # Below commands don't seem to work.
```
jberlanga: @eteq: I'd like to leave this leftover code here as a warning to future contributors that these particular commands don't work with the code as is--would that be ok?

eteq: sure, yeah, that makes sense.  The stuff I mentioned above didn't necessarily add information, but the fact that this didn't work is definitely useful to anyone in the future looking at this...

jberlanga:
@eteq: I'll have a look as soon as I can--it may be ready to merge by Friday, def by Monday.

eteq:
sounds good!

eteq:
@jberlanga - it looks like your change here was just to make a scalar input into a list.  That's not the same as what you did for the other plotting functions: there we interpreted scalar as "plot the day around that time".  So I think you should do the same thing here for consistency.  I'd say in this case it makes sense to do 1-hour increments (instead of the 100 points you did before) because it's a scatter plot.  So that would be something like: ``time = time +  np.arange(-12,13)*u.hour`` instead of ``time = [time]``

jberlanga:
@eteq:  Fixed and updated.

Although I'm not a fan of how the sky plots appear when using the 24-hour window (see [notebook here](https://github.com/jberlanga/astroplan/blob/plots_part2/examples/notebooks/plotting_examples--sky_plot.ipynb), scroll down to the bottom where it says "Testing Scalar Inputs" or see image below).  Without somehow labeling which positions were plotted at which times (or connecting "adjacent" positions), the movement of the targets is not clear. 

skychart_scalar_time

eteq:
@jberlanga - I'm not sure I understand what you're saying is ambiguous?  *Everything* rises in the east and sets in the west (at least for `FixedTarget`s)...

Or do you mean you're not sure *when* these happen?  One possible way to address that is to use `annotate` to mark the first and last positions.  Or you could use the `target_rise_time`/`target_set_time` to get the rise/set times and show them on the legend?  That might be useful to add as an *option*... But I think the plots you show here are pretty clear and useful as-is.

@astroplanners/astroplanners-contributors - any other opinions on this (for background, the plots above are what happen when you give a *scalar* time to `plot_sky` - like the other plotting functions, it interprets that as "plot the day surrounding this time".)

jberlanga:
@eteq -- The *times* at which these happen is ambiguous.  This is better explained with the two figures below, which break up the 24 hour period shown in the figures above.

First 12 hours:
skychart_scalar_time_part1

Second 12 hours:
skychart_scalar_time_part2

adrn:
I agree with @eteq that this could be an option, but I worry that the plot could quickly get impossibly busy with so many annotations floating around...I think it is useful even without times -- often you want to just see the track of a target across the sky. If you really need to know where it is at a particular time, you can pass in a length-1 list of times, right?

I also agree that, rather than putting text on the edge of the plot, it might make sense to have the option be to put the rise/set times in the legend just under the target name?

cdeil:
Also, not everything, maybe not even every common thing, has to be an option in our plotting functions.
You can use the docs to show how to add annotations or other things users might want simply with matplotlib calls before or after calling the astroplan plot functions.
I won't have time to look at this PR ... I'm sure you'll find a good balance between adding options or simply documenting how to do something common, without adding an option for it.

eteq:
How about this: if we're all ok with at least the idea that we should let a scalar indicate "show the whole day", we can merge this now so that it's in, and then @jberlanga can either issue a follow-on PR to make that clearer with an annotation option or a documentation example?

(One other thought which could also go in a later PR: you could use a different scatterplot symbol for the "middle" time - that is, the one that the scalar is actually at.  That way you have at least one reference, and the E/W bit resolves the ambiguity about which way time goes...?)

jberlanga:
@eteq @adrn: I'm fine with this PR being merged as is.  

That being said, I've been experimenting with annotations today, and came up with something that solves ambiguities for some cases (but not others, see figures below) by adding an option to label the first and last visible positions for each target.  I can add this option before merging or refine it for a later PR.

Where first/last labels are helpful:
skychart_firstlast_ex1
The first position where the target is visible is marked by an arrow, the last position is marked by a square.
The labels here are unambiguous and point out the direction of the target's movement.  The labels don't include additional text and are faint compared to the rest of the plot so as not to be too distracting.

---

Where first/last labels are not helpful:
skychart_firstlast_ex2
Note that the labels in this example don't clarify the original problem of the plot being misleading to the naive or casual user.  Since both Vega and Deneb appear at 15 positions, which are evenly spaced, a user quickly glancing at this may think that Vega and Deneb appearing above the horizon are simultaneous events, and that these two objects are more or less vertically aligned as they move across the sky.  Altair has 13 dots--again, on casual inspection, it appears to be vertically aligned with Vega and Deneb throughout the entire period of time.  The actual shape of the constellation is not preserved.

adrn:
Hm, I would actually vote (for this PR) to have no annotations, and then I agree that this PR is ready for final review.

astroplan/plots/core.py
```diff
@@ -69,19 +73,21 @@ def plot_airmass(target, observer, time, ax=None, style_kwargs=None):
...
     style_kwargs.setdefault('linestyle', '-')
-    style_kwargs.setdefault('color', 'k')
     style_kwargs.setdefault('fmt', '-')
 
     # Populate time window if needed.
+    time = Time(time)
     if time.isscalar:
         time = time + np.linspace(-12, 12, 100)*u.hour
+    elif len(time) == 1:
+        print("Danger: You have used a non-scalar Time object of length 1, and"
```
adrn: This should be a warning rather than a print statement. @eteq is `astropy.log.warning()` good enough?

astroplan/plots/core.py
```diff
@@ -165,12 +175,16 @@ def plot_parallactic(target, observer, time, ax=None, style_kwargs=None):
...
-    style_kwargs.setdefault('color', 'k')
     style_kwargs.setdefault('fmt', '-')
 
-    # If Time object is scalar, pad time window.
+    # Populate time window if needed.
+    time = Time(time)
     if time.isscalar:
         time = time + np.linspace(-12, 12, 100)*u.hour
+    elif len(time) == 1:
+        print("Danger: You have used a non-scalar Time object of length 1, and"
```
adrn: Should also be a warning.

astroplan/plots/sky.py
```diff
@@ -0,0 +1,204 @@
...
+    # Grab altitude and azimuth from Astroplan objects.
+    # Note that values must be made dimensionless before plotting.
+    # Modifying altitude is easier than inverting r-axis.
+    altitude = (91 * u.deg - observer.altaz(time, target).alt) *(1/u.deg)
+    # Azimuth MUST be given to plot() in radians.
+    azimuth = observer.altaz(time, target).az * (1/u.deg) * (np.pi/180.0)
+
+    # Some checks & info for labels.
+    if not hasattr(target, 'name'):
+        target_name = '?'
```
adrn: I thought we decided to make this a blank string if the input target has no name?

jberlanga: I put a `?` here because of [the warning](https://github.com/astroplanners/astroplan/pull/30/files#diff-01da251e6d4193d7d339fe7f85713f14R127) that is spit out when a user tries to plot positions that are below the horizon.

astroplan/plots/core.py
eteq: better is just this:
```
from warnings import warn # of course you'd put this further up
warn('Danger: ...')
```
or maybe

astroplan/plots/sky.py
eteq: @jberlanga - we definitely want `''` instead of `'?'` - if it's an empty string it does *not* appear in the legend, which is better if there's no name to show.  Instead, you can do this in the part of the code where you're making the warning: ``target_name if target_name else 'unknown name'``.  That will use the name if it's not the empty string (which python evaluates to boolean false), but if it's empty, it will instead be 'unkown name'

astroplan/plots/sky.py
```diff
@@ -0,0 +1,204 @@
...
+    else:
+        target_name = target.name
+    style_kwargs.setdefault('label', target_name)
+
+    # We only want to plot positions above the horizon.
+    az_plot = None
+    for alt in range(0, len(altitude)):
+        if altitude[alt] > 91.0:
+            print("Warning: Target " + str(target_name) +
+                  " is below the horizon at time: " + str(time[alt]))
```
eteq: I suggest the following (note that this will only work as-is if you change `target_name` back to ``''`` as discussed above, and you'll need to import `warn` as mentioned above):
```python
msg = 'Warning: Target "{0}" is below the horizon at time: {1}'
warn(msg.format(target_name if target_name else 'with an unknown name', time[alt]))
```

astroplan/plots/sky.py
eteq: @jberlanga - still need to do this, FYI

ejeschke:
Hmm, I don't quite understand the rationale for interpreting a scalar time as meaning that you want to plot the entire night, but if you want to plot just one point in time you have to pass in a list of one time.  I think this is just counterintuitive.

How about simply passing an extra kwarg that says to interpret a scalar as "to plot the times around it", whereas the normal behavior is the user passes in a scalar time for a single plot point and an array (or list) of times for the times they are interested in?

ejeschke:
Or better yet, simply have another function called `plot_night` that just constructs the array of times and passes it to the regular plot function.

eteq:
@ejeschke - I think the use case for this is more clear when considered in the context of the *other* plots, and this is for consistency.  I do see the benefit of `plot_night` (although it needs a slightly different name), though... We should probably just discuss this in the call, at any rate

jberlanga:
@eteq: Removed 24 hour plot behavior, fixed warnings & other minor edits.

Note that I suppressed the warnings in the updated notebook since they contained personal info, but they are in the code.

eteq:
merged! :clap: 

Note that right after this, in master, the `plots/core.py` got renamed to `time_dependent.py`.

PR #34, opened 2015-07-21
Vectorize calculations for multiple targets where sensible

bmorris3:


astroplan/core.py
```diff
@@ -295,15 +354,14 @@ def parallactic_angle(self, time, target):
         if not isinstance(time, Time):
             time = Time(time)
 
-        if not (hasattr(target, 'ra') or hasattr(target, 'dec') or
-                hasattr(target, 'coord')):
-            raise TypeError('The target must be a coordinate (i.e. a '
-                            'FixedTarget or SkyCoord).')
-
```
ejeschke: If we really must have a `SkyCoord` here (or `Target` with a SkyCoord), then I think an `isinstance` test would be better. i.e.
```python
if not isinstance(target, SkyCoord) or isinstance(target, Target):
     ....
```
Otherwise I understand this a good test for duck-typed objects.

astroplan/tests/test_core.py
```diff
@@ -99,13 +100,50 @@ def test_Observer_altaz():
...
+
+    # Check for single time
+    single_time = times[0]
+    vega_single_time = obs.altaz(single_time, single_target_list).alt
+    assert vega_single_time[0] == vega_alt[0]
+
+    # Check single target input without list
+    vega_no_list = obs.altaz(times, vega).alt
+    assert all(vega_no_list == vega_alt)
+
```
ejeschke: This is a good test, but it would be nice to have one also where the targets are specified using `StaticTarget`.

ejeschke: Does `obs.altaz()` handle the case where the `SkyCoord` (and by extension `StaticTarget`) is a vector object.  I am happy to see that we can pass targets in as a list (or array) like this, because it makes things easy in the expected use case I have in mind, but I thought in our discussions there was some talk of using the fact that SkyCoords are already vectorized.

bmorris3: Note: the fixed celestial object class is `FixedTarget`.

It does handle the case where vector SkyCoords are input – all it does when a list is submitted is turn the list of coordinates into a vector SkyCoord.

ejeschke:
@bmorris3, it looks to me like the basic functionality is there.  I don't have a lot of comments regarding the code style, so maybe @eteq, @adrn or @cdeil can chime in on those particulars.  I'd be in favor of merging soon if this is a blocker to finishing some of the basics on `Constraints`.

adrn:
I need to do a more detailed review of the code, but one point that immediately comes to mind is that I don't think we should be referring to "arrays" as "vectors" -- I understand now what you mean, when I first looked over the code I was confused. Perhaps the trivial change is to just replaced the word `vector` with `array` in the internal functions/methods you use? In some sense, even a single coordinate or target is a "vector," so it just feels a little confusing to refer to collections of targets as vectors. I'm open to other suggestions for naming though...

astroplan/tests/test_core.py
```diff
@@ -648,10 +850,86 @@ def print_pyephem_solar_transit_noon():
...
+    import ephem
+    obs = ephem.Observer()
+    obs.lat = lat
+    obs.lon = lon
+    obs.elevation = elevation
+    obs.date = time.datetime
+    obs.pressure = pressure
+    vega = ephem.FixedBody()
+    vega._ra = ephem.degrees(np.radians(vega_coords.ra.value))
```
cdeil: Can this be simplified?
I didn't look closely, but doesn't `vega_coords.ra.deg` work?

astroplan/tests/test_core.py
```diff
@@ -806,6 +1090,6 @@ def test_Observer_init(self):
             Observer(location=EarthLocation(0, 0, 0), timezone=-6)
 
     def test_Observer_altaz(self):
-        with self.assertRaises(TypeError):
+        with self.assertRaises(ValueError):
```
cdeil: I don't like `unittest.TestCase` ... `pytest` is much nicer.
Let's see if others agree first before you change this.

adrn:
I did a simple test to check the rise time for a list of targets, but was a bit surprised by how long it took -- to do 3 individual targets (in serial), it takes 0.5 seconds, to do the list of 3 targets it takes 45 seconds! [See here](https://github.com/adrn/snippets/blob/master/astroplan%20-%20vectorize%20test.ipynb). Could you look in to this @bmorris3?

ejeschke:
@adrn, which version of astropy are you using?  With 1.0.2 I get an exception traceback on the call with the list (the 3 individual calls are ok).

adrn:
@ejeschke I'm using 1.0.3, but that shouldn't matter -- isn't it a matter of what version of astroplan you use? Did you check out this branch?

adrn:
@bmorris3 looks like the speed is much better, nice. We probably shouldn't worry about this now, but I'm a little surprised that it is (a little bit) slower to pass in the targets as a list vs. doing each target in serial ([see notebook again](https://github.com/adrn/snippets/blob/master/astroplan%20-%20vectorize%20test.ipynb)).

(now using the latest dev version of astropy)

bmorris3:
@adrn The added step occurring for multiple targets is related to the fact that the `SkyCoord` constructor has to parse the inputs. The parser that determines the format of each element when a list is passed to `SkyCoord.__init__` is pretty slow if there are hundreds of targets, so we use a workaround that checks the representation of each coordinate in the list before passing them to the constructor. If the representations allow, we pass them into the `SkyCoord` constructor with appropriate kwargs to dodge the parser and speed things up.

astroplan/core.py
```diff
@@ -341,30 +426,61 @@ def _horiz_cross(self, t, alt, rise_set, horizon=0*u.degree):
         Returns the lower and upper limits on the time and altitudes
         of the horizon crossing.
         """
+        alt = Latitude(alt)
+
+        if len(np.shape(alt)) == 1:
```
adrn: This should be `alt.ndim`

astroplan/core.py
```diff
@@ -341,30 +426,61 @@ def _horiz_cross(self, t, alt, rise_set, horizon=0*u.degree):
...
+            time_inds = []
+            target_inds = []
+            for tgt, tm in zip(old_target_inds, old_time_inds):
+                if tgt not in target_inds:
+                    time_inds.append(tm)
+                    target_inds.append(tgt)
+            target_inds = np.array(target_inds)
+            time_inds = np.array(time_inds)
+
+        times = [t[i:i+2] if not np.isnan(i) else i for i in time_inds]
```
adrn: Hmm, I'm not totally convinced this and the line below are right. I need some time to play around with this some more...

astroplan/core.py
```diff
@@ -589,14 +660,28 @@ def _calc_transit(self, time, target, prev_next, antitransit=False, N=150):
...
-        altitudes = self.altaz(times, target).alt
+        altaz = self.altaz(times, target)
+        if target_is_vector:
+            #d_altitudes = [each_altaz.alt.diff() for each_altaz in altaz]
+            d_altitudes = [each_alt.diff() for each_alt in altaz.alt]
+        else:
+            altitudes = altaz.alt
+            d_altitudes = altitudes.diff()
+
+        print('alts:', [d_altitudes])
```
adrn: Leftover print statement?

astroplan/core.py
```diff
@@ -649,10 +734,20 @@ def calc_rise(self, time, target, which='nearest', horizon=0*u.degree):
                 return previous_rise
 
         if which == 'nearest':
-            if abs(time - previous_rise) < abs(time - next_rise):
-                return previous_rise
+            if _target_is_vector(target):
```
adrn: There is one fringe case in which this will fail: if the user passes in a list of targets, and one of them never rises. Right now this will die I think because at the end you pass all return times in to the `Time` constructor. I think we probably want to "soft fail" in keeping with our conventions elsewhere. Not sure what the best solution is -- one (slightly bad) option is to instead return a list of `Time`'s with a `nan` wedged in if the rise time fails for a given target...@eteq, thoughts?

bmorris3: You're right. Originally I was under the impression that the `Time` initializer would keep a nan without issue, but when you get a `ValueError` when you do. This leads to the question of how best to handle bad times. I think I like the list of `Time`'s with nan's where appropriate sounds OK.

astroplan/core.py
bmorris3: Gah the shame!

astroplan/core.py
eteq: I'm really against sometimes returning a `Time` and sometimes returning a list.  Users will *never* write code that supports that well.  And I think it's generally much better to return a single `Time` object no matter what for broader consistency.  

So as an alternative to `NaN`, What about masked arrays?  Those seem to work with `Time` objects:
```python
>>> np.ma.masked_where([False,True,False], [1,2,3])
masked_array(data = [1 -- 3],
             mask = [False  True False],
       fill_value = 999999)
>>> mjds=np.ma.masked_where([False,True,False], [1,2,3])
>>> t
<Time object: scale='utc' format='mjd' value=[ 1.  2.  3.]>
>>> t=Time(mjds,format='mjd')
>>> t
<Time object: scale='utc' format='mjd' value=[1.0 -- 3.0]>
>>> t.jd
masked_array(data = [2400001.5 -- 2400003.5],
             mask = [False  True False],
       fill_value = 1e+20)
```

bmorris3:
The current version of this PR will not work properly if one of the stars in the `target` list does not rise/set/transit, as `satrapy.time.Time` can not properly handle `nan`/`None`/masked arrays as input. This is beginning to be addressed in [astropy's Issue 4032](https://github.com/astropy/astropy/issues/4032), and I'm debating what to do from here in the mean time. 

@eteq – would you suggest merging this here, after temporarily raising a `NotImplementedError` when there are mixed good/bad input targets? We've gotta move onto constraints!

bmorris3:
In cases when a user asks to calculate a time that can't be computed, like the rise/set time of targets that don't rise or set, we need a placeholder in the `astropy.time.Time` object like `nan` that indicates a bad result. I've updated this PR to use a magic time value, `Time(-999, format='jd')`, to act a placeholder for a `nan` in `astropy.time.Time` objects. 

The magic time solution implemented here can be replaced once astropy/astropy#4032 is addressed. In the mean time, here's an example for how use with the magic time will work in astroplan:

```python
from astropy.time import Time
from astroplan import Observer, FixedTarget, MAGIC_TIME

polaris = FixedTarget.from_name("Polaris")
rigel = FixedTarget.from_name("Rigel")
vega = FixedTarget.from_name("Vega")
targets = [polaris, rigel, vega]

obs = Observer.at_site("CTIO")
set_times = obs.target_set_time(Time.now(), targets)

# Does each target set?
print("\n".join(["{} sets?: {}".format(target.name, t != MAGIC_TIME)
                 for target, t in zip(targets, set_times)]))

# Print time of setting:
print("\n".join(["{} set time: {.iso}".format(target.name, t)
                 for target, t in zip(targets, set_times)]))

set_time_polaris = obs.target_set_time(Time.now(), polaris)
assert set_time_polaris == MAGIC_TIME
```
returns
```
WARNING: TargetNeverUpWarning: Target(s) index [0] does not cross horizon=0.0 deg within 24 hours [astroplan.core]
WARNING: ErfaWarning: ERFA function "utctai" yielded 1 of "dubious year (Note 3)" [astropy._erfa.core]
Polaris sets?: False
Rigel sets?: True
Vega sets?: True
WARNING: ErfaWarning: ERFA function "d2dtf" yielded 1 of "dubious year (Note 5)" [astropy._erfa.core]
Polaris set time: -4715-02-28 12:00:00.000
Rigel set time: 2015-08-05 19:21:36.880
Vega set time: 2015-08-05 06:35:16.983
```

bmorris3:
Can we get towards a final review, maybe? I know @adrn has read this pretty closely, and @eteq has been closely involved and might be ready.

eteq:
@bmorris3 - I read over and the actual machinery looks good!  Two things, though:

1. It looks like the docstrings for all of the "user-level" methods are unchanged - they should probably say explicitly that a list of targets work, right?
2. This is not critical, but presumably you already have it saved somewhere, so if it can be done easily maybe add a function to the testing suite that checks the timing?  This would be most useful as function that we can call, e.g. ``obsplan.tests.vectorize_timing(ntargets)`` that then just returns the time to go to altaz and a rise/set/transit calc or something.  That way anyone can easily go in and check if future changes affect the performance. 
3. I may have missed it, but I don't see a test where you do something like this:
```python
arrtime = observer.something(multipletargets)
validtimes = arrtime[arrtime == MAGIC_TIME]
#or maybe also:
targetsthatrise = multipletargets[arrtime == MAGIC_TIME]
```
that's something people will definitely want, so it'd be good to ensure it continues to work.  It should probably also end up in the docs, but hopefully @jberlanga can use the tests here as a guide for writing some of the docs.

bmorris3:
Updates in fb71017 RE @eteq:
1)  Docstrings have been updated. 
2) I've created a vector-target timing test which you can run with: 
```
$ python -c "from astroplan.tests.test_core import vectorize_timing; print(vectorize_timing(10))"
3.25822901726
```
3) I've added another few lines to the mixed rises/don't rise [test](https://github.com/bmorris3/astroplan/blob/vectorize/astroplan/tests/test_core.py#L1050-L1051) to show that the manual array masking works.

adrn:
@eteq you happy with this?

bmorris3:
@eteq / @cdeil : If you are happy and can merge, I can start tinkering with constraints before tomorrow's meeting.

cdeil:
I didn't review this, so I'll leave this to @eteq.

astroplan/tests/test_core.py
adrn: @bmorris3: you ignoring this comment on purpose?

bmorris3: Yes I'm ignoring this one because it's an optimization of a function that just prints PyEphem results.

astroplan/tests/test_core.py
```diff
@@ -115,8 +221,8 @@ def print_pyephem_altaz(latitude, longitude, elevation, time, pressure,
     pyephem_obs.date = time.datetime
     pyephem_obs.pressure = pressure
     pyephem_target = ephem.FixedBody()
-    pyephem_target._ra = ephem.degrees(np.radians(target_coords.ra.value))
-    pyephem_target._dec = ephem.degrees(np.radians(target_coords.dec.value))
+    pyephem_target._ra = ephem.degrees(target_coords.ra.radian)
+    pyephem_target._dec = ephem.degrees(target_coords.ra.radian)
```
adrn: This should be `dec.radian`

PR #37, opened 2015-07-21
Remove dummy test function

cdeil:
At the start of the project I added a single dummy function and test.
Now that we have real code and tests, it should be removed, which is done in this PR.

@eteq – Can you please have a look and merge if it's OK and travis-ci passes?


eteq:
Will do.

PR #43, opened 2015-07-28
is_night convenience method

bmorris3:
Enhancement: `observer.is_night(time, [horizon])`

Is the sun below specified `horizon` at `time` for an `Observer`? This is a method I was writing for myself when coding up a use case, so I stuck it into `Observer` for everyone.

### DIFF
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -1166,6 +1166,34 @@ class Observer(object):
         else:
             return observable, altaz
 
+    @u.quantity_input(horizon=u.deg)
+    def is_night(self, time, horizon=0*u.deg, obswl=None):
+        """
+        Is the Sun below ``horizon`` at ``time``?
+
+        Parameters
+        ----------
+        time : `~astropy.time.Time` or other (see below)
+            Time of observation. This will be passed in as the first argument to
+            the `~astropy.time.Time` initializer, so it can be anything that
+            `~astropy.time.Time` will accept (including a `~astropy.time.Time`
+            object)
+
+        horizon : `~astropy.units.Quantity` (optional), default = zero degrees
+            Degrees above/below actual horizon to use
+            for calculating day/night (i.e.,
+            -6 deg horizon = civil twilight, etc.)
+
+        obswl : `~astropy.units.Quantity` (optional)
+            Wavelength of the observation used in the calculation.
+        """
+        if not isinstance(time, Time):
+            time = Time(time)
+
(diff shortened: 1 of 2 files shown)


astroplan/core.py
```diff
@@ -1166,6 +1166,34 @@ def can_see(self, time, target, horizon=0*u.degree, return_altaz=False):
...
+
+        horizon : `~astropy.units.Quantity` (optional), default = zero degrees
+            Degrees above/below actual horizon to use
+            for calculating day/night (i.e.,
+            -6 deg horizon = civil twilight, etc.)
+
+        obswl : `~astropy.units.Quantity` (optional)
+            Wavelength of the observation used in the calculation.
+        """
+        if not isinstance(time, Time):
```
adrn: BTW: was originally going to suggest removing this line because passing a `Time` in to `Time` gets you a `Time`, but the returned `Time` is a copy of the input `Time`

`Time`
...
`Time`

bmorris3: Is that a problem? We do this all over `Observer`.

adrn: Is which a problem?

astroplan/core.py
```diff
@@ -1166,6 +1166,34 @@ def can_see(self, time, target, horizon=0*u.degree, return_altaz=False):
         else:
             return observable, altaz
 
+    @u.quantity_input(horizon=u.deg)
+    def is_night(self, time, horizon=0*u.deg, obswl=None):
```
cdeil: How about making it clearer from the docstring what is returned?
Or is it completely obvious and adding a `Returns` section is overkill?

bmorris3:
@cdeil Added!

astroplan/core.py
```diff
@@ -1166,6 +1166,39 @@ def can_see(self, time, target, horizon=0*u.degree, return_altaz=False):
...
+        -------
+        sun_below_horizon : bool
+            `True` if sun is below ``horizon`` at ``time``, else `False`.
+        """
+        if not isinstance(time, Time):
+            time = Time(time)
+
+        solar_altitude = self.altaz(time, target=get_sun(time), obswl=obswl).alt
+        sun_below_horizon = solar_altitude < horizon
+        return sun_below_horizon
```
eteq: this is super minor, but you could just as well do ``return solar_altitude < horizon``, right?

bmorris3: Sure!

PR #45, opened 2015-07-28
Adding Observer.at_site class method

bmorris3:
Making it easier to initialize an `Observer`object at a site via the `sites` module, as per a conversation with @eteq on Slack. 

Now to initialize an observer at Kitt Peak, for example:
```python
from astroplan import Observer
kpno = Observer.at_site("kpno")
```

### DIFF
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -22,8 +22,8 @@ iers.IERS.iers_table = iers.IERS_A.open(download_file(iers.IERS_A_URL,
 
 from astropy.extern.six import string_types
 from .exceptions import TargetNeverUpWarning, TargetAlwaysUpWarning
+from .sites import get_site
 import warnings
-
 
 from abc import ABCMeta, abstractmethod
 
@@ -156,6 +156,28 @@ class Observer(object):
         else:
             raise TypeError('timezone keyword should be a string, or an '
                             'instance of datetime.tzinfo')
+
+    @classmethod
+    def at_site(cls, site_name, **kwargs):
+        """
+        Initialize an `~astroplan.core.Observer` object with a site name.
+
+        Parameters
+        ----------
+        site_name : str
+            Observatory name, must be resolvable with
+            `~astroplan.sites.get_site`
+
+        Returns
+        -------
+        `~astroplan.core.Observer`
+            Observer object.
+        """
+        name = kwargs.pop('name', site_name)
+        if 'location' in kwargs:
+            raise ValueError("Location kwarg should not be used if "
+                             "initializing an Observer with Observer.at_site()")
+        return cls(location=get_site(site_name), name=name, **kwargs)
 
     def astropy_time_to_datetime(self, astropy_time):
         """
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -162,47 +162,15 @@ class Observer(object):
         """
         Initialize an `~astroplan.core.Observer` object with a site name.
 
+        Extra keyword arguments are passed to the `~astroplan.core.Observer`
+        constructor (see `~astroplan.core.Observer` for available keyword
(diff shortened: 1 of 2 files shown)


astroplan/core.py
```diff
@@ -157,6 +157,28 @@ def __init__(self, location=None, timezone='UTC', name=None, latitude=None,
...
+    def at_site(cls, site_name, **kwargs):
+        """
+        Initialize an `~astroplan.core.Observer` object with a site name.
+
+        Parameters
+        ----------
+        site_name : str
+            Observatory name, must be resolvable with
+            `~astroplan.sites.get_site`
+
```
eteq: You should add a mention either in the `Parameters` section or the text above that any additional keywords are passed into this class' constructor.  Maybe also mention in the docs that if you *don't* specifically give a `name`, `site_name` will be used as the name.

astroplan/tests/test_sites.py
```diff
@@ -33,6 +34,12 @@ def test_add_site():
     retrieved_location = get_site(new_site_name)
     assert retrieved_location == new_site_location
 
+def test_Observer_classmethod():
+    site_name = 'kpno'
+    kpno = Observer.at_site(site_name)
+    assert kpno.location == get_site(site_name)
+    assert kpno.name == site_name
```
eteq: you should also test that the ``kwargs`` usage passes through like it's supposed to.

eteq:
looks good to me aside from the two comments above.

bmorris3:
@eteq / @cdeil Could we restart the travis build on fa427ed? One of the docs builds timed out.

cdeil:
I restarted https://travis-ci.org/astroplanners/astroplan/jobs/73399459

bmorris3:
@eteq / @cdeil Looks good!

cdeil:
The function signature is:
```
def at_site(cls, site_name, **kwargs)
```
8 `kwargs` parameters are passed to `get_site`.
Is it a good idea to duplicate those in this function's docstring (location, latitude, ...) or would it be better to avoid the duplication and refer to the `get_site` docstring?
The issue with duplicating it is that it's hard to keep up-to-date, i.e. it's easy to change something about `get_site` and forget updating `Observer.at_site` at the same time.
The pro of duplicating it is convenience.

@bmorris3 – You'll have to maintain this code, so you decide!
If you keep as is, maybe add a "see also" section to the `get_site` docstring pointing to `Observer.at_site`?

bmorris3:
@cdeil The kwargs get passed to the `Observer` constructor, not to `get_site`. 

I like the idea of leaving a note in the docstring to refer to the `Observer.__init__` docstring for the kwargs. How would you word that? Something like "`Observer.at_site` takes all of the keyword arguments accepted by the `Observer` constructor?"

cdeil:
Your wording sounds good to me.
I think I would have written something like:
"Extra keyword arguments are passed to the `~astroplan.Observer` constructor".
Maybe @eteq  or @adrn  know if there's a common pattern to describe this in docstrings in use in Astropy?
Or @bmorris3 , you could look at the wording we use in the plotting functions ... there we also pass kwargs along to matplotlib.

bmorris3:
@cdeil I've used wording like yours in a2c5471.

cdeil:
:+1: 

@eteq Merge or final review?

eteq:
@bmorris3 @cdeil - I think we more often say "initializer" rather than "constructor", but it's all the same thing really, more just personal style.

I did one last review and am happy with this.  I also see the tests are passing so I'll merge this now!

bmorris3:
@eteq: should we consider suggesting this as a starting point for astropy/astropy#3813?

eteq:
Yep, definitely.

PR #48, opened 2015-07-30
Shortening Target docstring, TimeWindow->TimeRange

bmorris3:
Addressing some comments by @cdeil on Slack.

### DIFF
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -1525,14 +1525,14 @@ class AltitudeRange(Constraint):
         raise NotImplementedError
 
 
-class AboveAirmass(Constraint):
+class AirmassRange(Constraint):
     """
-    An object containing an airmass lower limit.
+    Constrain the airmass within a given range.
     """
 
-    def __init__(self, low):
+    def __init__(self, upper_limit, lower_limit=1):
         """
-        Initializes an AboveAirmass object.
+        Initializes an AirmassRange object.
 
         Parameters
         ----------
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -1494,15 +1494,23 @@ class TimeRange(Constraint):
     An object containing start and end times for an observation.
     """
 
-    def __init__(self, start, end):
+    def __init__(self, min=None, max=None):
         """
         Initializes a TimeRange object.
 
         Parameters
         ----------
-        start : STRING OR astropy.time OBJECT ?
+        min : `~astropy.time.Time`, `None` or other (see below)
+            Earliest time. `None` designates no lower limit. This will be passed
+            in as the first argument to the `~astropy.time.Time` initializer,
+            so it can be anything that `~astropy.time.Time` will accept
+            (including a `~astropy.time.Time` object).
 
-        end : STRING OR astropy.time OBJECT ?
+        max : `~astropy.time.Time`, `None` or other (see below)
+            Latest time. `None` designates no upper limit. This will be passed
+            in as the first argument to the `~astropy.time.Time` initializer,
+            so it can be anything that `~astropy.time.Time` will accept
+            (including a `~astropy.time.Time` object).
         """
         raise NotImplementedError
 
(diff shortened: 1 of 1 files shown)


cdeil:
I think there also was a discussion on renaming "AboveAirmass" in the API spec PR.
I'm not sure any more, but I think I was in favour of renaming to "AirmassRange".
What was the conclusion?
Can you link to the relevant comments from here or summarise?

bmorris3:
@cdeil Correct, it should be `AirmassRange`. The conclusion was that we should use `AirmassRange` to be consistent with the other constraint subclasses (`*Range`), and that it may be a special case where it can accept either one argument or two. If a single argument is given, it is assumed to be an upper limit on the airmass with a lower limit of 1, since there is little motivation to set lower limits on airmass.

astroplan/core.py
```diff
@@ -1342,14 +1343,14 @@ def __init__(self, low, high):
...
 
-class AboveAirmass(Constraint):
+class AirmassRange(Constraint):
     """
-    An object containing an airmass lower limit.
+    Constrain the airmass within a given range.
     """
 
-    def __init__(self, low):
+    def __init__(self, upper_limit, lower_limit=1):
```
cdeil: Is this the best solution?
(upper, lower order is confusing).

Wouldn't this be better (None meaning no selection, allowing one-sided cuts):
```python
AirmassRange.__init__(lower_limit=None, upper_limit=None)
```

In any case, please update the docstring to reflect the current parameters.
And I'm +1 to add some more explanation on this one and an `Examples` section that explains common usage.
Probably when we see the examples and explanation it'll also be easier to decide what the best solution for this is.

bmorris3: @eteq @ejeschke Opinions?

ejeschke: Don't have a strong opinion, but I think simply specifying a range with upper and lower is reasonable, and where it defaults to None for either `upper_limit` or `lower_limit` indicating no limit in that direction.

cdeil: Maybe use `min` and `max` instead of `lower_limit` and `upper_limit` for all these min / max range constraints?
It's shorter and just as clear, no?

bmorris3: Maybe `*_min`/`*_max`, i.e. `airmass_min`, `airmass_max`, to avoid built-in name collisions.

cdeil: My motivation for changing to `min` / `max` was to shorten it.
So -1 on something longer and making the user type "airmass" several times:
```python
airmass_constraint = AirmassConstraint(airmass_max=1.42)
```
I think built-in name collissions like `min` or `max` or `format` are OK in short methods.
If you want to avoid them, `low` and `high` or `lo` / `hi` is another option.

eteq: I'm a bit confused now, @cdeil @bmorris3 - As I see it the main use for this is: ``AirmassConstraint(2)``, meaning "airmass below 2, meaning the altitude should be *above* 30 degrees.  I think 99% of users will immediately recognize that an airmass constraint means "should be lower airmass/higher altitude than that".  But whether that is an `upper_limit` or `max` or `min`, I'm not sure because airmass and altitude have opposite senses...

But I just spent about 10 min talking with my observer officemates, and we couldn't come up with a pair of words that were unambiguous.  So just so long as ``AirmassConstraint(2)`` means "the object should have a secz value less than 1``, I'm happy ;)

cdeil: @eteq The problem with making `AirmassConstraint(2)` mean what you want is that it means we have to switch min / max order, i.e. have it be `AirmassConstraint(max, min)`, where all other constraints are `SomeConstraint(min, max)`.
So that can also be confusing to users.

But actually: I'll probably never use this constraint (we use Altitudes), so whatever you agree on: fine with me.

eteq: Right, I see that, but I'm saying it's a moot point anyway because most observers I know use "above airmass" and "below airmass" interchangeably, as it's usually clear from context what you really mean.  But as a *constraint* it makes more sense the way I was saying.

adrn: @eteq I predict a PR in ~8 months requesting ability to specify a minimum airmass :)

I have no opinion on the options here, but if I were to code this up myself, for consistency with other constraints I would probably name it `AirmassRange` and do `AirmassRange(None, 2)` if I wanted an airmass less than 2 (and would name the kwargs `min` and `max` as @cdeil suggested).

eteq: I'm fine with `AirmassRange`, but I'm strongly against ``AirmassAnything(None,2)``.  Honestly, I'm not sure there's any use case for an "annulus" in Airmass: any real constraints I know of are more reasonably phrased as altitude.  The whole reason for this constraint is mainly a convenience for the (extremely common) case of "my spectra will be junk if they're below airmass 2" or similar.

adrn: :+1:

eteq: And in case it's not clear, I'm perfectly happy with allowing ``AirmassRange(2, 1.1)``: even though I don't know of any use case there might be an edge one.  I just don't think we should make the much more common ``AirmassRange(2)`` suffer because of the option for that.

cdeil: @bmorris3 – Can you please at least point out the weird argument order  in the `AirmassWhatever` docstring and explain with one or two sentences or examples why it's different from `AltitudeRange` (which has `min` as first parameter).

adrn:
I would recommend breaking out the new classes into a new module `constraints.py` to get them out of `core.py`

bmorris3:
This PR is outdated and overwritten by #62 and #61. Closing!

PR #53, opened 2015-08-06
__repr__'s for Observer, FixedTarget

bmorris3:
It's time to make some informative `__repr__` methods for `Observer` and `FixedTarget`. Here's what I've done to start the conversation: 

```python
>>> from astroplan import Observer, FixedTarget
>>> keck = Observer.at_site("Keck", timezone="US/Hawaii")
>>> print(keck)
<Observer: name=u'Keck',
    location (lon, lat, el)=(-155.478333333 deg, 19.8283333333 deg, 4160.0 m),
    timezone=<DstTzInfo 'US/Hawaii' LMT-1 day, 13:29:00 STD>>

>>> polaris = FixedTarget.from_name('Polaris')
>>> print(polaris)
<FixedTarget: name=u'Polaris',
    coord=<SkyCoord (ICRS): (ra, dec) in deg
    (37.95456067, 89.26410897)>>
```
Thoughts and feelings, @astroplanners/astroplanners-contributors?

### DIFF
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -158,6 +158,24 @@ class Observer(object):
             raise TypeError('timezone keyword should be a string, or an '
                             'instance of datetime.tzinfo')
 
+    def __repr__(self):
+        class_name = self.__class__.__name__
+        attr_names = ['name', 'location', 'timezone', 'pressure', 'temperature',
+                      'relative_humidity']
+        attr_values = [getattr(self, attr) for attr in attr_names]
+        attributes_strings = []
+        for name, value in zip(attr_names, attr_values):
+            if value is not None:
+                if name == 'location':
+                    formatted_loc = ["{} {}".format(i.value, i.unit)
+                                     for i in value.to_geodetic()]
+                    attributes_strings.append(
+                        "{} (lon, lat, el)=({})".format(name,
+                                                        ", ".join(formatted_loc)))
+                else:
+                    attributes_strings.append("{}={}".format(name, repr(value)))
+        return "<{}: {}>".format(class_name, ",\n    ".join(attributes_strings))
+
     @classmethod
     def at_site(cls, site_name, **kwargs):
         """
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -1411,14 +1411,8 @@ class FixedTarget(Target):
 
     def __repr__(self):
         class_name = self.__class__.__name__
-        attr_names = ['name', 'coord']
-        attr_values = [getattr(self, attr) for attr in attr_names]
-        attributes_strings = []
-        for name, value in zip(attr_names, attr_values):
-            if value is not None:
-                attributes_strings.append("{}={}".format(name, repr(value)))
-        return "<{}: {}>".format(class_name, ",\n    ".join(attributes_strings))
-
+        fmt_coord = repr(self.coord).replace('\n   ', '')[1:-1]
+        return '<{} "{}" at {}>'.format(class_name, self.name, fmt_coord)
 
 class NonFixedTarget(Target):
     """
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -166,6 +166,7 @@ class Observer(object):
         attributes_strings = []
         for name, value in zip(attr_names, attr_values):
             if value is not None:
+                # Format location for easy readability
                 if name == 'location':
                     formatted_loc = ["{} {}".format(i.value, i.unit)
                                      for i in value.to_geodetic()]
@@ -1407,6 +1408,17 @@ class FixedTarget(Target):
         if name is None:
             name = query_name
         return cls(SkyCoord.from_name(query_name), name=name, **kwargs)
+
+    def __repr__(self):
+        class_name = self.__class__.__name__
+        attr_names = ['name', 'coord']
+        attr_values = [getattr(self, attr) for attr in attr_names]
+        attributes_strings = []
+        for name, value in zip(attr_names, attr_values):
+            if value is not None:
+                attributes_strings.append("{}={}".format(name, repr(value)))
+        return "<{}: {}>".format(class_name, ",\n    ".join(attributes_strings))
+
 
 class NonFixedTarget(Target):
     """
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -159,6 +159,19 @@ class Observer(object):
                             'instance of datetime.tzinfo')
 
     def __repr__(self):
+        """
+        String representation of the `~astroplan.Observer` object.
+
+        Examples
+        --------
+
+        >>> from astroplan import Observer
+        >>> keck = Observer.at_site("Keck", timezone="US/Hawaii")
+        >>> print(keck)                                    # doctest: +FLOAT_CMP
+        <Observer: name='Keck',
+            location (lon, lat, el)=(-155.478333333 deg, 19.8283333333 deg, 4160.0 m),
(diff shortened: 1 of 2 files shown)


eteq:
I definitely like the idea of informative ``__repr__``s.  I think the `Observer` one looks good as-is, but for `FixedTarget`, perhaps something slightly simpler is better given that it's always got the same type of information?  That is, something like: ``<FixedTarget "Polaris" at SkyCoord (ICRS): (ra, dec) in deg
    (37.95456067, 89.26410897)>`` (note also the removal of the brackets around SkyCoord - I think the ``>>`` at the end is a bit confusing/ugly and it's still clear what's going on)

bmorris3:
RE @eteq: I've changed `FixedTarget.__repr__` to show:
```python
from astroplan import FixedTarget
polaris = FixedTarget.from_name('Polaris')
print(polaris)
<FixedTarget "Polaris" at SkyCoord (ICRS): (ra, dec) in deg (37.95456067, 89.26410897)>
```

cdeil:
:+1: 

How does it look for a vector of targets?
Please add tests for those repr functions.

bmorris3:
@cdeil for a vector of targets: 
```python
from astroplan import FixedTarget
ft_list = [FixedTarget.from_name(name) 
           for name in ["Vega", "Polaris", "Rigel"]]
print(ft_list)
```
gives
```
[<FixedTarget "Vega" at SkyCoord (ICRS): (ra, dec) in deg 
(279.23473479, 38.78368896)>, <FixedTarget "Polaris" at 
SkyCoord (ICRS): (ra, dec) in deg (37.95456067, 89.26410897)>, 
<FixedTarget "Rigel" at SkyCoord (ICRS): (ra, dec) in deg 
(78.63446707, -8.20163837)>]
```

cdeil:
I just tried the `Time` and `SkyCoord` classes. Once they reach some length the `repr` uses `...` to get a short string:
```
In [15]: SkyCoord(np.zeros(10000), np.ones(10000), unit=('deg','deg'))
Out[15]: 
<SkyCoord (ICRS): (ra, dec) in deg
    [(0.0, 1.0), (0.0, 1.0), (0.0, 1.0), ..., (0.0, 1.0), (0.0, 1.0),
     (0.0, 1.0)]>
```
I like short `repr`, but I'm not sure it's worth implementing something else than what is here now. @eteq What do you think?
One point that's important is that `repr` never raise, because that can result in hard to understand tracebacks. So please add at least one test for the vector case (not sure if there are other cases that should be tested).

cdeil:
There's travis-ci fails for the tests implemented here.
E.g. https://travis-ci.org/astroplanners/astroplan/jobs/74598958#L528

bmorris3:
@cdeil / @eteq / @adrn [This problem](https://travis-ci.org/astroplanners/astroplan/jobs/74598958#L528) is a 2/3 compatibility issue that I could use some help on. Strings in py3 are printed with a leading `u` that doesn't appear in py2, which is breaking some of my `repr` tests:
```
    def test_observer_repr():
        obs = Observer.at_site("Keck", timezone="US/Hawaii")
>       assert repr(obs) == ("<Observer: name='Keck',\n    location (lon, lat, el)="
                             "(-155.478333333 deg, 19.8283333333 deg, 4160.0 m),\n "
                             "   timezone=<DstTzInfo 'US/Hawaii' LMT-1 day, "
                             "13:29:00 STD>>")
E       assert '<Observer: n...3:29:00 STD>>' == '<Observer: na...3:29:00 STD>>'
E         - <Observer: name=u'Keck',
E         ?                 -
E         + <Observer: name='Keck',
E               location (lon, lat, el)=(-155.478333333 deg, 19.8283333333 deg, 4160.0 m),
E               timezone=<DstTzInfo 'US/Hawaii' LMT-1 day, 13:29:00 STD>>
```
 What's the best way to test for string equivalence universally on py2 and 3?

bmorris3:
Also @eteq / @cdeil: it'd be great if you took a look at the one-character PR astropy/astropy#4057, which is needed to make this astroplan PR (#53) work properly.

cdeil:
@bmorris3 – So this needs https://github.com/astropy/astropy/pull/4057 before the tests pass? If so we can't merge it, because we support Astropy 1.0 for now and this would break our continuous integration.

Maybe just mark these tests xfail for now and open a new reminder issue that they should be activated later?

cdeil:
> What's the best way to test for string equivalence universally on py2 and 3?

Usually using string literals works, no?
```python
assert "this" == "that"
```
What is difference about the repr string here that makes it fail on Py 2 or 3 (if that is the case, I didn't look)?

bmorris3:
@cdeil These tests will pass before #4057 is fixed, because they've been written to not test a case where the astropy issue will cause failure. I'm just pointing you to that issue because it's a simple review.

Re "what's the difference": Py 3 stores strings as unicode strings by default, Py 2 does not. If a string is generated by `repr` in Py 2 and a string is printed with like `"string"`, but the test is expecting Py 3 compatible output which looks like this `u"string"` the test will fail; and vice versa.

bmorris3:
@cdeil and I have found that there are two separate issues. In one case there are more decimals of the numbers represented one version of the `repr`, the other has to do with unicode string literals. We're saving this to discuss with @eteq or @adrn either in GitHub or in the Hangout tomorrow.

adrn:
Are you sure there is a problem with unicode literals? My dumb test that makes me question that idea:

```python
Python 3.3.2 (default, Jun 30 2013, 12:10:10)
[GCC 4.2.1 Compatible Apple Clang 3.1 (tags/Apple/clang-318.0.61)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> u'apple' == 'apple'
True
```

```python
Python 2.7.10 |Anaconda 1.8.0 (x86_64)| (default, May 28 2015, 17:04:42)
[GCC 4.2.1 (Apple Inc. build 5577)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://binstar.org
>>> u'apple' == 'apple'
True
```

bmorris3:
It's taking the `repr`, then printing a string so that it comes out either `"u'Keck'"` or `"'Keck'"`.

adrn:
RE: the number of decimals issue -- I've run in to this problem before (with @eteq ) when writing doctests for `astropy.coordinates`. I think there is a way to solve this using the doctest directive `FLOAT_CMP`, e.g., on the line of the `__repr__` test, add: `# doctest: +FLOAT_CMP` at the end of the line. Though, it looks like someone changed that to a plain ol' `# doctest: +SKIP` in astropy master, so not sure if that actually worked consistently...

e.g., see: https://raw.githubusercontent.com/astropy/astropy/master/docs/coordinates/sgr-example.rst

adrn:
@bmorris3 ah I misread where the 'u' was ending up

cdeil:
I just now thought of a possible solution for the float representation difference issue: use only integers in the repr test.

bmorris3:
@cdeil / @adrn: I got it to pass all of the tests! I it by: 
 * undoing the `repr` call within the `__repr__` method for the name argument, which gets rid of the pesky preceding `u` for the unicode string
 * swapping the `__repr__` tests in the `/tests` dir for doctests in each `__repr__` and using the `# doctest: +FLOAT_CMP` command suggested by @adrn to dodge the differing float length problem

Good to merge?

adrn:
Nice! I'm down to merge

adrn:
:clap:

PR #58, opened 2015-08-11
Added astroplan.plots API to docs.

jberlanga:
@eteq

### DIFF
--- a/astroplan/plots/__init__.py
+++ b/astroplan/plots/__init__.py
@@ -1,2 +1,10 @@
+"""
+`astroplan.plots` is a module for making plots of important quantities within
+the `astroplan` toolbox, such as airmass vs. time, etc.
+
+Note that, at minimum, you must construct `Observer`, `Target` and
+`astropy.time.Time` objects as input for all plots.
+"""
+
 from .time_dependent import *
 from .sky import *
--- a/astroplan/plots/__init__.py
+++ b/astroplan/plots/__init__.py
@@ -1,6 +1,7 @@
 """
-`astroplan.plots` is a module for making plots of important quantities within
-the `astroplan` toolbox, such as airmass vs. time, etc.
+`astroplan.plots` contains functions for making plots of commonly-used
+quantities in observation planning (e.g., airmass vs. time), using `astroplan`
+and `Matplotlib`.
 """
 
 from .time_dependent import *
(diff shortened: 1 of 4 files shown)


astroplan/plots/__init__.py
```diff
@@ -1,2 +1,10 @@
+"""
+`astroplan.plots` is a module for making plots of important quantities within
+the `astroplan` toolbox, such as airmass vs. time, etc.
+
+Note that, at minimum, you must construct `Observer`, `Target` and
+`astropy.time.Time` objects as input for all plots.
```
eteq: @jberlanga - I would suggest not saying this part here, because it's not necessarily true in general - future plots might use other things as inputs.  I think this is pretty straightforward in the API docs (or it should be made so if it isn't!), so specific mentions don't really belong in the module level docstring. One option might be to re-word this to just point to the main `astroplan` package, saying something like "these plotting function use functionality and classes from the `astroplan` package, so familiarity with that is assumed when using these plotting functions."

However, perhaps something to add here instead would be references to the tutorials?  That is, something like ``for worked examples on how to use some of these functions, see the :doc:`name_of_tutorial` page.``?

jberlanga: @eteq: Since there's no tutorials page yet, I'll add a link in to it in the other docs branch.

jberlanga:
@astroplanners/astroplanners-contributors -- Anything else to do here before merging?

astroplan/plots/__init__.py
```diff
@@ -1,2 +1,7 @@
+"""
+`astroplan.plots` is a module for making plots of important quantities within
+the `astroplan` toolbox, such as airmass vs. time, etc.
```
cdeil: I'd suggest you mention that, at least at the moment, we only support matplotlib, maybe this:
```
`astroplan.plots` contains functions to create common astro planning plots using `matplotlib`.
```

cdeil:
@jberlanga – I build the HTML docs from this branch locally and noticed that a few parameters in the plotting functions (see screenshot below) don't have the link to the parameter type formatted correctly.
Probably there's a missing space between the colon and type or some other minor formatting problem.

Do you want to fix those in this PR or should I merge this now (after the one in-line comment above is adressed) and you'll fix those in some other PR?

screen shot 2015-08-17 at 22 18 18

astroplan/plots/__init__.py
jberlanga: Ok, done!

jberlanga:
@cdeil -- Fixed the parameter type typos and edited the `.../plots/__init__.py` docstring.  Should be good to merge!

cdeil:
@jberlanga – Thanks!

I think there's still some issue in the `plot_sky` docstring ... I'll just fix that with a commit in master.

PR #59, opened 2015-08-12
Adding lots of examples to the docstrings in core

bmorris3:
Examples, you said? We've got examples.

Note: some of them can/should get updated as other PRs get merged.

### DIFF
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -110,7 +110,34 @@ class Observer(object):
     A container class for information about an observer's location and
     environment.
 
-    TODO: write this docstring
+    Examples
+    --------
+    We can create an observer at Subaru Observatory in Hawaii two ways. First,
+    by specifying the location with `~astropy.coordinates.EarthLocation`:
+
+    >>> from astropy.coordinates import EarthLocation
+    >>> import astropy.units as u
+    >>> location = EarthLocation.from_geodetic(-155.4761*u.deg, 19.825*u.deg,
+    ...                                        4139*u.m)
+    >>> subaru = Observer(location=location, name="Subaru", timezone="US/Hawaii")
+
+    Locations for some observatories are stored in astroplan, and these can be
+    accessed by name, like so:
+
+    >>> from astroplan import Observer
+    >>> subaru = Observer.at_site("Subaru", timezone="US/Hawaii")
+
+    To find out which observatories can be accessed by name, check out
+    `~astroplan.get_site_names`.
+
+    You can also create an observer without an
+    `~astropy.coordinates.EarthLocation`:
+
+    >>> from astroplan import Observer
+    >>> import astropy.units as u
+    >>> subaru = Observer(longitude=-155.4761*u.deg, latitude=19.825*u.deg,
+    ...                   elevation=0*u.m, name="Subaru", timezone="US/Hawaii")
+
     """
     @u.quantity_input(elevation=u.m)
     def __init__(self, location=None, timezone='UTC', name=None, latitude=None,
--- a/astroplan/core.py
+++ b/astroplan/core.py
@@ -1355,7 +1355,7 @@ class Observer(object):
 
     def moon_illumination(self, time):
         """
-        Calculate the illuminated fraction of the moon
+        Calculate the illuminated fraction of the moon.
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


bmorris3:
Calling astropy tests experts @cdeil / @eteq : some of the examples that are being run as doctests use `FixedTarget.from_name("<star name>")` to succinctly create a `FixedTarget`. I've added the `#doctest: +REMOTE_DATA` comment to those lines, but it turns out that the tests simply skip any lines with this tag. Since I later do operations that depend on these targets getting initialized, the tests are failing on this PR. 

What's the right way to move forward here? Is there a remote-data trick that I don't know, or should all `FixedTarget` instances be initialized directly with `SkyCoord`s that don't access the internet (i.e. no `.from_name` calls)?

astroplan/core.py
```diff
@@ -359,6 +415,22 @@ def altaz(self, time, target=None, obswl=None):
             If ``target`` is `None`, returns `~astropy.coordinates.AltAz` frame.
             If ``target`` is not `None`, returns the ``target`` transformed to
             the `~astropy.coordinates.AltAz` frame.
+
+        Examples
+        --------
+        Create an instance of the `~astropy.coordinates.AltAz` frame for an
+        observer at Apache Point Observatory at a particular time; and
+        transform that
```
cdeil: For this example I'd suggest to split it in two parts:

1. Up to the creation of the `altaz_frame`
2. Starting with the definition of `target` and transforming it.

The "and transform that" part here looks out of place.
If you split it in two sections you can make this a full sentence that starts the second part.

astroplan/core.py
```diff
@@ -359,6 +415,22 @@ def altaz(self, time, target=None, obswl=None):
...
+
+        >>> from astroplan import Observer
+        >>> from astropy.time import Time
+        >>> from astropy.coordinates import SkyCoord
+        >>> apo = Observer.at_site("APO")
```
cdeil: "APO" is cryptic ... no idea where that is. Can we always use "Subaru" or other well-known sites?

astroplan/core.py
```diff
@@ -359,6 +415,22 @@ def altaz(self, time, target=None, obswl=None):
...
+        >>> time = Time('2001-02-03 04:05:06')
+        >>> target = SkyCoord(0*u.deg, 0*u.deg)
+        >>> altaz_frame = apo.altaz(time)
+        >>> target_transformed_to_altaz_frame = target.transform_to(altaz_frame)
+        >>> target_transformed_to_altaz_frame = apo.altaz(time, target) # Alternatively
```
cdeil: Maybe shorten variable "target_transformed_to_altaz_frame" -> "target_altaz"?
And show the computed result output?

astroplan/core.py
cdeil: Ah ... OK, you're saying "Apache Point Observatory" above. Sorry for the noise.

cdeil:
I like the examples that use `from_name` ... makes them nice and short and lets you quickly set up what you need. So if possible, it would be good to keep this. And it would be nice to keep the doctests activated to make docs maintenance for the coming years easier.

I haven't done this myself, but it should be possible to mock the `from_name` methods for the Sphinx build and that might be the best solution.

Here's some info on how to mock packages, I haven't looked up info how to mock specific methods only:
https://docs.readthedocs.org/en/latest/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules

@bmorris3 – For now you could just try to do the mocking in `docs/conf.py`.
Eventually it might be better to put this in the package somewhere so that we can also use it for the tests.
Maybe have a `_from_name_mock` method that contains hard-coded values or reads them from a JSON file? And then how is it activated for the Sphinx build and tests?

@eteq, @adrn @ejeschke – Any thoughts / advice?

astroplan/core.py
bmorris3: I avoided showing the output because the `repr`s may be in flux (#53, astropy/astropy#4055). If/when those are merged/closed and the `repr`s won't lead to more confusion, we can show the outputs. 

At least once, I went through some gymnastics to avoid showing `repr`s while still showing outputs, like [here, for example](https://github.com/astroplanners/astroplan/pull/59/files#diff-8237cb1c7bdb65945cb3e8f60519cf85R1473). Would you like to see that kind of print statement in this alt/az example as well?

cdeil:
@bmorris3 – Can you try the mocking of `from_name`?
(I'm only available until tomorrow if there are questions or issues.)

bmorris3:
@cdeil I'll try now.

bmorris3:
I'm closing this PR and replacing it with #62.

PR #61, opened 2015-08-12
Constraints

bmorris3:
Here's a first implementation of the constraints module, based on a rough pseudocode sketch for the inheritance scheme by @eteq. 

To see how this constraints module would be used, take a peek at this [demo notebook](https://github.com/bmorris3/gsoc2015/blob/master/constraints-demo.ipynb). UPDATE: I've added [high level docs](http://staff.washington.edu/bmmorris/_build/html/tutorials/constraints.html) for you to inspect also.

This PR is intended to be a conversation starter. I've made some decisions in naming that will likely be controversial, so let's discuss!


bmorris3:
@eteq – The travis build failed on some Python 3 runs because `Time` is not a hashable type. Any idea why this would affect py3 but not py2? Should I make a tuple of the JDs instead?

adrn:
Nice! I'm digging this.

I was talking to @astrofrog earlier about a use case that would be nice to have: given an observer, a *single* time, a list of targets, constraints, and (optionally) a list of exposure times (or is this an attribute of `Target`?), which targets can I observe? This would be super nice to have when observing -- not for scheduling, but for more free form observing :). 

Amateur astronomers would probably also find this useful -- we could imagine building a way to, given a catalog name, return all objects that are visible (e.g., what Messier objects are observable right now from XX?) Thinking a bit more long term, it would also be straightforward to build a web front-end with Flask and yadda yadda...

jberlanga:
This looks good! I like the `is_observable`/`is_always_observable` methods in particular--exactly the questions I was trying to answer in my Summer Triangle tutorial.

bmorris3:
A wishlist of constraints that could/should be added is developing [on my wiki](https://github.com/bmorris3/astroplan/wiki/Constraints-Wishlist). Contribute if there's anything missing!

@astroplanners/astroplanners-contributors: @jberlanga brought up one type of constraint that we haven't discussed in a while – target-intrinsic constraints. For example: give me all targets in this list with 11<V<14. Of course, by themselves these target-intrinsic constraints could be sorted out by the user in a simple for loop before they hit the `constraints` module, but it would be more natural to write them into a subclass of constraint. 

One option for doing this could be to allow users to add their own kwargs to `FixedTarget` which gets stored on the object, and then have a generic `TargetPropertyConstraint(property_keyword, [min, max])` that loops through the targets and applies the cut by the supplied keyword. That might look like this:
```python
from astroplan import FixedTarget
from astroplan.constraints import Constraint

polaris = FixedTarget.from_name("Polaris", Bmag=2.62)
aldebaran = FixedTarget.from_name("Aldebaran", Bmag=2.4)
algol = FixedTarget.from_name("Algol", Bmag=2.07)

class TargetPropertyConstraint(Constraint):
    def __init__(self, property_keyword, min=None, max=None):
        self.min = min
        self.max = max
        self.property_keyword = property_keyword

    def _compute_constraint(self, time_range, observer, targets):
        if self.min is None:
            mask = [self.min < target[self.property_keyword]
                    for target in targets]
        elif self.max is None:
            mask = [self.max > target[self.property_keyword]
                    for target in targets]
        elif self.min is not None and self.max is not None:
            mask = [self.min < target[self.property_keyword]
                    for target in targets] & [self.max > target[self.property_keyword]
                    for target in targets]
        return mask
```
How would you like to see this?

astroplan/constraints.py
```diff
@@ -0,0 +1,303 @@
...
+    def _compute_constraint(self, time_range, observer, targets):
+        sun_altaz = self._get_solar_altitudes(time_range, observer, targets)
+        solar_altitude = sun_altaz['altitude']
+        mask = solar_altitude < self.max_solar_altitude
+        return mask
+
+def is_always_observable(constraints, time_range, targets, observer):
+    """
+    Are the ``targets`` always observable throughout ``time_range`` given
+    constraints in ``constraints_list`` for ``observer``?
```
ejeschke: Looks like "constraints_list" should be changed to "constraints" to match the code?

bmorris3:
@cdeil and others: there are some high level docs for the `constraints` module [here](http://staff.washington.edu/bmmorris/_build/html/tutorials/constraints.html). I'll make a commit with those docs included in the PR once @jberlanga 's high level docs get posted, so I can link to the tutorial from her tutorials page.

astroplan/constraints.py
```diff
@@ -0,0 +1,565 @@
...
+            if force_zero_pressure:
+                observer.pressure = observer_old_pressure
+
+        return observer._altaz_cache[aakey]
+
+class AltitudeConstraint(Constraint):
+    """
+    Constrain the altitude of the target.
+
+    Note: this will misbehave if you try to constrain negative altitudes, as
```
eteq: Instead of "Note:", I suggest using:
```
.. note:: 
    This will misbehave ...
```
If you do that, the "note" will appear much prettier in the sphinx documentation. (Same comment below for all the places you do "note:")

astroplan/constraints.py
```diff
@@ -0,0 +1,565 @@
...
+        altaz = cached_altaz['altaz']
+        lowermask = self.min < altaz.alt
+        uppermask = altaz.alt < self.max
+        return lowermask & uppermask
+
+class AirmassConstraint(AltitudeConstraint):
+    """
+    Constrain the airmass of a target.
+
+    The airmass is approximated by the secant of the zenith angle.
```
eteq: Maybe change this to something like "in the current implementation the airmass is approximated..."?  It might change if we decide to add one of the slightly more complex but better airmass definitions...

astroplan/constraints.py
```diff
@@ -0,0 +1,565 @@
...
+    """
+    def __init__(self, max=None, min=None):
+        """
+        Note: the ``max`` and ``min`` arguments appear in the order (max, min)
```
eteq: Definitely good to include this note (although see my comment above about using ``.. note::`` to make it slightly prettier).  But I suggest just being even a bit more explicit and including an example along the lines of 
```
Examples
--------
To create a constraint that requires the airmass be better than 2::

    AirmassConstraint(2)

```
I think just about anyone would read that and understand it to be "the airmass value should be less than 2, meaning closer to zenith, not closer to horizon... Whereas the current text is still subject to confusion by the in-a-hurry observer who thinks of "min" as meaning "lowest altitude" because altitude and airmass are sort of mixed together in their head (hint: a good example of such a person is me ;)

astroplan/constraints.py
```diff
@@ -0,0 +1,565 @@
...
+        ``time_range`` by checking constraint at linearly-spaced times separated
+        by ``time_resolution``. Default is 0.5 hours.
+
+    Returns
+    -------
+    ever_observable : list
+        List of booleans of same length as ``targets`` for whether or not each
+        target is ever observable in the time range given the constraints.
+    """
+    if not hasattr(constraints, '__len__'):
```
eteq: can you clarify why you're using `hasattr` here instead of, say `astropy.utils.isiterable`? Not sure it really matters much, but thought I'd check

bmorris3: Only because I've had wonky experiences with `isiterable`, no other reason.

astroplan/constraints.py
```diff
@@ -0,0 +1,565 @@
...
+    def __call__(self, time_range, observer, targets,
+                 time_resolution=DEFAULT_TIME_RESOLUTION):
+
+        if not isinstance(time_range, Time):
+            time_range = Time(time_range)
+        cons = self._compute_constraint(time_range, observer, targets,
+                                        time_resolution=time_resolution)
+        return cons
+
+    def _compute_constraint(self, time_range, observer, targets):
```
eteq: I think this method should get the ``@abstractmethod`` decorator.  That way, if a user tries to do ``Constraint()``, they see an error right away telling them this is an abstract base class.

eteq:
Looks pretty good to me, @bmorris3  - I left a few inline comments and I'm about to issue a PR to bmorris3/astroplan to add another function.  A few other bits that weren't easy to express inline:

* Perhaps you want to move `time_grid_from_range` to `utils`? It's not really constraint-specific - in fact I think @jberlanga might have done something like this as part of the plotting stuff?
* Along similar lines, I think it might be better if instead of using both `time_range` and `time_resolution` everywhere, you have the `Constraint` and subclasses just accept a single `Time` object which is a vector. Then the `is_observable` and `is_always_observable` function could have the `time_resolution` version as a *convenience*
* It would be great if the high-level docs had an example of a "user"-written `Constraint`.  After all, that's one of the main motivations of having this object-oriented approach as opposed to just saying "do stuff with arrays".
* Along those lines, perhaps you should change `Constraint._compute_constraint` to public (i.e. make it `Constraint.compute_constraint` and add a docstring.  Even though it causes a `NotImplementedError` right now (and I further suggested inline you make it abstract), making it public and adding a docstring makes it much easier for users to figure out *how* to subclass it - in the method docstring you can explain exactly what subclasses should actually do with the method and any important "gotchas" in what they should do there.

astroplan/constraints.py
```diff
@@ -0,0 +1,577 @@
...
+
+        times = time_grid_from_range(time_range,
+                                     time_resolution=time_resolution)
+
+        # convert times, targets to tuple for hashing
+        aakey = (tuple(times.jd), tuple(targets))
+
+        if aakey not in observer._altaz_cache:
+            if force_zero_pressure:
+                observer_old_pressure = observer.pressure
```
ejeschke: You can use a `try...finally` to avoid the problem of the `observer` having the `pressure` not restored because of e.g. throwing an exception in `altaz`.  You would do this by:
```python
try:
    if force_zero_pressure:
        observer_old_pressure = observer.pressure
        observer.pressure = 0

    altaz = observer.altaz(times, targets)
   observer._altaz_cache[aakey] = dict(times=times,
                                                              altaz=altaz)
finally:
    if force_zero_pressure:
        observer.pressure = observer_old_pressure
```

astroplan/constraints.py
```diff
@@ -0,0 +1,577 @@
...
+
+        times = time_grid_from_range(time_range,
+                                     time_resolution=time_resolution)
+
+        aakey = (tuple(times.jd), 'sun')
+
+        if aakey not in observer._altaz_cache:
+            if self.force_pressure_zero:
+                observer_old_pressure = observer.pressure
+                observer.pressure = 0
```
ejeschke: Ditto.  This form of stateful, temporary overrides is prone to problems when there are intervening exceptions before the restoration of the state.  e.g. someone working in an interactive session with an `observer` they won't realize that the pressure that they carefully set when constructing the object has been zeroed.

ejeschke:
I think this is a decent start at `Constraints` and good enough for 0.1.  I'm wondering a little about the caching things in `Observer`--I have a feeling that will come back to be an issue later, but I can't articulate a good objection at present.

cdeil:
@bmorris3 – Merge conflicts, please rebase.

cdeil:
@bmorris3 – There's again merge conflicts ... can you please rebase one more time.
Otherwise ... is this ready? Or do you want to do more things here or more review?

bmorris3:
@cdeil The last of @eteq's [suggestions](https://github.com/astropy/astroplan/pull/61#issuecomment-135586476) that I need to complete is to add a high level docs tutorial for user-written constraints. I'm hoping to get to this tomorrow morning (Pacific), then it will be good to go.

cdeil:
I'll do a final review of this now.

astroplan/constraints.py
```diff
@@ -0,0 +1,614 @@
...
+
+# Package
+from .moon import get_moon, moon_illumination
+from .utils import time_grid_from_range
+
+__all__ = ["AltitudeConstraint", "AirmassConstraint", "AtNightConstraint",
+           "is_observable", "is_always_observable", "time_grid_from_range",
+           "SunSeparationConstraint", "MoonSeparationConstraint",
+           "MoonIlluminationConstraint", "LocalTimeConstraint", "Constraint"]
+
```
cdeil: Please fix this everywhere (otherwise the code is PEP8 clean, nice!)
```
astroplan/constraints.py:31:1: E302 expected 2 blank lines, found 1
```

astroplan/constraints.py
```diff
@@ -0,0 +1,614 @@
...
+        # get timezone from time objects, or from observer
+        if self.min is not None:
+            timezone = self.min.tzinfo
+
+        elif self.max is not None:
+            timezone = self.max.tzinfo
+
+        if timezone is None:
+            timezone = observer.timezone
```
cdeil: Unused local variable: `timezone`!?

astroplan/constraints.py
```diff
@@ -0,0 +1,614 @@
...
+        targets = [target.coord if hasattr(target, 'coord') else target
+                   for target in targets]
+
+        moon_separation = Angle([angular_separation(moon.spherical.lon,
+                                                    moon.spherical.lat,
+                                                    target.spherical.lon,
+                                                    target.spherical.lat)
+                                 for target in targets])
+        # The line below should have worked, but needs a workaround.
+        # TODO: once bug has been fixed, replace workaround with simpler version.
```
cdeil: Please either fix this TODO in this PR or say which bug you're referring to here.

astroplan/constraints.py
```diff
@@ -0,0 +1,614 @@
...
+        return lowermask & uppermask
+
+class AirmassConstraint(AltitudeConstraint):
+    """
+    Constrain the airmass of a target.
+
+    In the current implementation the airmass is approximated by the secant of
+    the zenith angle.
+    """
+    def __init__(self, max=None, min=None):
```
cdeil: My IDE (PyCharm) says "call to __init__ of superclass is missed".
I guess this is not an issue in this case, i.e. is it useful to put the super __init__ call in?

astroplan/constraints.py
```diff
@@ -0,0 +1,614 @@
...
+            times = Time([times])
+
+        cons = self.compute_constraint(times, observer, targets)
+        return cons
+
+    def compute_constraint(self, times, observer, targets):
+        # Should be implemented on each subclass of Constraint
+        raise NotImplementedError
+
+    def _get_altaz(self, times, observer, targets,
```
cdeil: This method is not accessing `self`. -> Make it a `staticmethod` or standalone function?

astroplan/tests/test_constraints.py
```diff
@@ -0,0 +1,215 @@
...
+                                 time_range=time_range) ==
+                   len(targets)*[observer_is_night_any])
+
+        assert all(is_always_observable(AtNightConstraint(), subaru, targets,
+                                        time_range=time_range) ==
+                   len(targets)*[observer_is_night_all])
+
+def test_observability_table():
+    subaru = Observer.at_site("Subaru")
+    time_ranges = [Time(['2001-02-03 04:05:06', '2001-02-04 04:05:06']), # 1 day
```
cdeil: Local variable "time_ranges" not used.

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,238 @@
...
+  `astroplan.constraints.AltitudeConstraint` class.
+
+* We place an upper limit on the airmass of each target during observations
+  using the `astroplan.constraints.AirmassConstraint` class.
+
+* Since we're optical observers, we only want to observe targets at night, so
+  we'll also call the `astroplan.constraints.AtNightConstraint` class. We're
+  not terribly worried about sky brightness for these bright stars, so we'll
+  define "night" times as those between civil twilights by using the class
+  method `AtNightConstraint.twilight_civil`::
```
cdeil: Make this a link to the method in the HTML docs?
I think you'd have to type this:
```
`~astroplan.AtNightConstraint.twilight_civil`
```

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,238 @@
...
+
```
cdeil: At the moment it looks like this code block only is connected to the last bullet point:
screen shot 2015-09-01 at 22 50 21

I'd say remove the `::` and use a newline and separate:
```
.. code-block:: python
```
then the code block should be non-indented and it's more clear that this is the code for all three previous bullets.

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,238 @@
...
+    from astroplan import (AltitudeConstraint, AirmassConstraint,
+                           AtNightConstraint)
+    constraints = [AltitudeConstraint(10*u.deg, 80*u.deg),
+                   AirmassConstraint(5), AtNightConstraint.twilight_civil()]
+
+This list of constraints can now be applied to our target list to determine
+whether:
+
+* the targets are observable given the constraints at *any* times in the time
+  range, using `astroplan.constraints.is_observable`,
```
cdeil: This does not generate a link in the HTML docs ... I think it has to be:
```
`~astroplan.is_observable`
```
i.e. the place where users import from and where it shows up in the API docs.

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,238 @@
...
+
+* the targets are observable given the constraints at *all* times in the time
+  range, using `astroplan.constraints.is_always_observable`::
```
cdeil: Same here ... use this:
```
`astroplan.is_always_observable`
```

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,238 @@
...
+
+Built-In Constraints
+====================
+
+Frequently, we have a long list of targets that we want to observe, and we need
+to know which ones are observable given a set of constraints imposed on our
+observations by a wide range of limitations. For example, your telescope may
+only point over a limited range of altitudes, your targets are only useful
+in a range of airmasses, and they must be separated from the moon by some
+large angle. The `astroplan.constraints` module is here to help!
```
cdeil: The `astroplan.constraints` module is not a namespace that's part of the official astroplan API.
At the moment everything is exposed via the top-level `astroplan` namespace.

So either change this to not refer to `astroplan.constraints` (probably easier) or expose the `astroplan.constraints` namespace (and not put it's content in the top-level namespace) (also very reasonable, but not what we discussed / agreed so far).

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,238 @@
...
+    # Are targets *always* observable in the time range?
+    always_observable = is_always_observable(constraints, subaru, targets, time_range=time_range)
+
+These two functions will return boolean arrays which tell you whether or not
+each target is observable given your constraints. Let's print these results in
+tabular form:
+
+    >>> from astropy.table import Table
+    >>> import numpy as np
+    >>> observability_array = np.array([[t.name, ever, always]  for t, ever, always in
```
cdeil: This is not a good way to build the `Table`.
As you can see in the output below, it leads to the bool columns being string columns.

Instead I'd suggest this:
```
table = Table()
table['targets'] = [target.name for target in targets]
table['ever_observable'] = ever_observable
table['always_observable'] = always_observable
```

astroplan/tests/test_constraints.py
bmorris3: My PyCharm also grays out the `time_ranges` variable, but it is being used [here](https://github.com/astropy/astroplan/pull/61/files#diff-9162b487e33fd6ae22fe36890f1827aeR53).

cdeil: I think no, there you use `time_range` (no "s") which is defined a few lines below.

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,238 @@
...
+
+.. _constraints-user_defined_constraints:
+
+User-Defined Constraints
+========================
+
+There are many possible constraints that you could find useful which have not
+been implemented (yet) in astroplan. This example will walk you through creating
+your own constraint which will be compatible with the tools in the `constraints`
+module.
```
cdeil: Again, there is no user-visible `constraint` module at the moment, so the docs shouldn't talk about it.

Just in case you're leaning that way: I'm 50:50 to make it a user-visible sub-module, either way is fine with me.

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,238 @@
...
+
+We will begin by reading the text file of stellar coordinates defined in the
```
cdeil: Change the order in that sentence to match the order of the code (and the example above):

1. observer
2. target coordinates

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,238 @@
...
+    target_table = ascii.read('targets.txt')
+
+    # Create astroplan.FixedTarget objects for each one in the table
+    from astropy.coordinates import SkyCoord
+    import astropy.units as u
+    targets = [FixedTarget(coord=SkyCoord(ra=ra*u.deg, dec=dec*u.deg), name=name)
+               for name, ra, dec in target_table]
+
+In the above example, you may have noticed that constraints are assembled by
+making a list of calls to the initializers for classes like `AltitudeConstraint`
```
cdeil: Link to the API docs of these classes?
```
`~astroplan.AltitudeConstraint`
```
etc?

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,238 @@
...
+and `AirmassConstraint`. Each of those constraint classes is subclassed from
+the abstract `Constraint` class, and the custom constraint that we're going to
+write must be as well.
+
+In this example, let's design our constraint to ensure that all targets must be
+within some number of degrees from Vega – we'll call it
```
cdeil: "number of degrees" -> "angular separation"?

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,238 @@
...
+
+    from astroplan import Constraint, is_observable
+    from astropy.coordinates import Angle
+
+    class VegaSeparationConstraint(Constraint):
+        """
+        Constraint the separation from Vega
+        """
+        def __init__(self, min=None, max=None):
+            """
```
cdeil: I'd suggest to put the `__init__` docstring in the class-level docstring, in a `Parameters` section.

Either way will work, but having the `__init__` parameters documented in the class-level docstring is how I think most (all?) classes in Astropy are documented.

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,238 @@
...
+within some number of degrees from Vega – we'll call it
+`VegaSeparationConstraint`. Two methods, `__init__` and `compute_constraint`
+must be written for our constraint to work:
+
+* The `__init__` method will accept the minimum and maximum acceptable separations
+  a target could have from Vega.
+
+* We'll also define a method `compute_constraints` which takes three arguments:
+  an array of times to test, an `Observer` object, and one or a list of
+  `FixedTarget` objects. `compute_constraints` will return a matrix of
```
cdeil: Why only `FixedTarget`?
Shouldn't all `Target`s work?

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,238 @@
...
+            """
+            self.min = min
+            self.max = max
+
+        def compute_constraint(self, times, observer, targets):
+
+            # Vega's position is essentially unchanging, but if it were to be a
+            # moving target, we would need an array of coordinates for Vega as a
+            # function of time. Here we'll simulate that behavior with multiple
+            # copies of the Vega coordinate
```
cdeil: Is this complication really needed or useful for this example?
If you just have one `SkyCoord`, does this work via broadcasting, or do you still need to explicitly duplicate by `len(times)`?

docs/tutorials/constraints.rst
bmorris3: Right now only `FixedTarget` is implemented.

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,238 @@
...
+            vega = SkyCoord(len(times)*[vega])
+
+            # If `targets` is a FixedTarget object, get the SkyCoord
+            target_coords = SkyCoord([target.coord if hasattr(target, 'coord')
+                                      else target for target in targets])
+
+            # Calculate separation between target and vega
+            vega_separation = Angle([vega.separation(target)
+                                     for target in target_coords])
+
```
cdeil: Can't the various min / max cases be handled with less boilerplate code?
Does this work? Is it better?
```python
mask = np.ones_like(vega_separation, dtype=bool)
if self.max:
    mask &= (vega_separation > self.max)
if self.min:
    mask &= (vega_separation <= self.min)
```

cdeil:
I've left a bunch of inline comments.

Two more big-picture comments:
- Is the user-defined constraint tested via doctest? If no, do you think it's worth copying over to `test_constraints.py` (with a note that it is the example from the docs)?
- I think the constraints section can stand on it's own, it's nice! But I'd add a sentence at the start, saying something like "this is a start ... more high-level functionality to compute observability and do scheduling will be added to astroplan soon ... feature requests and contributions welcome!"

docs/tutorials/constraints.rst
bmorris3: It will not work via broadcasting, which is why I did it. I discussed this example with @eteq and @adrn , but after coding this line, I started to doubt whether or not it's the best one to use. We wanted an example that was trivial to code and easy to understand, and I think this line doesn't meet the second criteria exactly.

cdeil:
I forgot to say: IMO this is nice and it's been open for a while, so @bmorris3, if you want to merge it, I'd say go ahead! Or leave it open for one more day and ping the others for final review one more time, as you like.

astroplan/constraints.py
cdeil: As discussed on Slack, this is probably not important, you can just ignore this comment on calling `__init__` of the superclass (which doesn't have an `__init__` in this case).

eteq: I agree it is ok to ignore the IDE error: it's true that you're not calling the superclass `__init__`, but that doesn't matter here because the superclass `__init__`s aren't supposed to be doing anything anyway.  It *could* end up confusing if any user wanted to subclass `AirmassConstraint`, but I can't imagine why someone would actually want to do that...

In fact, as it's *currently* implemented, it's not really necessary for `AirmassConstraint ` to be an `AltitudeConstraint` subclass, because you use *none* of the`AltitudeConstraint` behavior in  `AirmassConstraint `.  The main reason to keep it this way is conceptual: I think people think of airmass as sort of another way of thinking about altitude/zentih angle. It also might make more sense down the road for it to be a subclass if more agressive caching of something or other happens: you can implement it all in `AltitudeConstraint` and have `AirmassConstraint` just do a thin layer that converts an airmass into an altitude.  But for now leaving it as-is seems just fine.

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,255 @@
+:orphan:
```
eteq: This is unnecessary - it was needed in the past when there was no index pointing to the tutorials, but now there is, so you should *not* use ``:orphan:`` (it indicates that the document in question isn't linked to from anywhere else).

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,255 @@
...
+
+Contents
+========
+
+* :ref:`constraints-built_in_constraints`
+* :ref:`constraints-user_defined_constraints`
+
+.. _constraints-built_in_constraints:
+
+Built-In Constraints
```
eteq: Maybe this should be something like "Using built-in constraints"?  Or split it into "Introduction to constraints" and "Built-in constraints"?  Right now it's rather imposing-looking because it sounds like you need to know immediately what constraints are and how they are used.

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,255 @@
...
+    Algol 47.042218553 40.955646675
+    Rigel 78.634467067 -8.201638365
+    Regulus 152.092962438 11.967208776
+
+We'll read in this list of targets using `astropy.io.ascii`, and create a list
+of `astroplan.FixedTarget` objects out of them::
+
+    # Read in the table of targets
+    from astropy.io import ascii
+    target_table = ascii.read('targets.txt')
```
eteq: An alternative approach, which *may* be simpler to users (because they've already seen `table` elsewhere in the tutorials) is this:
```
from astropy.table import Table
target_table = Table.read('targets.txt', format='ascii')
```
I'd suggest switching to that just to lower cognitive burden

docs/tutorials/constraints.rst
eteq: (in the second option I mean have this first paragraph be "introduction to constraints" and then have "built-in constraints" be the rest, although maybe with a new lead sentence that says "first we'll talk about using constraints built into astroplan.  Later we'll show you how to define your own" or similar.

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,255 @@
...
+    >>> print(table)
+    target name ever observable always observable fraction of time observable
+    ----------- --------------- ----------------- ---------------------------
+        Polaris            True              True                         1.0
+           Vega            True              True                         1.0
+        Albireo            True             False              0.833333333333
+          Algol            True             False              0.166666666667
+          Rigel           False             False                         0.0
+        Regulus           False             False                         0.0
+
```
eteq: Can you add in the example from the ipython notebook you sent around which actually makes a plot showing points for when things are observable?  Or was that too complicated? (Ideally you should just be able to use `plot_sky`, and point users to the plotting tutorial section for more on that)

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,255 @@
...
+    Albireo 292.68033548 27.959680072
+    Algol 47.042218553 40.955646675
+    Rigel 78.634467067 -8.201638365
+    Regulus 152.092962438 11.967208776
+
+We'll read in this list of targets using `astropy.io.ascii`, and create a list
+of `astroplan.FixedTarget` objects out of them::
+
+    # Read in the table of targets
+    from astropy.io import ascii
```
eteq: You should also add ``from astropy import units as u`` here or further up

eteq: oops, my bad, I see you did that just below

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,255 @@
...
+                raise ValueError("No max and/or min specified in "
+                                 "VegaSeparationConstraint.")
+
+            # Return the boolean mask
+            return mask
+
+Then as in the earlier example, we can call our constraint::
+
+    constraints = [VegaSeparationConstraint(min=5*u.deg, max=30*u.deg)]
+    observability = is_observable(constraints, subaru, targets,
```
eteq: this is definitely a place for `>>>` because you're showing it just below

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,255 @@
...
+            self.max = max
+
+        def compute_constraint(self, times, observer, targets):
+
+            # Vega's position is essentially unchanging, but if it were to be a
+            # moving target, we would need an array of coordinates for Vega as a
+            # function of time. Here we'll simulate that behavior with multiple
+            # copies of the Vega coordinate
+            vega = SkyCoord(ra=279.23473479*u.deg, dec=38.78368896*u.deg)
+            vega = SkyCoord(len(times)*[vega])
```
eteq: Actually, I think you don't need to do this.  If instead you define it as:
```
vega = SkyCoord(ra=[279.23473479]*u.deg, dec=[38.78368896]*u.deg)
```
ordinary numpy broadcasting happens and all is fine without the extra computational expensive of duplicating `vega` a bunch

In the future, it might be better to try to change the observability calcs to have something like ``numpy.atleast_1d`` in them to allow even the scalar case.  But for now I think just the small change I said above is fine.

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,255 @@
...
+
+            # If `targets` is a FixedTarget object, get the SkyCoord
```
eteq: Why is this necessary?  The *superclass* `Constraint` should take care of stuff like this, right?  That is, `compute_constraint` should be able to make better assumptions about what ``targets`` is?

docs/tutorials/constraints.rst
```diff
@@ -0,0 +1,255 @@
...
+            # If both a minimum and a maximum are specified
+            elif self.min is not None and self.max is not None:
+                mask = ((self.min < vega_separation) & (vega_separation < self.max))
+
+            # Otherwise, raise an error
+            else:
+                raise ValueError("No max and/or min specified in "
+                                 "VegaSeparationConstraint.")
+
+            # Return the boolean mask
```
eteq: maybe add just a bit more here along the lines of "return an array that is True where the target is observable and False where it is not"?  The typical user here might not quite know that's what you mean by "mask".

eteq:
Ok, I'm done with my review - mostly doc suggestions (although one possibly-complicated bit about user-defined constraints... we may want to discuss that more on slack)

This has already been discussed some on slack, but I think the whole constraints tutorial should be doctest-ed.  Annoyingly the easiest way is probably to add ``>>>`` in front of basically everything.  But if we don't want to have to figure out how to do this right for 0.1 it could also be moved to the test suite as @cdeil suggested.

docs/tutorials/constraints.rst
bmorris3: I'm going to err on the side of inaction and just leave things as they are.

bmorris3:
Updates done, tests passed, merging! Feel free to leave more comments in issues if I jumped the gun on anything.

cdeil:
:balloon: :balloon: :balloon:  :tada: :balloon: :balloon: :balloon:

eteq:
:fireworks:

PR #62, opened 2015-08-13
Adding lots of examples to the docstrings in core (part two)

bmorris3:
> Examples, you said? We've got examples.
> Note: some of them can/should get updated as other PRs get merged.

This PR fixes some problems in PR #59 by adding the `--remote-data` astropy test option to the travis runs, and adding the `# doctest: +FLOAT_CMP` tag on some print statements that would otherwise yield different results in Python 2 and 3.


cdeil:
Using `--remote-data` and querying [Sesame](http://cds.u-strasbg.fr/cgi-bin/Sesame) all the time via `from_name` is a bit risky. If the uptime of Sesame is no good, it'll mean annoying sporadic travis-ci fails that are unrelated to the PRs.
So I still think mocking `from_name` as suggested in https://github.com/astroplanners/astroplan/pull/59#issuecomment-130194176 is the better approach.

But I realise time is short until the 0.1 release and learning how to implement the mocking takes time, so OK to put this in if others agree. If there are issues we can always do the mocking later.

@adrn @eteq OK to merge?

astroplan/core.py
```diff
@@ -110,7 +110,34 @@ class Observer(object):
     A container class for information about an observer's location and
     environment.
 
-    TODO: write this docstring
+    Examples
```
adrn: Good! I would switch the order though for those who glance at this quickly -- we want them to first try to find their site in the site list and if it's not there show them how to create a site from scratch. That cool with you?

bmorris3: @adrn: Done! (e78f3bb)

cdeil:
:+1:
(I'll merge this tomorrow, unless someone else does a review and / or merges it today.)

jberlanga:
@bmorris3 -- As I was reading through `core.py`, I was confused as to where airmass came from, then I came upon [the altaz frame in `astropy`](http://astropy.readthedocs.org/en/latest/api/astropy.coordinates.AltAz.html).  Am I correct in assuming that the altaz frame in `astroplan` is functionally the same?  In that case, we need to point this out to users in our own altaz docstring.

cdeil:
@bmorris3 – I just realised that this PR breaks `python setup.py test` without the `--remote-data` option:
https://gist.github.com/cdeil/4b477928d76ec1e537f0#file-gistfile1-txt-L260
I.e. someone without internet connection can't run the astroplan tests.
That's not OK and we have to find another solution.

Like I said in https://github.com/astroplanners/astroplan/pull/59#issuecomment-130194176, I think mocking from_name for the tests and docs is the right approach.

I just now thought about this some more, and this would be my suggestion how to implement it:

* Add a function `astroplan.remote_data_mock()` that assigns mock functions to `Target.from_name` and `Observer.at_site`. For a hand-written list of inputs from the docs / tests the outputs are hard-coded. Whether you do this in a JSON file or in the `.py` file itself doesn't matter... whatever you prefer.
* Call this function in `docs/conf.py` so that it's active when building the docs.
* Call this function in `astroplan/conftest.py` so that it's active when running the tests. (I think, here I'm not sure, but if this is not the right location, we can figure out where it needs to go.)

bmorris3:
@cdeil I could use some help with this, I can't figure out if `mock` can be used that way.

bmorris3:
@cdeil I used your mocking suggestion in f36dac0, and it seemed to work fine on my computer, but it failed in travis. Might that `urllib` error be a travis problem? If so, would you mind restarting the run?

bmorris3:
@adrn / @eteq: The current build has an error that looks like weird quantities behavior. In almost all contexts I've been able to compute an inequality for a `Quantity` that's a time duration (days, minutes) and a `TimeDelta`, but it's giving an error on some of the builds (but not all), [see one here](https://travis-ci.org/astroplanners/astroplan/jobs/77017286#L555).

bmorris3:
The errors in this build are related to those being address in #67.

cdeil:
The unrelated travis-ci issues are gone, so we can try to get this finalised.

@adrn What do you think?

astroplan/conftest.py
```diff
@@ -16,3 +16,7 @@
     del PYTEST_HEADER_MODULES['h5py']
 except NameError:  # needed to support Astropy < 1.0
     pass
+
+# Mock FixedTarget.from_name class method for tests without remote data
```
cdeil: I think it would be nicer if you hide the implementation of how the mocking is done in a utility function that you call before running the tests or building the docs. Maybe `astroplan._mock_remote_data`?
Currently it would only contain this one line, but I think there will be other stuff in the future, and it gives us a nice place to put a docstring explaining what this is.
Since this is mostly for us, I wouldn't make it part of the public astroplan API.

astroplan/core.py
```diff
@@ -1512,12 +1788,32 @@ def __repr__(self):
         fmt_coord = repr(self.coord).replace('\n   ', '')[1:-1]
         return '<{} "{}" at {}>'.format(class_name, self.name, fmt_coord)
 
+    @classmethod
+    def _fixed_target_from_name_mock(cls, name):
+        """
+        Mock method to replace `FixedTarget.from_name` in tests.
+        """
+        if name.lower() == "rigel":
```
cdeil: Maybe put the "data" in a list of tuples or some other Python data structure to reduce the boilerplate code a little bit?
```python
targets = [
    ("rigel", 78.63446707, -8.20163837),
    ...
]
```

bmorris3:
Re @cdeil: Addressed your comments in be46506

adrn:
Sorry I'm just looking at the most recent commits now.

I'm a bit confused as to why we are "mocking" `from_name`. I understand we don't want the tests to be pinging the internet all the time (I agree this is *bad*), but what is the advantage to creating targets using `from_name` rather than just having a few sets of coordinates manually entered in... I'm :-1: on the mocking stuff that was added and would much prefer to just see the 4 test-case stars hard-coded in somewhere.

astroplan/core.py
```diff
@@ -1512,12 +1788,27 @@ def __repr__(self):
         fmt_coord = repr(self.coord).replace('\n   ', '')[1:-1]
         return '<{} "{}" at {}>'.format(class_name, self.name, fmt_coord)
 
+    @classmethod
+    def _fixed_target_from_name_mock(cls, name):
+        """
+        Mock method to replace `FixedTarget.from_name` in tests.
+        """
+        stars = {"rigel":(78.63446707*u.deg, -8.20163837*u.deg),
```
adrn: Ok, I think this is fine. Note: I would have done this just to be safe:

```python
stars = {
    "rigel": {"ra": 78.63446707*u.deg, "dec": -8.20163837*u.deg},
    "sirius": {"ra": 101.28715533*u.deg, "dec": -16.71611586*u.deg},
    "vega": {"ra": 279.23473479*u.deg, "dec": 38.78368896*u.deg},
    "aldebaran": {"ra": 68.98016279*u.deg, "dec": 16.50930235*u.deg}
}

if name.lower() in stars:
    return cls(coord=SkyCoord(**stars[name.lower()]),
               name=name)
...
```

PR #67, opened 2015-08-24
IERS A table workaround

bmorris3:
Here's one solution to our troubles with the IERS Bulletin A table. I'll try to summarize what this is all about before describing the PR, and hopefully a discussion can emerge from @astroplanners/astroplanners-contributors about whether or not this is an OK fix. I acknowledge the inevitable gripe that this PR should really be to astropy, but for the sake of getting astroplan done before the summer is over, I'm keeping my work in astroplan.

### Background

Earth orientation data is necessary to precisely calculate alt/az positions. The Earth orientation data used by astropy comes from [IERS](http://www.iers.org/IERS/EN/Home/home_node.html), and they post it in two forms. The IERS Bulletin B (denoted throughout astropy as "IERS B" or the "IERS B table") contains the orientation data for the recent past, based only on observations. The IERS Bulletin A has the data in Bulletin B and also extrapolations into the past and future. Leap seconds get factored into the future extrapolations in the IERS A table, making them necessary for precision O(seconds) in the near future. This table gets updated once a week.

iers a/b
See this [blog post](http://bmmorris.blogspot.com/2015/06/ut1-utc-and-astropy.html) that I wrote up a few weeks ago for details on how to make those plots with astropy.

In the current version of astropy, the IERS Bulletin B is included in the package, but IERS Bulletin A is not. If you try doing a calculation with a time in the near future, you get:
```
IndexError: (some) times are outside of range covered by IERS table.
```

In astroplan so far, we've been running a script to download/cache the IERS Bulletin A (~3 MB) every time the user runs astroplan. This hasn't been a big deal because (1) we're connected to the internet, (2) after the first time you download it, you don't notice it's happening. 

Since we want the tests to run without internet access, and we want users to be able to do these calculations without the internet, a workaround is needed that gives users access to IERS Bulletin A without an internet connection.

### This PR

The solution here is to include IERS Bulletin A in the astroplan package `data/` dir, stored as a text file. When the user accesses `astroplan/core.py`, a new function `get_recent_IERS_A_table` in the `utils` module is called that manages IERS Bulletin A. Here's what it does: 
 * It first opens `data/iers_bulletin_a.txt` and looks to find the date when the observed Earth orientation data runs out and the predictions begin (which is flagged within the table). This tells you when the file was last updated
 * If the table has observations <7 days old, that table is returned and used for calculations. 
 * If the table has no observations <7 days old, it tries to download a new copy of the table and store it in the same place as the old one (`data/iers_bulletin_a.txt`)
 * If the table has no observations <7 days old and the internet is not accessible, it uses the old (and only available) table and raises a warning like this: 

```
WARNING: OldEarthRotationDataWarning: Your local copy of the IERS Bulletin A 
table is 8.2 days old, and the internet can not be accessed to download the latest 
copy. Earth rotation calculations will not be as precise as possible until you 
reconnect to the internet and attempt to access the IERS Bulletin A table again.  
[astroplan.utils]
```


adrn:
Thanks for the context and background!

I'm still against including a copy of the table in astroplan because then the file is tied to the release schedule of astroplan. Imagine that releases only happen every 6 months. Then a typical user will install astroplan and have a ~3 month old version of the file. Based on the criteria you outline above for when to grab a new version of the file, all users will have to download a new version of the file anyways...?

bmorris3:
RE @adrn : 

> Then a typical user will install astroplan and have a ~3 month old version of the file.

That's not how this PR works, actually. If the user has internet access, any table more than one week old will get refreshed. If the user does not have internet access, we simply can't do better than using an old table. 

It's true though that we have update the astroplan version of the file every so often to ensure that people downloading the package for the first time get a reasonably recent version of the table to fall back on if they don't have consistent internet access.

adrn:
Sorry, I guess I wasn't clear.

What I mean is this: if the release schedule is every 6 months, and a user installs astroplan 3 months after a release, then their IERS table will be 3 months old. So whenever they use astroplan and have an internet connection, a download will be triggered to get the new version anyways (as it happens now, right?). All I'm saying is that maybe there is a better solution than packaging a perpetually-old version of the table in to astroplan, especially since it is a few MB. For example, there may be a way to include an external file download in `setup.py`? I'm not sure, but @eteq might know.

bmorris3:
> So whenever they use astroplan and have an internet connection, a download will be triggered to get the new version anyways (as it happens now, right?).

In the PR, whenever a download is done it gets saved and replaces the `data/iers_bulletin_a.txt` file. The download only has to happen once a week to be perfectly up-to-date – the download does not happen every time, it just checks whether the present table is up-to-date each time. 

Before this PR, the download happens once and gets cached (similar to in this PR, except that it's saved in the astropy cache).

### Important Aside
Let's remind ourselves of the precision we're quibbling about. We can only calculate rise/set times to an accuracy of a few seconds right now anyways. For the average user, a few second offset of UTC-UT1 is a negligible difference, and even if the user doesn't connect to the internet for a year, the offset is likely still <3 seconds.

bmorris3:
@adrn / @eteq: The current build has an error that looks like weird quantities behavior. In almost all contexts I've been able to compute an inequality for a `Quantity` that's a time duration (days, minutes) and a `TimeDelta`, but it's giving an error on some of the builds (but not all), [see one here](https://travis-ci.org/astroplanners/astroplan/jobs/77017286#L555).

adrn:
I understand that the file is cached now, and I understand the offsets are small. My point is just that even if we include a version of the file in astroplan, most users will still have to download an updated version of the file. I'm not arguing/quibbling about precision, I just don't think we should package the file in astroplan! If most users will need to download a more recent version of the file anyways because the packaged version is out of date, then we should just find a solution that grabs the newest version of the IERS file available *on install*. Again, I don't know how to do this, but I figure there must be a way!

bmorris3:
I think the whole problem that @cdeil was trying to avoid was that users should be able to install/test without downloading anything.

adrn:
And my point is that -- even with this PR -- the typical user will install astroplan, then immediately try to use it. Meaning it will immediately trigger a download to get the latest version of the file because the packaged version of the file is old. So I'm proposing we look for a solution that does the download *on install* rather than on first usage of the code. Then we don't have to package the IERS file with astroplan.

adrn:
(and if that solution I suggest doesn't exist then fine, I give in)

bmorris3:
Ok, now I understand.

I'm not disagreeing with you (yet), but is there a significant difference between doing it on install rather than on first use? Astroplan builds a lot of stuff at first use, rather than doing it on install.

adrn:
In this case, the difference is just not packaging an out of date, 4 MB data file with astroplan that we then have to deal with. I know it doesn't seem like a hassle, but (a) even though it's only MB's, git doesn't like MBs and (b) I'm only making a fuss if there is an easy alternative solution, and this is the kind of something that @eteq will know immediately :)

bmorris3:
Got it. Waiting on @eteq!

cdeil:
@bmorris3 – Thanks for implementing a possible solution for this issue! (Apart from https://travis-ci.org/astroplanners/astroplan/jobs/77017285#L555 which needs to be resolved.)

@adrn – I don't think we have to find and implement the long-term solution for this issue when to download or update the IERS table here. Download on install instead of first import has pretty much the same issues, i.e. install and travis-ci tests can fail if the site that hosts the IERS table is not available.

@olebole as Astropy Debian packager – Any advice on when the IERS A table should be downloaded / updated by Astropy (or astroplan for now) so that it would work e.g. for Astropy in Debian?

@bmorris3 – It seems that @adrn's main concern here is with bundling these few MB in the code repo. (And I share this concern a bit, it's not a great solution.)
Can you come up with a different solution that makes the IERS A table completely optional for astroplan? Maybe fall back to the IERS B table only? Or if that is not possible for some reason (the IndexError in Astropy), then fake a IERS A table using the IERS B table as input and adding one entry in the far past and future that does linear interpolation? I think that would be acceptable if a warning is emitted. Not super-high precision results are better than refusing to produce any results!

cdeil:
@eteq – Assigning this PR to you for the final decision.

bmorris3:
@cdeil Would you suggest also revising all of the tests to use dates that exist in the IERS B table? Otherwise they'll break.

cdeil:
I think you should still test for dates in the the past and future.
But it's OK if those tests are low precision, at a level that corresponds to 'linear interpolation' of the always-available IERS B table (whatever that means exactly).
There can still be high-recision tests that require the IERS A table, but those should be marked `@remote_data` if IERS A isn't bundled.

ejeschke:
No strong opinion here because this seems like trying to pick the "least bad" thing to do, but given a choice between requiring a download at installation vs. first use I would prefer at installation time.  If the cache freshness period is 7 days then at least tests should succeed without requiring a subsequent download.

What about at least providing a separate option in setup.py to download the table?  Then at least the travis builds can run that step before running tests.  I like the idea of the tests giving a warning that the table is out of date (and maybe should suggest the command to run to refresh the table).

olebole:
@cdeil I still do not understand the whole context yet (it is the first time I heared about astroplanner :-) ), but I would put the data file into `/var/cache/...`. If it is used by more than one independent package (so, if astroplanner does not depend on astropy), I would create an extra package for that and  let the others depend on it.
About the update process:  Debian wants that the user keeps control the remote activities like downloads, so an automated download needs to be switched on somehow -- I will find out.

astroplan/core.py
```diff
@@ -37,6 +28,9 @@
 
 #__doctest_requires__ = {'*': ['scipy.integrate']}
 
+from astropy.utils import iers
+iers.IERS.iers_table = get_recent_IERS_A_table()
```
eteq: right now this functions gets called *at import*.  What about instead having it get called the first time the user actually *does something* that requires the `iers` tables?  Or is that too hard to keep track of?

astroplan/utils.py
```diff
@@ -2,9 +2,60 @@
...
+    # If the IERS bulletin is more than 7 days old, try to download a fresh one.
+    # Note that this doesn't ensure that the file that you download will be
+    # newer, it just ensures that you've tried to download the latest copy.
+    if warn_update < time_since_last_update:
+        try:
+            path_to_tmp_file = download_file(iers.IERS_A_URL, cache=False,
+                                             show_progress=False)
+            os.remove(local_iers_a_path)
+            os.rename(path_to_tmp_file, local_iers_a_path)
+            table = iers.IERS_A.open(local_iers_a_path)
```
eteq: This is dangerous.  If I understand right, you're writing the new IERS file into the *source* directory.  In many situations (i.e., just about any linux system-level install) this will not work because the user's permissions are set to not allow users to alter the source (after all, that's a major security risk if a malicious piece of code edits a file to do ``os.system('rm -rf /')``!), and that would include the `data` directory.

So I would instead suggest doing this:
```python
from astropy.config.paths import get_cache_dir
path_to_write_to = os.path.join(get_cache_dir(), 'astroplan', 'iers_bulletin_a.txt')
```
You'll have to add a bit more logic to first check the data dir with this approach, but perhaps the easiest way is to *always* use the `path_to_write_to` as I define above as the file to read from, and do something like "if the file is absent, copy over what's in `get_pkg_data_filename`'.  Note this will also make the approach I'm about to describe in the comments a bit easier.

eteq:
@bmorris3 @adrn - I left some inline suggestions on this (including an important but probably fairly easy-to-implement bit about location of the files).  But I'd also like to give an opinion on the at-install vs as-import question (and hopefully we can resolve this in the telecon later today): 

I'm generally in favor of @adrn's idea of downloading at build-time if possible (more on how to do that below).  But I have one question for @bmorris3: do you *need* the IERS A file for some of this to work at all?  That is, is there functionality that works with even a horribly out-of-date IERS A that *won't* work with just IERS B?  If yes, then I'd say we have to include one with each release (updated, if possible, immediately before release), because that accounts for the user who gets the package and then later tries to install it without internet access.  So in that case I'm in favor of this PR *plus* what @adrn is suggesting.

As for the mechanics of downloading at-install: I think this isn't going to be that hard to implement if we follow my second in-line comment above about having the "preferred" IERS A file be in the astropy cache directory.  If that's what we want, all you have to do is do the download into that same location at the end of the install procedure.  Adding a final step to the install isn't too hard because we can just add in a custom install build command that runs a function to do this.  It should be only ~10 lines of code to do that (although it does require a bit of modification to the base level `setup.py`).

bmorris3:
Closing in favor of #71, after conversation over Hangout and Slack.

PR #69, opened 2015-08-25
IERS Bulletin A download on install and/or when out of date

bmorris3:
Here I've used setuptools ninja skills to download the IERS Bulletin A on install, and also removed the dependence of the tests on the IERS Bulletin A. We can add extra tests that do depend on IERS A and mark them with `@remote_data` so they only get triggered if the tests are run with `python setup.py test --remote-data`. 

This PR preserves all of the development in #67 (see that one for background) except it does NOT include the pre-bundled IERS Bulletin A copy within the repository.


eteq:
@bmorris3 : hah, this is *just* the sort of thing I suggested in #67 !  So I'm in favor of this modulo the comments I left there.

cdeil:
Putting this import in setup.py results in this fail because at that time Astropy isn't installed yet:
https://travis-ci.org/astroplanners/astroplan/jobs/77197801#L379

cdeil:
There's also still this issue:
https://travis-ci.org/astroplanners/astroplan/jobs/77197813#L518
@eteq - Help?

bmorris3:
Closing in favor of #71, after conversation over Hangout and Slack.

PR #42, opened 2015-07-27
Add high-level docs

jberlanga:
Added additional docs/*.rst files for documentation.

Addresses Issue #41.

PR #71, opened 2015-08-26
IERS Bulletin A workaround and download methods

bmorris3:
This PR is the culmination of discussions generated by #67 and #69, and conversations with @eteq on Slack. @cdeil / @adrn, fire away. Here's how it works: 

### Outline
 * The IERS Bulletin A table is not packaged in astroplan, and does not get downloaded at installation
 * `utils.get_IERS_A_or_workaround()` gets called from `astroplan/__init__.py`, which does one of two things:
  * If the IERS Bulletin A table was previously downloaded and cached, and the table is less than a week old, that table will be used without warning
     * If the table is more than a week old, a warning is printed, instructing the user that higher precision is available by downloading a fresh IERS A table, and giving instructions on how to do so, and the "old" table is used anyways
  * If the IERS Bulletin A table is not in the cache, then `Time._get_delta_ut1_utc`, which calculates UT1-UTC and raises an error without IERS A, gets overridden by our monkey patch `_low_precision_utc_to_ut1`. This patch uses the approximation that UT1-UTC=0 always. This approximation is good to a few seconds over the last 15 years. The same precision warning is printed as in the previous bullet, instructing the user how to download a newer table
 * When a user manually triggers the IERS A download:
  * It first checks to see if a copy already exists, and overwrites it if so
  * Download progress is obviated by the progress bar
  * The monkey patch on `Time._get_delta_ut1_utc` is reversed back to its normal state. 

### Examples
Let's say you've just downloaded astroplan and you want to run the test for the parallactic angle method, which uses the date "2015-01-01", which ordinarily must rely on the IERS A table. You haven't downloaded the IERS A table yet, so you will get a warning, but the monkey patch kicks in, the tests will pass and you don't raise any assertions.
```
% python -c "from astroplan.tests.test_core import test_parallactic_angle as f; f()"
WARNING: OldEarthOrientationDataWarning: For best precision (on the order of 
arcseconds), you must download an up-to-date IERS Bulletin A table. To do so, run:

>>> from astroplan import download_IERS_A
>>> download_IERS_A()
 [astroplan.utils]
```
Now you can copy and paste those commands and the download will begin
```
>>> from astroplan import download_IERS_A
>>> download_IERS_A()
Downloading http://maia.usno.navy.mil/ser7/finals2000A.all
|==========================================| 3.0M/3.0M (100.00%)         2s
```
and you can check that the monkey patch has been undone after download: 
```
In [11]: astroplan.utils.Time._get_delta_ut1_utc
Out[11]: <unbound method Time._low_precision_utc_to_ut1>
In [12]: download_IERS_A()
Downloading http://maia.usno.navy.mil/ser7/finals2000A.all
|==========================================| 3.0M/3.0M (100.00%)         2s
In [13]: astroplan.utils.Time._get_delta_ut1_utc
Out[13]: <unbound method Time._get_delta_ut1_utc>
```
Sound good?

##### Aside
For intrepid cloners – if you clone this branch to experiment with it, you might be thankful to have this pattern for copying and pasting for deleting your locally cached IERS A table:
```python
from astropy.utils.data import clear_download_cache
from astropy.utils.iers import IERS_A_URL
clear_download_cache(IERS_A_URL)
```


olebole:
Does this also work if the package is installed system-wide?
What about downloading the table via crontab, at least as an option?

cdeil:
I think this solution is OK, my main concern has been addressed: astroplan is usable without internet connection. So :+1:  to merge.

IMO the Time monkeypatching and the IERS table download should be handled in Astropy.
So I'd prefer if an issue or PR is opened against the Astropy repo, this code is moved to `astroplan/extern`, and a note is left at the top linking to the Astropy PR.

I also think it would be good to add a section in the high-level docs for this, explaining to the user what the IERS A table is, when it's downloaded or updated, for which `Time` or `astroplan` functions it's used, and roughly what precision loss results if it's not present. @bmorris3 – you have written a lot about this in your blog post and in the original descriptions of the three PRs you've made for this, but a user won't find this and all users will see the download happening or get the warning and wonder if they did something wrong.
Since we want to get the release out the door, this docs section could be very short, but I think it should be there and probably also put the URL to that explanation in the warning.

I don't know if it's possible to implement tests for this, excercising the different scenarios. It's propably too hard and not the best use of @bmorris3's time now, but this IERS table handling is the one thing that has me worried when we ship `astroplan 0.1`, so if implementing at least some tests is possible, please do.

cdeil:
@eteq – I've assigned this issue to you for final review and merging.

bmorris3:
Re @olebole: 
> Does this also work if the package is installed system-wide?

I believe it should work no matter where the package is installed. The caching is managed by 
`astropy.utils.data` which seems to be pretty stable.

> What about downloading the table via crontab, at least as an option?

A user can write a simple script like this to execute by crontab weekly, which will download a new IERS A table and overwrite the one in the cache: 
```python
from astroplan import download_IERS_A
download_IERS_A()
```
Is that what you mean?

bmorris3:
@cdeil It might be more verbose than you were looking for, but I've made some new high level docs for IERS A in e1d3c32.

docs/tutorials/iers_a.rst
```diff
@@ -0,0 +1,70 @@
...
+.. _iers_a-What_are_IERS_Bulletins_A_and_B:
+
+What are IERS Bulletins A and B?
+================================
+
+The International Earth Rotation and Reference Systems Service (IERS) is
+responsible for measuring the Earth's orientation as a function of time, and
+making predictions for the Earth's orientation in the near future (accounting
+for scheduled leap seconds). The data products released by the IERS used by
+astroplan are the IERS Bulletins A and B. IERS Bulletin B is a table with Earth
```
cdeil: The title of this section is "What are A and B", but then you only say what "B" is.
My suggestion would be to make two bullets here, one for A and B, saying what they are and how they differ.
Maybe also link to http://astropy.readthedocs.org/en/latest/time/index.html?highlight=iers#transformation-offsets ?

docs/tutorials/iers_a.rst
```diff
@@ -0,0 +1,70 @@
...
+rotation does not change in a predictable way, we must rely on observations
+of the Earth's orientation as a function of time to accurately calculate
+UT1-UTC.
+
+.. _iers_a-What_are_IERS_Bulletins_A_and_B:
+
+What are IERS Bulletins A and B?
+================================
+
+The International Earth Rotation and Reference Systems Service (IERS) is
```
cdeil: Add link to IERS page.

docs/tutorials/iers_a.rst
```diff
@@ -0,0 +1,70 @@
...
+atomic clock, UTC, one that matches up with the Earth's rotation, UT1. The
+difference between UT1 and UTC is constantly changing as the Earth's rotation
+changes and as leap seconds get added to UTC to compensate.
+
+In order to accurately predict the apparent position of a celestial object from
+the Earth, for example in altitude and azimuth coordinates, one needs to know
+the orientation of the Earth at any point in time, and since the Earth's
+rotation does not change in a predictable way, we must rely on observations
+of the Earth's orientation as a function of time to accurately calculate
+UT1-UTC.
```
cdeil: I think this description in "Background" is excellent ... keeps it very simple, but addresses the core point.
Maybe add your plot of UT1-UTC here, or link to your blog post?

docs/tutorials/iers_a.rst
```diff
@@ -0,0 +1,70 @@
...
+
+The International Earth Rotation and Reference Systems Service (IERS) is
+responsible for measuring the Earth's orientation as a function of time, and
+making predictions for the Earth's orientation in the near future (accounting
+for scheduled leap seconds). The data products released by the IERS used by
+astroplan are the IERS Bulletins A and B. IERS Bulletin B is a table with Earth
+orientation observations from the recent few decades up through nearly the
+present time. `Astropy` relies on IERS B to compute UT1-UTC, and by default
+will throw an error if you try to compute UT1-UTC for a time that is outside
+the bounds of the IERS Bulletin B table.
```
cdeil: I think here it would be good to be more explicit for the user what this implies.
Without astroplan, just using Astropy, for which time range can and can't I compute UT1 given a UTC time?
I'd even be +1 to give the code example that shows the error (not the full traceback, just the one line, I think it was an IndexError?).

Then say that if the user imports astroplan, one can compute UT1 (and Alt / Az) for any time in the past and future?
(because of the monkey patching)

How do the results differ when IERS A or B are present?

astroplan/utils.py
```diff
@@ -1,10 +1,118 @@
...
+from astropy.utils.data import download_file, clear_download_cache
+from astropy.utils import iers
+from astropy.time import Time
+import astropy.units as u
+from .exceptions import OldEarthOrientationDataWarning
+from astropy.utils.data import (_get_download_cache_locs, CacheMissingWarning,
+                                _open_shelve)
 
-__all__ = ["_mock_remote_data"]
+__all__ = ["_mock_remote_data", "get_IERS_A_or_workaround", "download_IERS_A"]
```
cdeil: Remove `_mock_remote_data` from `__all__` because it's not part of the public API?

astroplan/utils.py
```diff
@@ -1,10 +1,118 @@
...
+
+IERS_A_WARNING = ("For best precision (on the order of arcseconds), you must "
+                  "download an up-to-date IERS Bulletin A table. To do so, run:"
+                  "\n\n"
+                  ">>> from astroplan import download_IERS_A\n"
+                  ">>> download_IERS_A()\n")
```
cdeil: +1 to add a link to the online docs page describing this in the warning.

cdeil:
The `tutorials/iers_a.rst` page isn't connected to the docs at the moment, I think.
Is that on purpose to not conflict with other in-progress PRs?

@eteq – Can you comment if the implementation of this looks good to you?

eteq:
(will look at this today sometime)

bmorris3:
Re @cdeil:
 
> The tutorials/iers_a.rst page isn't connected to the docs at the moment, I think.

This was intentional because I think the pages that should link to this tutorial are ones getting merged in from #42 (could be wrong here). I'm waiting to see the final structure of all of those docs before figuring out where to insert mine.

docs/tutorials/iers.rst
```diff
@@ -0,0 +1,135 @@
...
+wall or the calendar, but the Earth's rate of rotation varies measurably
+on human timescales. As a result, the time measured by an atomic clock and
+the time that you would infer from measuring how long it has been since the
+last solar noon are constantly becoming offset from one another with
+stochastic changes in Earth's moment of inertia and slowly acting tidal forces.
+
+For this reason, there is a time system that keeps track of seconds like an
+atomic clock, UTC, and one that matches up with the Earth's rotation, UT1. The
+difference between UT1 and UTC is constantly changing as the Earth's rotation
+changes and as leap seconds get added to UTC to compensate.
```
eteq: I know you're trying to be pedagogical here and don't want to get into the details here: that's very good! This is easy-to-read, which is probably going to be good for astroplan users.

But I think you need to make a modification here to make it clear that UTC is *not* just a continuous progression of atomic clock time - that's what TAI is.  UTC is that plus a constant offset that changes every so often to make it stay within a second of UT1.  This isn't just pedantic - it's important for users to know that any leap seconds that get added in between now and whatever time they're thinking about do *not* get accounted for because we can't accurately predict leap seconds.

astroplan/utils.py
```diff
@@ -1,10 +1,118 @@
...
+    # astropy.utils.data.download_file()
+    try:
+        dldir, urlmapfn = _get_download_cache_locs()
+    except (IOError, OSError) as e:
+        msg = 'Remote data cache could not be accessed due to '
+        estr = '' if len(e.args) < 1 else (': ' + str(e))
+        warnings.warn(CacheMissingWarning(msg + e.__class__.__name__ + estr))
+        return False
+    with _open_shelve(urlmapfn, True) as url2hash:
+        if url_key in url2hash:
```
eteq: I'm getting an error on this line when I try this right now.  It appears to stem from `url_key` being a unicode string instead of a regular string, but `shelve` doesn't accept unicode as keys.  So I *think* it will work to just do ``str(url_key)``?  It worked for me, anyway...

bmorris3: So to clarify, you're suggesting that line 68 should become: 
```python
        if str(url_key) in url2hash:
```
?

cdeil: Add test?

docs/tutorials/iers.rst
```diff
@@ -0,0 +1,135 @@
...
+    ax.axhline(0, color='gray', ls='--', lw=2)
+
+    ax.plot_date(time_range.plot_date,
+                 DUT1_a, '-', label='IERS Bulletin A + extrapolation')
+    ax.plot_date(time_range.plot_date[measurements_from_b],
+                 DUT1_b[measurements_from_b], 'r--', label='IERS Bulletin B')
+
+    ax.set(xlabel='Year', ylabel='UT1-UTC [seconds]')
+    ax.legend(loc='upper right')
+    plt.show()
```
eteq: I like this example and think it's quite instructive. One, though based on my comment above, though: you might add a third line that's UT1-TAI to this (and adjust the y label to note this as well as the y-limit to go up to a few seconds or so).  For me at least, that makes it very clear that UTC is TAI+integer seconds and thereby justifies pretty clearly why we have these crazy scales.

docs/tutorials/iers.rst
```diff
@@ -0,0 +1,135 @@
+:orphan:
+
+.. doctest-skip-all
+
+**********************************
+Using and updating IERS Bulletin A
```
eteq: @bmorris3 - based on our slack conversation, you might consider adding a section here about how to silence all the warnings we were talking about.  As we concluded there, not critical now, but it seems like a logical place.

astroplan/utils.py
eteq: yes, @bmorris3, that's what seemed to work for me.  There's the possibility that this will get us into trouble in ways I don't really understand (a lot of py 2.x unicode stuff is basically just "try it till it works" in my mind...), but this at least will keep shelve from freaking out.

As for a test, I'm not sure how to really test it robustly on different platforms...

If it helps, though, I guess the "test" gets run anytime someone does any tests because the import of astroplan will invoke this, right?

docs/tutorials/iers.rst
bmorris3: I tried this and found it a bit more confusing. For the purposes of this high level example, I think sticking to UTC and UT1, which are the names most likely to show up in warnings, errors and tracebacks, might be sufficient.

eteq: Ok, fair enough

eteq:
@bmorris3 - Ok, I think this is good once you merge bmorris3/astroplan#3 (assuming you like it and the tests then pass)

PR #90, opened 2015-09-02
Fix pytest-mpl and pytest usage

### LINKED ISSUE #86: tests cannot be run if matplotlib and pytest-mpl are not installed
I've encountered an issue that I think is caused by ``pytest-mpl``, which it looks to me was integrated into astroplan in #13 (@cdeil ?): if I try to run the tests for astroplan on a machine where matplotlib is installed but ``pytest-mpl`` is *not*, I see the following error before the tests even start:
```
% python setup.py test
running test
running build
running build_py
usage: -c [options] [file_or_dir] [file_or_dir] [...]
-c: error: unrecognized arguments: --mpl --mpl-baseline-path=astroplan/plots/tests/baseline_images
```
I think this is due to the line in the pytest section of ``setup.cfg`` which says ``addopts = --mpl --mpl-baseline-path=astroplan/plots/tests/baseline_images``.  Presumably that fails if ``pytest-mpl`` is absent.  Is there a way to still use ``pytest-mpl`` when present but just skip the tests (and not have the command line failure) if it's absent?

cc @astrofrog as I imagine this might show up in other ``pytest-mpl`` situations

eteq:
This PR makes two different changes dealing with the testing suite: First, it fixes #86 by programmatically deciding to turn pytest-mpl on or off based on whether it's installed or not. (cc @cdeil for this.)  @astrofrog my want to weigh in on whether this is a recommended way or not, too, but I think we want to merge this ASAP to allow for 0.1 to go forward, even if we go with some better way in 0.2.

Second, it changes all ``import pytest`` statements to instead be ``from astropy.tests.helper import pytest``.  The reasoning for this is that we don't want ``pytest`` to be a necessary dependency for users to run the tests.  So using the pytest packaged with astropy lets us do this without having to include pytest ourselves.  Note that this machinery still allows use of any system- installed pytest by doing invoking the tests as ``ASTROPY_USE_SYSTEM_PYTEST=1 python setup.py test`` (cc @jberlanga and @bmorris3 because I think you both might have done ``import pytest`` simply because you didn't know about this option.)

### DIFF
--- a/astroplan/conftest.py
+++ b/astroplan/conftest.py
@@ -21,3 +21,14 @@ except NameError:  # needed to support Astropy < 1.0
 # within the tests
 from .utils import _mock_remote_data
 _mock_remote_data()
+
+def pytest_configure(config):
+    try:
+        import matplotlib
+        HAS_MATPLOTLIB = True
+    except ImportError:
+        HAS_MATPLOTLIB = False
+
+    if HAS_MATPLOTLIB and config.pluginmanager.hasplugin('mpl'):
+            config.option.mpl = True
+            config.option.mpl_baseline_path = 'astroplan/plots/tests/baseline_images'
--- a/astroplan/conftest.py
+++ b/astroplan/conftest.py
@@ -25,10 +25,11 @@ _mock_remote_data()
 def pytest_configure(config):
     try:
         import matplotlib
-        HAS_MATPLOTLIB = True
+        import nose  # needed for the matplotlib testing tools
+        HAS_MATPLOTLIB_AND_NOSE = True
     except ImportError:
-        HAS_MATPLOTLIB = False
+        HAS_MATPLOTLIB_AND_NOSE = False
 
-    if HAS_MATPLOTLIB and config.pluginmanager.hasplugin('mpl'):
+    if HAS_MATPLOTLIB_AND_NOSE and config.pluginmanager.hasplugin('mpl'):
             config.option.mpl = True
             config.option.mpl_baseline_path = 'astroplan/plots/tests/baseline_images'
--- a/.continuous-integration/travis/setup_dependencies_common.sh
+++ b/.continuous-integration/travis/setup_dependencies_common.sh
@@ -20,13 +20,6 @@ fi
 # CORE DEPENDENCIES
 conda install --yes pytest Cython jinja2 psutil pytz
 
-# For now testing requires matplotlib and pytest-mpl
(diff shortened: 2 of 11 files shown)


astroplan/conftest.py
```diff
@@ -21,3 +21,14 @@
 # within the tests
 from .utils import _mock_remote_data
 _mock_remote_data()
+
+def pytest_configure(config):
+    try:
```
cdeil: Probably should also try importing `nose` here.
That's the matplotlib test runner, and if I remember correctly, there's a cryptic error message from pytest if matplotlib is there, but nose.

cdeil:
@eteq – Thanks for figuring this out and making the pull request.

Maybe mention that matplotlib, nose and pytest-mpl are needed as optional dependencies to run the image comparison tests?
http://astroplan.readthedocs.org/en/latest/installation.html#testing

And for the travis-ci test matrix ... the installation commands here should probably be moved down to the optional dependencies section, so that travis-ci has a build that runs the tests, but matplotlib / nose / pytest-mpl are not present:
https://github.com/astropy/astroplan/blob/master/.continuous-integration/travis/setup_dependencies_common.sh#L23

cdeil:
cc @bsipocz , since she's setting up image comparison tests for photutils.

eteq:
Thanks @cdeil - I think I addressed all your comments here - we'll see if travis is happy soon enough...

cdeil:
:+1:

bmorris3:
@eteq It's failing in [places](https://travis-ci.org/astropy/astroplan/jobs/78472582#L587) for not being able to access matplotlib.

eteq:
@bmorris3 - ahh, good we changed the deps as @cdeil suggested - those tests are missing the `HAS_MATPLOTLIB` test.  Will update momentarily.

eteq:
@bmorris3 - ah, good we changed this as @cdeil suggested... this is a genuine failure due to a top-level import of matplotlib - which we should *not* do so that users can use astroplan even if matplotlib is not installed.  Will fix shortly.

eteq:
ok, hopefully this one will actually pass - I just moved the matplotlib imports inside the function instead of at module-level.  So then everything is importable (although of course the plotting functions will still fail immediately if matplotlib isn't installed)

bmorris3:
@eteq Passed! Good to merge?

eteq:
Yes, hooray! (will merge it myself now)

Issue #76, opened 2015-08-28
FixedTarget mock is active even when remote-data flag is used


bmorris3:
In #62, we came up with a mock for `FixedTarget.from_name` when the internet is not accessible during tests. That mocked `from_name` method currently gets called during the tests whether or not there is an internet connection, but should only be used when the `--remote-data` flag is not used.

eteq:
I'd suggest this approach: Write two tests, one which is marked with ``@remote_data``, and the other with ``@pytest.mark.skipif('...')``.  Most of the "guts" of the test can be in a private function called by each of those tests, and you can just use the two versions to do the mocking or not.

The one catch here is the ``...`` in `skipif`.  I'm not exactly sure how to ask pytest "is remote_data on or not".  If you look deep into the astropy testing machinery, there's [a line](https://github.com/astropy/astropy/blob/master/astropy/tests/pytest_plugins.py#L134) that does ``if not config.getoption('remote_data')``... but the part I'm not sure about is how to get a `config` object actually in a testing module (the usage in the testing machinery is inside a plugin).  There must be some way to do so, but I don't have time right now to delve through the pytest docs to find out...

PR #93, opened 2015-09-03
Add tests with remote-data on

eteq:
This adds just adds two tests to the travis test matrix that have ``--remote-data`` turned on.  This is normally pretty safe on travis because it has to be connected to the internet anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,15 +63,9 @@ matrix:
         # build (with latest numpy). We also note the code coverage on Python
         # 2.7.
         - os: linux
-          env: PYTHON_VERSION=2.7 SETUP_CMD='test -V --coverage' OPTIONAL_DEPS=true
+          env: PYTHON_VERSION=2.7 SETUP_CMD='test --remote-data -V --coverage' OPTIONAL_DEPS=true
         - os: linux
-          env: PYTHON_VERSION=3.4 SETUP_CMD='test -V' OPTIONAL_DEPS=true
-
-        # Also try with remote-data option (and do optional_deps in case there's overlap there)
-        - os: linux
-          env: PYTHON_VERSION=2.7 SETUP_CMD='test --remote-data -V' OPTIONAL_DEPS=true
-        - os: linux
-          env: PYTHON_VERSION=3.4 SETUP_CMD='test --remote-data-V' OPTIONAL_DEPS=true
+          env: PYTHON_VERSION=3.4 SETUP_CMD='test --remote-data -V' OPTIONAL_DEPS=true
 
         # Try older numpy versions
         - os: linux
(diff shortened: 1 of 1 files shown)


cdeil:
Because we're mocking `FixedTarget.from_name` for the tests, it wasn't clear to me if there are any tests that only run when `--remote-data` is used.

There are two:
```
astroplan/tests/test_core.py:57: test_FixedTarget_from_name SKIPPED
astroplan/tests/test_utils.py:15: test_iers_download SKIPPED
```

@bmorris3 – Should  `test_FixedTarget_from_name` be skipped if `--remote-data` is not used ... didn't you mock that?

@eteq -- I'm 50:50 on adding this. A few weeks ago we had for ~ 1 week failing travis-ci tests on many PRs, because the IERS table access was flaky, it was annoying.

If you do add it, I'd suggest using `--remote-data` for the test that computes the coverage.
The IERS table code is currently reported as untested here (https://coveralls.io/github/astropy/astroplan), but presumably it's at least tested some via `test_iers_download`.

bmorris3:
@cdeil `test_FixedTarget_from_name` was meant to test that names get resolved and the `FixedTarget` object is constructed appropriately. I don't think there'd be a point to keep the test on without remote-data, unless it was to test that the mock was working.

@eteq: does this affect #76?

cdeil:
> unless it was to test that the mock was working.

That's what I was thinking ... shouldn't we just test the mock as well?
(Sorry if this doesn't make sense ... I didn't look / think closely what exactly is happening)

bmorris3:
@cdeil I guess we could. I suppose all tests that use `from_name` test that the mock is working, in a sense.

eteq:
Ah, you both raises a good point that it's not really being tested because of the mock.  I think I figured out a solution to  #76, though, and will roll that into this shortly.

eteq:
Actually, it ended up making more sense to implement the #76 fix separately (#98) so that it can be merged even if we don't want to do this.  I've just rebase this over #98 so that we can see if that fixed the test failure (I'm not sure, actually...)

Note that this means this should *not* get merged until #98 is in, though, as I'll want to rebase it over master instead of over #76 if we do go with this.

bmorris3:
@eteq Now this is [failing](https://travis-ci.org/astropy/astroplan/jobs/78631832#L579) with 
```
AssertionError: <class 'astroplan.exceptions.OldEarthOrientationDataWarning'> not found in []
```

eteq:
I rebased this now against the latest #76 so we can see if that fixed the issues here.

eteq:
Now rebased on master because #98 is merged

bmorris3:
@eteq There was a [failure](https://travis-ci.org/astropy/astroplan/jobs/78675809#L580) on travis:
```
ImportError: --coverage requires that the coverage package is installed.
```
Think that run needs to be restarted?

bmorris3:
My fix worked, all tests pass, merging!

PR #97, opened 2015-09-03
Harmonize warnings

eteq:
This adjusts a few areas of code so that astroplan consistently uses customized warning classes.  This is useful because it makes it easy to have fine-grained control over the warnings filter so that users can easily hide warnings that they actually expect to see.

It also adjusts the existing warnings to descend from a base `AstroplanWarning`.  Yet another layer that makes warnings control easier.

### DIFF
--- a/astroplan/exceptions.py
+++ b/astroplan/exceptions.py
@@ -8,19 +8,22 @@ __all__ = ["TargetAlwaysUpWarning", "TargetNeverUpWarning",
            "OldEarthOrientationDataWarning", "PlotWarning",
            "PlotBelowHorizonWarning"]
 
-class TargetAlwaysUpWarning(AstropyWarning):
+class AstroplanWarning(AstropyWarning):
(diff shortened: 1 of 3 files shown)


bmorris3:
:+1: Looks good, waiting for tests to pass.

cdeil:
Does it make sense to add `Astroplan` at the start of each warning, i.e. have `AstroplanPlotWarning` instead of `PlotWarning` to make it clearer that this is coming from astroplan?

@eteq If you think no, please just go ahead and merge!

eteq:
@cdeil - I think not because by making them all `AstropyWarning` descendents, you get the astropy warning infrastructure, which already tells you it's coming from astroplan.  E.g.: 
```
WARNING: PlotBelowHorizonWarning: Target "Regulus" is below the horizon at time: 2457235.97917 [astroplan.plots.sky]
```
so I think simpler names are better to not duplicate information here.  Merging!

Issue #96, opened 2015-09-03
FixedTarget.from_name shows up with Mock docstring in the docs


cdeil:
@bmorris3 – Our mock solution is mocking us ... didn't think of this issue:
http://astroplan.readthedocs.org/en/latest/api/astroplan.FixedTarget.html?highlight=mock#astroplan.FixedTarget.from_name

@bmorris3 – Why is the mock needed again for the Sphinx build?
The doctests are something different, that shouldn't require it to be active during the docs build, right?
Is it because we have `.. plot::` directives that use `FixedTarget.from_name`?

For those that didn't follow that ... the mock was added in #62 and it's here:
https://github.com/astropy/astroplan/blob/master/docs/conf.py#L181

Another open mock related issue: #76

PR #98, opened 2015-09-03
Fix 76

### LINKED ISSUE #76: FixedTarget mock is active even when remote-data flag is used

eteq:
This fixes #76 by adjusting the `from_name` mocking to turn on or off depending on whether a test is ``remote-data`` or not. I can confirm via the method of unplugging my ethernet cord that after this the remote_data tests fail if the internet is off.

### DIFF
--- a/astroplan/conftest.py
+++ b/astroplan/conftest.py
@@ -1,7 +1,14 @@
-# this contains imports plugins that configure py.test for astropy tests.
-# by importing them here in conftest.py they are discoverable by py.test
-# no matter how it is invoked within the source tree.
+"""
+This file contains functions that configure py.test like astropy but with
+additions for astroplan.  Py.test looks for specially-named functions
+(like  ``pytest_configure``) and uses those to configure itself.
 
+Here, we want to keep the behavior of astropy while *adding* more for astroplan.
+To do that, in the functions below, we first invoke the functions from astropy,
+and then after that do things specific to astroplan.  But we also want astropy
+functionality for any functions we have *not* overriden, so that's why the
+``import *`` happens at the top.
+"""
 from astropy.tests.pytest_plugins import *
 
 # also save a copy of the astropy hooks so we can use them below when overriding
@@ -46,6 +53,12 @@ def pytest_configure(config):
             config.option.mpl_baseline_path = 'astroplan/plots/tests/baseline_images'
 
 def pytest_runtest_setup(item):
+    """
+    This overrides the tests so that if they are marked ``remote_data`` they get
+    run without any mocking of functions, but if they are not, then the mocking
+    happens.  This means that functionality that uses mock data should have both
+    a ``remote_data`` test *and* a separate one that is not ``remote_data``.
+    """
     if hasattr(astropy_pytest_plugins, 'pytest_runtest_setup'):
         # sure ought to be true right now, but always possible it will change in
         # future versions of astropy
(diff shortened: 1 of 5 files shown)


cdeil:
The mock is being called from `docs/conf.py` leading to this error:
https://travis-ci.org/astropy/astroplan/jobs/78631829#L649

See  #96

cdeil:
There's also this fail!?
https://travis-ci.org/astropy/astroplan/jobs/78631828#L402

eteq:
@cdeil - that second failure is a slightly hard-to-parse indicator that a warning isn't being raised that should be.  Investigating now.

As for the second, thanks for the tip there - I didn't realize the docs used the mock... Will fix.

cdeil:
As I said in #96  ... I'm not actually sure any more the Sphinx build needs the mock ...

eteq:
Ok, will check that and just remove if that's the case.

eteq:
Turns out this is fixing a problem that I created in #90: it turns out that the way we're doing ``conftest.py``, I was turning *off* a lot of the built-in astropy behavior.  The latest commits turn it back on, and also fix a problem that had snuck in because this stuff was off.

So it's possible this will fail somewhere in some apparently unrelated place because it has substantial testing side-effects

But I think this now fixes #96  simply by not doing the mocking (it appears to not be necessary right now - I built the docs without warnings without an internet connection)

eteq:
FYI, @jberlanga, as an aside: this is why I think you were seeing weird changes in the beahvior of ``summer_triangle.rst`` in #87 : between when you made it and when you rebased, the testing infrastructure (uintentionally) changed.  So that's why I'm now skipping a lot more lines then you did there.

cdeil:
@eteq – I find this quite complicated to understand.

Could you please add a few sentences explaining the logic of what is happening in `conftest.py`, e.g. in a docstring at the top of the file?

Is it as simple as this?
* Tests that use `from_name` should **not** be marked with `@remote_data`
* If tests are run with `--remote-data`, the normal `from_name` is called which accesses the internet.
* If tests are run without `--remote-data`, the mock `from_name` is called which which does not access the internet.

eteq:
Alright, the tests passed one commit back, and the last commit adds an explanation as @cdeil requested, so this should be good to go.

To answer your sepcific quetsion here, @cdeil, the idea after this change is that something with ``from_name`` should have *two* tests: if ``remote_data`` is marked, then the test will run actually accessing the internet, but if it is not, then the mock will be used.  So you'd write a remote_data test specifically to test the internet-based functionality (i.e. actually looking up some location that's *not* in the mocks), while the rest would be for when you just want ``from_name`` as a convenient way to create a `FixedTarget` or the like (especially for the docs we'll want this once the doc tests are up and running smoothly).

cdeil:
@eteq – Hmm.

In your setup you say that the test should be duplicated (once with, once without the `@remote_data` decorator), but the test that's of interest here `astroplan/target.py::astroplan.target.FixedTarget.from_name`, currently isn't duplicated.

I'll try to come up with a solution that doesn't require duplicating the test in #103.
If that doesn't work, I can duplicate the test there, so that the mock actually is tested.

PR #103, opened 2015-09-04
Simplify pytest remote data mock setup

cdeil:
This is an attempt to simplify the pytest remote data mock setup that was introduced in #98.

Please see https://github.com/astropy/astroplan/pull/98#issuecomment-137570869 and https://github.com/astropy/astroplan/pull/98#issuecomment-137680812 as well as the 
docstring of  `_mock_remote_data` in this PR where I try to explain how it works and why it's an improvement.

### DIFF
--- a/astroplan/conftest.py
+++ b/astroplan/conftest.py
@@ -18,18 +18,17 @@ import warnings
 from .utils import _mock_remote_data, _unmock_remote_data
 from .exceptions import AstroplanWarning
 
-## Uncomment the following line to treat all DeprecationWarnings as
-## exceptions
+# Comment out this line to avoid deprecation warnings being raised as exceptions
 enable_deprecations_as_exceptions()
 
-## Uncomment and customize the following lines to add/remove entries
-## from the list of packages for which version numbers are displayed
-## when running the tests
-try:
-    PYTEST_HEADER_MODULES['pyephem'] = 'pyephem'
-    del PYTEST_HEADER_MODULES['h5py']
-except NameError:  # needed to support Astropy < 1.0
-    pass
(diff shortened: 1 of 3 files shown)


cdeil:
@eteq Can you please review this?

eteq:
@cdeil - I'm not sure this actually solves the problem #98 was initially meant to fix.  With this version, you *always* turn on mocking of ``--remote-data`` is off, and not if it's on.  That means there's no way to test the mocked versions if ``remote_data`` is on.  

My thinking is that that's a *feature*, not a bug.  That is, it's better to write ``remote_data`` and non-``remote_data`` tests as distinct and different entities.  Then it's clearer what the origin of failures are: the actual ``remote_data`` behavior (i.e. downloads), or the *other* pieces that just use remote downloads incidentally.

To be more concrete, here we should have tests that actually *test* if ``from_name`` downloads the correct result, in addition to tests (mostly doctests) that use ``from_name`` just to make things easier.  We want the latter to use the mock results regardless of ``remote_data`` status, and the former should *never* use mock data.  The original implementation does that, but in this version, if you turn ``remote-data`` on and SIMBAD is down, ``from_name`` will yield tons of spurious failures, rather than just one.  So I would advocate keeping ``pytest_runtest_setup`` as it was in #98.

The rest of it are good changes, though.

eteq:
After a lengthy discussion on slack, @bmorris3 broke the tie and said we should switch to this behavior.  So I'll merge this PR, and after this, ``--remote-data`` means `from_name` *always* goes to the internet, regardless of whether a test is marked `remote_data` or not.

PR #104, opened 2015-09-04
test_image_example fails when run via astroplan.test()

cdeil:
I noticed that `test_image_example` fails when run via `astroplan.test()`:
https://gist.github.com/cdeil/f5dc761311130ff1062a#file-gistfile1-txt-L1

I'll add a travis-ci build here that tests for this case and then will try to fix it.
(My guess is that the issue is that the test image filename needs to be declared in the test decorator, so that it's installed in the tmp test folder by pytest-mpl before running the test?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,6 +59,14 @@ matrix:
         - os: linux
           env: PYTHON_VERSION=3.4 SETUP_CMD='test -V'
 
+        # Try running tests from an installed version via the `astroplan.test()` function
+        # Because of the way this travis config file is written this is a bit cryptic.
+        # - $MAIN_CMD is used to trigger this case in the install section below
+        # - $SETUP_CMD is used to communicate which dependencies should be installed
+        - os: linux
+          env: PYTHON_VERSION=3.4 $MAIN_CMD='test installed' SETUP_CMD='test --remote-data -V' OPTIONAL_DEPS=true
+
+
         # Now try with all optional dependencies on 2.7 and an appropriate 3.x
         # build (with latest numpy). We also note the code coverage on Python
         # 2.7.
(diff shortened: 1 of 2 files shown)


cdeil:
Here's the new build on travis-ci that shows this issue:
https://travis-ci.org/astropy/astroplan/jobs/78726956#L726

cdeil:
This is weird ... travis-ci passed, although there's this fail:
https://travis-ci.org/astropy/astroplan/jobs/78727693#L724

Does anyone know what's wrong with this `.travis.yml` file ... why doesn't this make the build fail?

I also wasn't able to figure out yet how to get `pytest-mpl` working.
I tried adding the filename to the decorator, this doesn't change anything:
```
@pytest.mark.mpl_image_compare(filename='test_image_example.png')
```

The issue seems to be that `baseline_dir = '/tmp/astroplan/plots/tests/baseline_images'` and `baseline_image_ref = '/tmp/astroplan/plots/tests/baseline_images/test_image_example.png'`, but that's not the correct location of the image if the tests are run from an installed version via `astroplan.test()`

@astrofrog or @bsipocz or anyone -- Do you know how to fix this?

I'll try to have another look here in ~ 5 hours, but I'm not sure, so if anyone wants to take this over, please go ahead. For now I've labeled this 0.1, because this test fail will occur for all users that pip-install `astroplan` and run the tests as described in the docs. But feel free to move to 0.2 if you prefer.

eteq:
@cdeil - the test succeeded because the exit code of ``python -c 'import astroplan; astroplan.test()'`` will always be 0 - you need to do something like ``sys.exit(1)`` to get a non-zero exit code out.  I suggest doing this:``python -c 'import astroplan, sys; sys.exit(astroplan.test())'`` - that will correctly give a non-0 exit code if the tests fail

eteq:
(no idea about the actual problem itself, though... perhaps we should just turn off the image comparison test for 0.1)

cdeil:
@eteq - Thanks for the tip about the exit code.
I've put that in and turned the image comparison tests off.

So if travis-ci passes, I'd say let's merge this so that 0.1 can go out, and figure out the image comparison test setup later.

eteq:
:+1:

eteq:
(also, I went ahead and canceled all but the last of the travis and appveyor builds, since it's the last one we really care about anyway)

eteq:
Oops, looks like there was a typo - I fixed that and in https://travis-ci.org/eteq/astroplan/builds/78827200 you can see the result - all passing!  So I'll go ahead and merge manually with that extra change (https://github.com/eteq/astroplan/commit/ba1ecbb6d2a16bb74ea7f92461722632277b0ef8)

PR #105, opened 2015-09-04
Numeric earth coords

adrn:
Ok...I'll probably get shot down on this one, but I still want to see numbers expressed as numbers! This is for 2 silly reasons: 

1. I'm making an interactive javascript visualization of the location of all observatories and I don't want to have to parse these strings to numbers, and
2. When I make a sites table in the docs (for 0.2), I want to easily generate Google earth or Google maps links from the site location for each row in the table.

I'm not concerned about precision: floating point precision correspond to ~10 nanometer differences in te site position. This also makes expression of any quantities in the JSON file consistent (e.g., before we had `elevation_meters`, now all quantities have a separate `_unit` keyword that can be passed directly into an `astropy.units.Unit()` call).

### DIFF
--- a/astroplan/data/observatories.json
+++ b/astroplan/data/observatories.json
@@ -1,287 +1,371 @@
 {
-    "BAO": {
+    "tona": {
+        "source": "IRAF Observatory Database", 
+        "elevation": 0, 
+        "name": "Observatorio Astronomico Nacional, Tonantzintla", 
+        "longitude_unit": "degree", 
+        "latitude_unit": "degree", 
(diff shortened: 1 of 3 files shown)


bmorris3:
You won't get shot down for the PR, I'm glad you made it. But you will get shot down for labeling this a bug.

bmorris3:
Looks good, thanks for doing what I was too stubborn to rewrite :metal:.

eteq:
There, I added enhancement so the record will show it as suitably ambiguous ;)

Issue #107, opened 2015-09-04
turn image comparison tests back on and get them working


eteq:
In #104 we turned the image comparison tests off because of some configuration issues, and didn't want that to interfere with 0.1.  They should be turned back on (and fixed!) in 0.2

cc @cdeil @bsipocz (also #81, although I think that's different from the original issue in #104 ?)

wtgee:
Marking Issue as stale and closing. Looks like it was fixed in #115. Please re-open if still affected by this Issue.

PR #115, opened 2015-09-28
Simple finder chart

bmorris3:
This is an implementation of a simple finder chart, first implemented with examples [here](https://github.com/bmorris3/gsoc2015/blob/master/finder_chart.ipynb) (see also comments [here](https://github.com/bmorris3/gsoc2015/commit/88e06aebb7ee3cd71f4b9406302f7d62fc5266db)).

@jberlanga / @eteq / @cdeil Should we make some basic tests for this method? I wasn't following the `astroplan.plots` testing strategy very closely.

### DIFF
--- a/astroplan/plots/finder.py
+++ b/astroplan/plots/finder.py
@@ -1,0 +1,125 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+
+import numpy as np
+
+import astropy.units as u
+
+from astropy.coordinates import SkyCoord
+from astropy.wcs import WCS
+
+@u.quantity_input(fov_radius=u.deg)
+def plot_finder_image(target, survey='DSS', fov_radius=10*u.arcmin,
+                      log=False, ax=None, grid=False, reticle=False,
+                      style_kwargs=None, reticle_style_kwargs=None):
+    """
+    Plot survey image centered on ``target``.
+
+    Survey images are retrieved from NASA Goddard's SkyView service via
+    `~astroquery.skyview.SkyView`, and plotted using WCSAxes.
+
+    If a `~matplotlib.axes.Axes` object already exists, plots the finder image
+    on top. Otherwise, creates a new `~matplotlib.axes.Axes`
+    object with the finder image.
+
+    Parameters
+    ----------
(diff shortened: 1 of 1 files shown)


cdeil:
@bmorris3 – Concerning tests for images, this should be done later (see #107), once someone figures out how to pin the plotting style across MPL versions (see https://github.com/astrofrog/pytest-mpl/issues/12).

For now, my suggestion would be that you do excercise the code you add here by adding a test and return the fig, as done in the one plot test example we have in astroplan. Then once we get pytest-mpl working smoothly, we add the decorator which does the image comparison and add the baseline image.

OK?

astroplan/plots/finder.py
```diff
@@ -0,0 +1,125 @@
...
+        ax.axhline(y=0.5*pixel_width, xmin=0.5-inner, xmax=0.5-outer,
+                   **reticle_style_kwargs)
+
+    # Labels, title, grid
+    ax.set(xlabel='RA', ylabel='DEC')
+    if target_name is not None:
+        ax.set_title(target_name)
+    ax.grid(grid)
+
+    # Redraw the figure for interactive sessions.
```
cdeil: Is it a good idea to call `ax.figure.canvas.draw()` in plotting functions to trigger a redraw for Figures in interactive sessions?
Does this imply inefficient plotting (i.e. double-draw) somehow?

Is this relevant?
http://matplotlib.org/devdocs/users/whats_new.html#interactive-oo-usage

@bmorris3 – If no-one watching this issue comes up with a good answer within ~ a day you could ping e.g. Michael Droettboom with this question.

bmorris3: This is how @jberlanga / @eteq / @adrn have the other plotting functions set up -- maybe they could illuminate the motivation for the redraw call?

jberlanga: @bmorris3 - I don't recall the exact motivation for this--something to do
with multiple plots on top of one `ax` object not necessarily showing up in
an interactive session without a redraw?  @eteq?

bmorris3: @cdeil Would you mind if I opened an issue on the `ax.figure.canvas.draw()` question, and merge this commit, which follows the form of the other plotting functions?

cdeil: :+1:

bmorris3:
How did we let me get away with not adding docs for this function? :x:

adrn:
Oops... just looking at this now...

PR #117, opened 2015-09-28
FAQ note on adding an observatory

bmorris3:
Today I got asked by a user if it's OK to add observatories to the database, and I realized there was no way users would naturally find the convenience method I wrote for this purpose. 

In this PR I've added a quick note in the FAQ on how to add an observatory to the observatory database (simply directing the user to the [`new_site_info_to_json`](https://github.com/astropy/astroplan/blob/master/astroplan/sites.py#L135) method). 

cc @eteq / @cdeil

### DIFF
--- a/astroplan/sites.py
+++ b/astroplan/sites.py
@@ -21,7 +21,7 @@ from astropy.utils.data import get_pkg_data_contents
 from astropy.coordinates import EarthLocation
 import astropy.units as u
 
-__all__ = ['get_site', 'get_site_names', 'add_site']
+__all__ = ['get_site', 'get_site_names', 'add_site', 'new_site_info_to_json']
 
 # Observatory database and list of names:
 _site_db = None
--- a/docs/faq/addobservatory.rst
+++ b/docs/faq/addobservatory.rst
@@ -1,0 +1,23 @@
+.. doctest-skip-all
+
+.. _addobservatory:
+
+********************************************************
+How can I add observatories to the observatory database?
+********************************************************
+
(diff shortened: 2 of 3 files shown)


cdeil:
@bmorris3 – I don't understand why `new_site_info_to_json` is useful.

Importing it, executing it, pasting the output in the correct file is more work than just copy & pasting & editing an entry in https://github.com/astropy/astroplan/blob/master/astroplan/data/observatories.json directly, no?

So I'm -1 (but don't care too much) on recommending `new_site_info_to_json` to users, and would instead link to that file on Github and tell them to add an entry to it.
There could be a test to check validating the format and content for all entries if that doesn't exist already.

bmorris3:
@cdeil  – I think it's less work for users to run the `new_site_info_to_json` method than it is to figure out exactly what format the JSON data are expected to be in (by context? by a note in the docs that doesn't yet exist?). This is how I entered new observatories to the database when I did it, because I didn't trust myself to copy and paste the data correctly.

> There could be a test to check validating the format and content for all entries if that doesn't exist already. 

Agreed, I'll get on that.

eteq:
@bmorris3 - it *might* make sense to wait on this until #121 is dealt with, as presumably after that the answer will change to "make a PR against astropy-data"...?

bmorris3:
I'm going to close this, since @eteq's work on astropy/astropy#4042 has made `astroplan`'s sites redundant. When #121 is done, a note will be added to FAQ to with links to astropy's docs on sites.

PR #116, opened 2015-09-28
Pin coverage to version 3.7.1 on travis-ci

cdeil:
This PR should fix the unrelated build fail in #115 by @bmorris3 :
https://travis-ci.org/astropy/astroplan/jobs/82493187#L716

This is an issue in Astropy: https://github.com/astropy/astropy/issues/4175

PR #118, opened 2015-09-29
Fix sphinx_rtd install on travis-ci

cdeil:
After merging #116 the issue came up that sphinx_rtd wasn't installed, but needed for the Sphinx build (don't know why this appeared only now)?
https://travis-ci.org/astropy/astroplan/jobs/82734139#L643

So I added a `pip install sphinx_rtd` to `setup_dependencies_common.sh`, but now I'm getting this error:
https://travis-ci.org/astropy/astroplan/jobs/82742624#L613

It might be a pip or conda issue or packaging issue with one of the `sphinx_rtd` dependencies, I don't know. This seems releated: https://github.com/pypa/pip/issues/2751#issuecomment-144073543

So here I'll try installing this package differently, e.g. using pip instead of conda to install sphinx.
@astrofrog Maybe you've seen this before or have an idea what to do?

### DIFF
--- a/.continuous-integration/travis/setup_dependencies_common.sh
+++ b/.continuous-integration/travis/setup_dependencies_common.sh
@@ -53,9 +53,9 @@ fi
 # build_sphinx needs sphinx as well as matplotlib and wcsaxes (for plot_directive). 
 if [[ $SETUP_CMD == build_sphinx* ]]
 then
-  $CONDA_INSTALL Sphinx Pygments matplotlib
-  pip install wcsaxes
+  $CONDA_INSTALL matplotlib
   pip install sphinx_rtd_theme
+  pip install wcsaxes
 fi
 
 # COVERAGE DEPENDENCIES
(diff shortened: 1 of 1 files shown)


.continuous-integration/travis/setup_dependencies_common.sh
```diff
@@ -53,9 +53,9 @@ fi
 # build_sphinx needs sphinx as well as matplotlib and wcsaxes (for plot_directive). 
 if [[ $SETUP_CMD == build_sphinx* ]]
 then
-  $CONDA_INSTALL Sphinx Pygments matplotlib
-  pip install wcsaxes
+  $CONDA_INSTALL matplotlib
   pip install sphinx_rtd_theme
```
astrofrog: what about trying the same as before but with sphinx_rtd_theme in the conda command? (it's available in conda)

cdeil: travis-ci passed with what's here now: https://travis-ci.org/astropy/astroplan/builds/82747045
But OK, I'll try conda installing Sphinx and sphinx_rtd_theme instead ... if it works it could be a bit faster / more robust.

PR #120, opened 2015-10-02
Adding EclipsingFixedTarget, PrimaryEclipseConstraint

bmorris3:
In the interest of satisfying the needs of my collaborators and myself, I'm going to start pushing for features like predictions of exoplanet transits and eclipsing binary occultations, beginning with this PR. This is a commit meant to start a discussion, and I anticipate adding more to this PR (like docs!) before it gets merged.

I made a new subclass of `FixedTarget` called `EclipsingFixedTarget` which takes the following additional arguments: `epoch`, `period` and `duration`, which are the most general parameters that I can think of for a periodically eclipsing object. These properties can then be used by methods specific to these targets in other modules (like `constraints`, for example...). There are three methods on the subclass that does a computation, `mid_eclipse_time(time, [which])`, `ingress_time(time, [which])` and `egress_time(time, [which])`, which computes the nearest/next/previous time of mid-eclipse/ingress/egress relative to the input `time`.

I also made a new constraint `PrimaryEclipseConstraint` which cycles through the list of targets passed to e.g. `is_observable` or `is_always_observable` and computes when `EclipsingFixedTarget`s are eclipsing.

Opinions on implementation and approach? cc @eteq @cdeil @adrn

[FYI @adleorocks @kolbylyn – this PR will replace my old transit/eclipse predicting code soon enough!]

### DIFF
--- a/astroplan/constraints.py
+++ b/astroplan/constraints.py
@@ -22,13 +22,13 @@ import numpy as np
 # Package
 from .moon import get_moon, moon_illumination
 from .utils import time_grid_from_range
-from .target import FixedTarget
+from .target import FixedTarget, EclipsingFixedTarget
 
 __all__ = ["AltitudeConstraint", "AirmassConstraint", "AtNightConstraint",
            "is_observable", "is_always_observable", "time_grid_from_range",
            "SunSeparationConstraint", "MoonSeparationConstraint",
            "MoonIlluminationConstraint", "LocalTimeConstraint", "Constraint",
-           "observability_table"]
+           "observability_table", "PrimaryEclipseConstraint"]
 
 
 def _get_altaz(times, observer, targets,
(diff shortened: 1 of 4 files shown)


cdeil:
I think it's OK to put such things in astroplan for now, even if it's only useful to a small subset of astronomers.

If a ton of stuff accumulates over the coming ~ year we might have to re-discuss before the astroplan 1.0 release if all of that really belongs in the main astroplan namespace or if we want to re-structure it somehow. But my guess is that that won't happen anyways and we'll just end up with a somewhat random collection of things that are strongly biased by the interests of a few developers / user requests.

(I'll unsubscribe from this issue now because I won't have time to follow the detailed discussions / code review.)

kolbylyn:
One feature that would be particularly useful to people taking multiple observations of EBs would be something where you could provide a text file of observation dates and the function would return a mock eclipse light curve with labels/arrows showing you at what phase you have data.

Really just something quick like this:
screen shot 2015-10-02 at 2 07 16 pm

bmorris3:
@kolbylyn – this brings up the point that we will also want an object to contain binaries that don't eclipse, and perhaps the `EclipsingFixedTarget` should be a subclass (special case) of some binary target. I need to think about this.

bsipocz:
@bmorris3 - I think an additional and useful addition would be the to add ``SecondaryEclipseConstraint``, but then more orbital elements should be added (e.g. eccentricity).

bmorris3:
Agreed @bsipocz. I was starting this way because it's easiest to define orbits by an epoch+period, but it's definitely not the most generic way. I'm going to start another branch and approach this more generally, but I'll leave this one open for more comments.

bsipocz:
On the other hand I'm not that convinced about @kolbylyn's suggestion:
Predicting times for arbitrary phases is one thing but visualize data already taken goes beyond a generic observation planning package and more about data bookkeeping, etc.

bmorris3:
@bsipocz – The precedent within `astroplan` so far would suggest that exposing functions to easily grab phase would be appropriate, as I think @kolbylyn really wants.

kolbylyn:
That's true that the visualization part is very specific and would go unused by most astroplan users, but I still think an easy add would include outputting phase info somewhere. Just a quickie routine where you can give a planetary system/EB, give a date and then get where it's at in its orbit.

bmorris3:
Closing this in favor of #315.

PR #122, opened 2015-10-09
Plot fixes

bmorris3:
After trying to make some demo plots using `astroplan.plots` I found some quirks to clean up. I wanted to make two subplots in one figure and noticed that the `ax.figure.autofmt_xdate()` call in `astroplan.plots.plot_airmass` is being applied to all axes within the figure, which means if you make two kinds of subplots in the same figure, you can get incorrectly rotated axis labels on other subplots like so 

demo_bad

Also, the time axis label in the airmass plot has a leftover hyphen in it, but doesn't show the date as I think it was supposed to. 

I've corrected these two bugs by writing a custom axis formatter and a new xlabel to go with it. The same plotting script now produces a plot like this: 

demo_good

cc @jberlanga @adrn

### DIFF
--- a/astroplan/plots/time_dependent.py
+++ b/astroplan/plots/time_dependent.py
@@ -113,7 +113,7 @@ def plot_airmass(target, observer, time, ax=None, style_kwargs=None):
 
     # Set labels.
     ax.set_ylabel("Airmass")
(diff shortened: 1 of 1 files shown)


adrn:
Looks fine to me!

Issue #138, opened 2015-11-05
test_sun_separation failing


eteq:
Probably due to the recent GCRS changes in astropy (see #136).  I think I understand the cause and will work on a PR either today or tomorrow to fix it.

bmorris3:
Thanks @eteq!

Issue #134, opened 2015-11-03
constraints cannot be applied to a single target


eteq:
I noticed the following surprising behavior:

```
>>> obs = astroplan.Observer.at_site('Palomar')
>>> targ = astroplan.FixedTarget.from_name('Albireo')
>>> t = Time('J2000')
>>> c = astroplan.AirmassConstraint(2)
>>> c(obs,[targ],t)
array([[ True]], dtype=bool)
>>> c(obs,targ,t)
TypeError: 'FixedTarget' object is not iterable
```

Is this intended?  I would have expected constraints to give a 1D array if either the target or time passed in is a scalar (or a scalar output if _both_ are scalar).  

cc @bmorris3

bmorris3:
The module will force scalar inputs to non-scalars because the logical operators are applied along axes that otherwise wouldn't exist, [see here for example](https://github.com/astropy/astroplan/blob/master/astroplan/constraints.py#L623-L625). 

One solution could be to simply add an extra statement before returning the outputs to scalarize the output arrays if the inputs are scalar.

eteq:
Gotcha, that makes sense - I like the solution of picking the dimension at the end based on the inputs.  I.e.
```
if input_is_scalar:
    return output.ravel()[0]
elif input_had_scalar_time:
    return output[0]
elif input_had_scalar_target:
    return output[:, 0]
else:
    return output
```
is basically what you mean, right?

bmorris3:
Yes that's what I mean. Hopefully I can implement this soon. Reception of astroplan at dotastro has been pretty good, so I hope I can work on these things soon! Maybe I'll see if I can get this particular fix done now...

PR #139, opened 2015-11-06
Better shape handling in constraints [#134]

bmorris3:
This addresses #134, which showed that some shapes of inputs to the `is_observable` and `is_always_observable` functions produced non-intuitive outputs. 

I catch the scalar input case and ensure appropriately scalar outputs now. Would @eteq volunteer to do the quick review? 

The ci tests should fail due to the unrelated issue #138, but the new constraints tests should pass.

### DIFF
--- a/astroplan/constraints.py
+++ b/astroplan/constraints.py
@@ -511,7 +511,8 @@ def is_always_observable(constraints, observer, targets, times=None,
     targets_is_scalar = False
     times_is_scalar = False
 
-    if not hasattr(targets, '__len__'):
+    if (not hasattr(targets, '__len__') or
+        (hasattr(targets, 'isscalar') and targets.isscalar)):
         targets = [targets]
         targets_is_scalar = True
 
@@ -579,7 +580,8 @@ def is_observable(constraints, observer, targets, times=None,
     targets_is_scalar = False
     times_is_scalar = False
 
-    if not hasattr(targets, '__len__'):
+    if (not hasattr(targets, '__len__') or
+        (hasattr(targets, 'isscalar') and targets.isscalar)):
         targets = [targets]
         targets_is_scalar = True
 
--- a/astroplan/constraints.py
+++ b/astroplan/constraints.py
@@ -526,9 +526,9 @@ def is_always_observable(constraints, observer, targets, times=None,
                                       time_range=time_range,
                                       time_grid_resolution=time_grid_resolution)
                            for constraint in constraints]
-    contraint_arr = np.logical_and.reduce(applied_constraints)
+    constraint_arr = np.logical_and.reduce(applied_constraints)
 
-    always_observable = np.all(contraint_arr, axis=1)
+    always_observable = np.all(constraint_arr, axis=1)
 
     if targets_is_scalar and times_is_scalar:
         return always_observable.ravel()[0]
@@ -595,9 +595,9 @@ def is_observable(constraints, observer, targets, times=None,
                                       time_range=time_range,
                                       time_grid_resolution=time_grid_resolution)
                            for constraint in constraints]
-    contraint_arr = np.logical_and.reduce(applied_constraints)
+    constraint_arr = np.logical_and.reduce(applied_constraints)
 
-    ever_observable = np.any(contraint_arr, axis=1)
+    ever_observable = np.any(constraint_arr, axis=1)
 
     if targets_is_scalar and times_is_scalar:
         return ever_observable.ravel()[0]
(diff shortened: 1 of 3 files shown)


astroplan/constraints.py
```diff
@@ -506,15 +506,35 @@ def is_always_observable(constraints, observer, targets, times=None,
...
+    times_is_scalar = False
+
+    if not hasattr(targets, '__len__'):
+        targets = [targets]
+        targets_is_scalar = True
+
     if not hasattr(constraints, '__len__'):
         constraints = [constraints]
 
+    if not hasattr(times, '__len__') or not hasattr(time_range, '__len__'):
```
eteq: The first of these won't work because a scalar `Time` object still has a `__len__` - it's just that if you try to use it you get a `TypeError`.  Instead, the best thing is to do to use ``times.isscalar`` to decide if it's a scalar.  (Or maybe ``(times is not None and times.isscalar)``?  I think ``times`` can be None if you give a ``time_range``, right?)

astroplan/constraints.py
```diff
@@ -506,15 +506,35 @@ def is_always_observable(constraints, observer, targets, times=None,
         List of booleans of same length as ``targets`` for whether or not each
         target is observable in the time range given the constraints.
     """
+    targets_is_scalar = False
+    times_is_scalar = False
+
+    if not hasattr(targets, '__len__'):
```
eteq: As with `Time` below, checking for `__len__` may not work if a `SkyCoord` is given, because it has a `__len__` but it raises an error when used.  (If you're wondering, this behavior is also true of numpy arrays...).  So maybe ``not hasattr(targets, '__len__') or (hasattr(targets, 'isscalar') and targets.isscalar)``?

astroplan/constraints.py
```diff
@@ -506,15 +506,35 @@ def is_always_observable(constraints, observer, targets, times=None,
...
     contraint_arr = np.logical_and.reduce(applied_constraints)
-    return np.all(contraint_arr, axis=1)
+
+    always_observable = np.all(contraint_arr, axis=1)
+
+    if targets_is_scalar and times_is_scalar:
+        return always_observable.ravel()[0]
+
+    elif targets_is_scalar:
+        return always_observable[:, 0]
```
eteq: what about the `times_is_scalar` but  not `targets_is_scalar` case?  Don't you then want to do ``always_observable[0]`` ? Or does that case broadcast to 1-d already?

eteq:
I confess to be a bit confused why the tests seem to pass despite my first comment above... (Although the `SkyCoord` case looks like its not tested, so you might add that for good measure)

Otherwise, looks good!

bmorris3:
Ah that's right, I'm testing lists of `FixedTarget`s here, I'll do the tests with `SkyCoord`s also.

astroplan/constraints.py
bmorris3: Maybe I'm confused, but that doesn't seem to be happening. Here `non_scalar_target_sc` is a length=3 sky-coordinate object:
```python
In [14]: non_scalar_target_sc
Out[14]: 
<SkyCoord (ICRS): (ra, dec) in deg
    [(279.23473479, 38.78368896), (78.63446707, 8.20163837),
     (37.95456067, 89.26410897)]>

In [15]: len(non_scalar_target_sc)
Out[15]: 3

In [16]: non_scalar_target_sc.isscalar
Out[16]: False

In [17]: non_scalar_target_sc.__len__
Out[17]: 
<bound method SkyCoord.__len__ of <SkyCoord (ICRS): (ra, dec) in deg
    [(279.23473479, 38.78368896), (78.63446707, 8.20163837),
     (37.95456067, 89.26410897)]>>

In [18]: non_scalar_target_sc.__len__()
Out[18]: 3
```

astroplan/constraints.py
bmorris3: In that case, the above code outputs: 
```python


In [35]: non_scalar_target_ft
Out[35]: 
[<FixedTarget "Vega" at SkyCoord (ICRS): (ra, dec) in deg (279.23473479, 38.78368896)>,
 <FixedTarget "Rigel" at SkyCoord (ICRS): (ra, dec) in deg (78.63446707, 8.20163837)>,
 <FixedTarget "Polaris" at SkyCoord (ICRS): (ra, dec) in deg (37.95456067, 89.26410897)>]

In [36]: scalar_time
Out[36]: <Time object: scale='utc' format='iso' value=2001-02-03 04:05:06.000>

In [34]: is_observable(AirmassConstraint(max=2), subaru, 
non_scalar_target_ft, times=scalar_time)
Out[34]: array([ True,  True, False], dtype=bool)
```
which looks right to me.

bmorris3:
Hey @eteq - after rebasing this fix, there's a recursion error rearing its ugly head. Any thoughts on how this is happening? It stems from an [`Angle.separation()` call](https://travis-ci.org/astropy/astroplan/jobs/94438200#L694).

bmorris3:
This PR has been superseded by #165.

PR #141, opened 2015-11-09
Fix airmass constraint when no min is present for targets below horizon

eteq:
It turns out that ``AirmassConstrain(2)`` will return True for targets that are below the horizon.  While this is correct in the sense that allowing a negative airmass is meeting the requirement of "no minmum", I think it misses the point of the primary use case of wanting things to be at a higher airmass than x, while still being *up*.  So this changes the default value of  ``min`` to be 0 instead of ``None``, which strikes me as more intuitive for the typical use case.

It also adds a regression test to make sure this stays true.

### DIFF
--- a/astroplan/constraints.py
+++ b/astroplan/constraints.py
@@ -160,30 +160,32 @@ class AirmassConstraint(AltitudeConstraint):
 
     In the current implementation the airmass is approximated by the secant of
     the zenith angle.
+
+    .. note::
+        The ``max`` and ``min`` arguments appear in the order (max, min)
+        in this initializer to support the common case for users who care
+        about the upper limit on the airmass (``max``) and not the lower
+        limit.  For the same reason, if ``max`` is given without ``min``, the
+        default ``min`` is 0, not ``None``
+
+    Parameters
+    ----------
+    max : float or `None`
+        Maximum airmass of the target. `None` indicates no limit.
+
+    min : float or `None`
+        Minimum airmass of the target. `None` indicates no limit.  Note that in
+        the `None` case, this will mean *negative* airmasses (below the horizon)
+        are accepted.
+
+    Examples
+    --------
+    To create a constraint that requires the airmass be "better than 2",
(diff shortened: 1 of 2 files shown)


bmorris3:
Would it be any more rigorous to constrain airmass >= 1?

eteq:
Hmm, that does make sense, @bmorris3, *except* for the corner case of ``secz=1``: right now, if I set the default `min` to `1`, it will cause something at ``secz=1`` to *fail* the constraint (because it's ``secz>min``, not ``secz>=min``).  We could change it so that min/max mean ``<=`` instead of ``<`` and change the min to ``1``, but then I should probably change *all* the inequalities to ``<=`` for consistency.  I guess it does make sense, because generally the words "min" and "max" imply inclusive... Should I do that?

bmorris3:
@eteq – I think so. Inclusive boundaries make sense, and in most practical cases won't make a difference except in the `z=0` corner case, where it makes a difference for the better.

cdeil:
`test_sun_separation` is failing in master and also with this PR:
https://travis-ci.org/astropy/astroplan/jobs/90028565#L662
and in #139 and #140 also.

I didn't follow the discussion and ongoing PRs, but may I suggest we fix the test in master ASAP to get rid of unrelated fails in PRs? If it's not obvious / quick how to fix it, putting a `@pytest.mark.xfail` in front of it is better than leaving master broken for weeks. OK if I do this? Or will the fail be fixed by one of these PRs soon?

eteq:
@cdeil @bsipocz - FYI re: the `test_sun_separation` issue: it's because conda doesn't make new astropys for py3.3 ... #140 should fix that.

bmorris3:
@eteq Do you need to rebase for that fix to take effect in the travis builds?

eteq:
@bmorris3 - actually just a new commit would have worked (travis automatically merges with master before running the tests), but I rebased anyway to make it easier to look at locally.

bmorris3:
Thanks! Merged.

PR #147, opened 2016-02-01
add Very Large Array to database

keflavich:


bmorris3:
Thanks @keflavich! You might also want to add the same coordinate for VLA to the [sites archive in astropy over in astropy-data](https://github.com/astropy/astropy-data/blob/gh-pages/coordinates/sites.json). @eteq ported astroplan's sites machinery to astropy, and astroplan will soon use the astropy version to eliminate the redundancy.

eteq:
@keflavich - for an example of how to do that, see astropy/astropy-data#10

eteq:
I'm going to close this as it should be in astropy-data, but feel free to just port this straight to there, @keflavich

PR #150, opened 2016-03-22
Adding airmass plot options

bmorris3:
Adding two toggle-on features to the `plot_airmass` method: 
1. `brightness_shading` shades the background according to the darkness of the sky (light shading for 0 deg twilight, darker for -18 deg twilight)
2. `altitude_yaxis` adds a new y-axis on the right side of the figure with the equivalent altitudes in degrees. See below for an example plot with both options on:

```python
from astroplan.plots import plot_airmass
import matplotlib.pyplot as plt
from astropy.time import Time
from astroplan import FixedTarget, Observer

time = Time.now()
target = FixedTarget.from_name('HD 189733')
apo = Observer.at_site('APO')
plot_airmass(target, apo, time, brightness_shading=True, altitude_yaxis=True)
plt.show()
```

figure_1

cc @eteq @ejeschke @Barentsen

### DIFF
--- a/astroplan/plots/time_dependent.py
+++ b/astroplan/plots/time_dependent.py
@@ -5,6 +5,7 @@ from __future__ import (absolute_import, division, print_function,
 import numpy as np
 import astropy.units as u
 from astropy.time import Time
+from astropy.coordinates import get_sun
 import warnings
 
 from ..exceptions import PlotWarning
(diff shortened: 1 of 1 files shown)


ejeschke:
Nice!  Is it still pretty easy to change the color of the avxspan region plotted?

ejeschke:
Suggestion: it would be really great to be able to plot two avyspan's--one for a low altitude/airmass limit and one for a high altitude/airmass limit.  These are to show telescope or instrument limitations wrt target positions.

Of course one can augment the plot, but I think this just might be generally useful enough to consider adding kwarg options for it.

Issue #143, opened 2015-11-20
Common feature requests from #dotastro


bmorris3:
Some features that were requested in several conversations at the .Astronomy conference: 
* an `is_observable` alternative that sets bounds on the number of consecutive hours that a target is observable per night before yielding which is observable
* an `is_observable` alternative that returns which months are the best ones to observe a target given a list of constraints (this might go hand in hand with the above requested method)

jberlanga:
@bmorris3  

* It seems to me like the first request assumes the questions "Is my object observable continuously for X hours" whereas the functionality of `is_observable` asks "Is my object observable at any point from time Y to time Z."  If so, could the user not simply set `time_grid_resolution` and `time_range` to something appropriately small, and repeat? (BTW, I'm confused on what [`times`](https://github.com/astropy/astroplan/blob/62583a0d184506d3c3d126f26be325805c66135f/astroplan/constraints.py#L537)'s exact role is--could it potentially be a list of times over which to repeatedly apply `is_observable`?).

* Does the second request wander into scheduling territory?  Or does it make sense to have a convenience function that asks this for only one target?

bmorris3:
* You're right that this could be achieved by looping over calls to `is_observable`, but I'm wondering if there's a more vectorized way we can achieve that computation to speed it up, since this is such a common and natural question to ask. I'll clarify the `times` stuff at the end of my response.

* It's almost a scheduling task, but not quite, it could be achieved quite easily. Since the constraint calls return boolean areas for each time, all we'd really need to do on the outputs is something like this: 
```python
c = AirmassConstraint(max=2)
observable = c(observer, target, times)
set(times[observable].datetime.month)
```
but I haven't thought about the most efficient way to write that function so that we get sufficient precision for the number of hours that each target is up. 

*** 

`times` is an array of times on which each constraint is evaluated. You can enter your own array of times, or have `is_observable` compute one by supplying the `time_range` kwarg with lower and upper bounds on the times you care about.

jberlanga:
@bmorris3 Ah, I see.  I think it would be simpler if users could supply times to one variable (a la `plot_airmass` & friends)--either specify an array of times or provide a scalar (of one time) and let `is_observable` use a default time range of 24 hours to compute an array surrounded on that time.

bmorris3:
You can do both of those operations with `is_observable` as it is written currently like this:
```python
is_observable(constraints, observer, targets, times=array_of_times)
is_observable(constraints, observer, targets, 
[...]

PR #152, opened 2016-03-24
Months observable [#143]

bmorris3:
This PR addresses a feature request in #143 from the dotastro conference for a convenience method that returns the months in which targets are observable given constraints. I've made a slow but thorough implementation of this method.  

It computes whether or not the target is observable every half hour of every day for a year, then identifies the months during times that are observable. Ideally this would be the kind of calculation that uses the trigonometric target altitude function rather than our iterative/interpolative method that propagates through `SkyCoord`->`AltAz`, adding extra computation time. But I'll leave that for a future PR.

Here's a demo script showing the functionality: 

```python
from astroplan import (FixedTarget, Observer, months_observable, 
                       AltitudeConstraint, AtNightConstraint)
import astropy.units as u

keck = Observer.at_site('Keck')
constraints = [AltitudeConstraint(min=70*u.deg), 
               AtNightConstraint.twilight_astronomical()]
targets = [FixedTarget.from_name('Vega'), FixedTarget.from_name('Regulus')]

months_observable(constraints, keck, targets)
# Returns: [{4, 5, 6, 7, 8, 9}, {1, 2, 3, 4, 5, 11, 12}]
# (Vega is high in summer, Regulus is high in winter)
```

### DIFF
--- a/astroplan/constraints.py
+++ b/astroplan/constraints.py
@@ -570,7 +570,7 @@ def is_observable(constraints, observer, targets, times=None,
 def months_observable(constraints, observer, targets,
                       time_grid_resolution=0.5*u.hour):
     """
-    During which months is the target observable?
+    During which months are the targets observable?
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


bmorris3:
@eteq – would you restart the AppVeyor run, I don't think I have permission? It looks like the anaconda install failed.

PR #170, opened 2016-06-13
clarifying tutorial

kvyh:
Referencing the previous section of the tutorial rather than the code block above it.


bmorris3:
Thanks!

PR #171, opened 2016-06-13
Constraint docstring tweaks

eteq:
While doing a post-hoc review of #152 (sorry about not doing that sooner, @bmorris3), I noticed that a number of the constraint docstrings were in "question-form".  That's not PEP8 compliant, but more importantly, I think it's a bit confusing, because it's not clear that the goal of the function is to *answer* that question.  So this just re-words those docstrings to be descriptive rather than "question-form".

This also removes a "to do" section that I think @bmorris3 added in #152 - my justification for that is that in my experience "todo's" can live a long time, so it's better to leave them as hints for those perusing the code rather than the documentation... But if there was a particular reason you really wanted that one in there, I can certainly remove that commit.

### DIFF
--- a/astroplan/constraints.py
+++ b/astroplan/constraints.py
@@ -604,11 +604,9 @@ def months_observable(constraints, observer, targets,
         observable, one set per target. These integers are 1-based so that
         January maps to 1, February maps to 2, etc.
 
-    To Do
-    -----
-    TODO: This method could be sped up a lot by dropping to the trigonometric
-    altitude calculations.
     """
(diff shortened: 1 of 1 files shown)


bmorris3:
Looks good. 

I'm happy with the TODO move. I am still thinking about whether or not it would be simple to add a switch somewhere in the constraints API to use the trigonometric altitude calculations. I could imagine a user computing observability with trigonometric altitude approximations first for a few iterations, then double-checking their schedule with the "full" or "precise" constraints.

PR #174, opened 2016-06-14
Dark Grey or Bright time constraint

kvyh:
Adding a constraint that allows a user to define dark/grey/bright time using how high above the horizon and how illuminated the moon is.

### DIFF
--- a/astroplan/constraints.py
+++ b/astroplan/constraints.py
@@ -402,7 +402,7 @@ class DarkGreyBrightConstraint(Constraint):
     """
     # TODO: make a new test
     @u.quantity_input(horizon=u.deg)
-    def __init__(self lunar_horizon=90*u.deg, min_moon_illumination=0,
+    def __init__(self, lunar_horizon=90*u.deg, min_moon_illumination=0,
                                                 max_moon_illumination=1):
         '''
         Parameters
@@ -425,8 +425,8 @@ class DarkGreyBrightConstraint(Constraint):
         >>> greytime = DarkGreyBrightConstraint(max_moon_illumination=.5)
         '''
         
-        self.min_illumination = min_max_moon_illumination[0]
-        self.max_illumination = min_max_moon_illumination[1]
+        self.min_illumination = min_moon_illumination
+        self.max_illumination = max_moon_illumination
         self.horizon = lunar_horizon
         
     def _get_lunar_altitudes(self, times, observer, targets):
@@ -436,15 +436,15 @@ class DarkGreyBrightConstraint(Constraint):
         aakey = (tuple(times.jd), 'sun')
 
         if aakey not in observer._altaz_cache:
-            try:
-                # Broadcast the lunar altitudes for the number of targets:
-                altaz = observer.altaz(times, get_moon(times))
-                altitude = altaz.alt
-                altitude.resize(1, len(altitude))
-                altitude = altitude + np.zeros((len(targets), 1))
+            # TODO: make sure this works properly
+            # Broadcast the lunar altitudes for the number of targets:
+            altaz = observer.altaz(times, get_moon(times))
+            altitude = altaz.alt
+            altitude.resize(1, len(altitude))
+            altitude = altitude + np.zeros((len(targets), 1))
 
-                observer._altaz_cache[aakey] = dict(times=times,
-                                                    altitude=altitude)
+            observer._altaz_cache[aakey] = dict(times=times,
+                                                altitude=altitude)
 
         return observer._altaz_cache[aakey]
         
--- a/astroplan/constraints.py
+++ b/astroplan/constraints.py
@@ -395,58 +395,78 @@ class MoonIlluminationConstraint(Constraint):
                              "MoonSeparationConstraint.")
         return mask
 
-class TimeBrightnessConstraint(Constraint):
+    
+class DarkGreyBrightConstraint(Constraint):
     """
     Constrain whether it is dark, gray or bright time or how bright.
     """
-    #following the format of Erik and Stephanie
-    # TODO: make a better test
-    def __init__(self, time='dark', max=None, boolean_constraint=True):
+    # TODO: make a new test
+    @u.quantity_input(horizon=u.deg)
+    def __init__(self lunar_horizon=90*u.deg, min_moon_illumination=0,
+                                                max_moon_illumination=1):
         '''
         Parameters
         ----------
-        time : string (optional)
-            'dark' for dark time, 'grey' for grey time, 'bright' for bright
-            time.p.array()))
-        max : float or 'None' (optional)
-            
-        boolean_constraint : bool
-            
+        Lunar_horizon : `~astropy.units.Quantity` (optional)
+            The altitude of the ``horizon`` and defaults to 90 degrees       
+        min_moon_illumination : float or `None` (optional)
+            the minimum value of the fraction of the moon illuminated
+        max_moon_illumination : float or `None` (optional)
+            the maximum value of the fraction of the moon illuminated
         
+        Example
+        -------
+        Constrain the observations to times where the moon :
+
+        >>> from astroplan import Observer
+        >>> from astroplan.constraints import DarkGreyBrightConstraint
+        >>> subaru = Observer.at_site("Subaru")
+        >>> darktime = DarkGreyBrightConstraint(lunar_horizon=-5*u.deg) 
+        >>> greytime = DarkGreyBrightConstraint(max_moon_illumination=.5)
         '''
-        # TODO: let grey/bright be customizable
-        if time.lower() == 'dark':
-            self.max = 0
-        elif max:
-            self.max = max
-        elif time.lower() == 'grey':
-            self.max = 0.5
-        elif time.lower() == 'bright':
-            self.max = 1
-        else:
-            self.max = 1
-        self.boolean_constraint = boolean_constraint
+        
+        self.min_illumination = min_max_moon_illumination[0]
+        self.max_illumination = min_max_moon_illumination[1]
+        self.horizon = lunar_horizon
+        
+    def _get_lunar_altitudes(self, times, observer, targets):
+        if not hasattr(observer, '_altaz_cache'):
+            observer._altaz_cache = {}
+
+        aakey = (tuple(times.jd), 'sun')
+
+        if aakey not in observer._altaz_cache:
+            try:
+                # Broadcast the lunar altitudes for the number of targets:
+                altaz = observer.altaz(times, get_moon(times))
+                altitude = altaz.alt
+                altitude.resize(1, len(altitude))
+                altitude = altitude + np.zeros((len(targets), 1))
+
+                observer._altaz_cache[aakey] = dict(times=times,
+                                                    altitude=altitude)
+
+        return observer._altaz_cache[aakey]
+        
         
     def compute_constraint(self,times,observer,targets):
+        #find when it is up
+        moon_altaz = self._get_lunar_altitudes(times, observer, targets)
+        moon_altitude = moon_altaz['altitude']
+        moon_down_mask = moon_altitude <= self.horizon
+        #find when it satisfies illumination constraints
         moon = get_moon(times, observer.location, observer.pressure)
-        targets = [target.coord if hasattr(target, 'coord') else target
-                   for target in targets]
-
-        moon_separation = Angle([angular_separation(moon.spherical.lon,
-                                                    moon.spherical.lat,
-                                                    target.spherical.lon,
-                                                    target.spherical.lat)
-                                 for target in targets])
-        #moon_separation is [[time1,time2,time3...][][]] with a bin for each target
-        # TODO: refine brightness computation, it's a lot more complex than below
-        brightness = np.multiply(np.multiply(np.array([moon_illumination(time,observer.location) for time in times]), np.array([observer.is_moon_up(time) for time in times])),1-moon_separation.radian/180.)
-        if self.boolean_constraint:
-            mask = brightness <= self.max
-            return mask
-        #non-boolean will not work until merged with Stephanie's code
+        illumination = moon_illumination(times, observer.location)
+        if self.min is None and self.max is not None:
+            mask = self.max >= illumination
+        elif self.max is None and self.min is not None:
+            mask = self.min <= illumination
+        elif self.min is not None and self.max is not None:
+            mask = ((self.min <= illumination) &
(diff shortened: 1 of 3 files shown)


astroplan/tests/test_constraints.py
```diff
@@ -206,7 +206,12 @@ def test_moon_illumination():
     is_constraint_met = [constraint(lco, None, times=time) for time in times]
     assert np.all(is_constraint_met == [False, True, True])
 
-
+    
+def test_Dark_Grey_Bright():
+    # TODO: write the test
+    pass
```
bmorris3: Time to fill in this test. 

Don't forget to run the tests locally before making a pull request by running this from the top-level of the repository: 
```
python setup.py test
```
Run it with both py 2.7 and 3.5, which Anaconda makes [possible without infinite effort](http://stackoverflow.com/a/24415581/1340208).

astroplan/constraints.py
```diff
@@ -394,7 +395,79 @@ def compute_constraint(self, times, observer, targets):
                              "MoonSeparationConstraint.")
         return mask
 
+    
+class DarkGreyBrightConstraint(Constraint):
```
bmorris3: What's the best name for this constraint. Opinions @ejeschke, @eteq, @adrn?

Some alternative suggestions: 
* `MoonBrightnessConstraint` (which is ambiguous given that we already have a `MoonIlluminationConstraint)
* `MoonContaminationConstraint` (which is too long to type)
* `DarkTimeConstraint` (which is shorter to type but more vague)

astroplan/constraints.py
```diff
@@ -394,7 +395,79 @@ def compute_constraint(self, times, observer, targets):
...
+            # TODO: make sure this works properly
+            # Broadcast the lunar altitudes for the number of targets:
+            altaz = observer.altaz(times, get_moon(times))
+            altitude = altaz.alt
+            altitude.resize(1, len(altitude))
+            altitude = altitude + np.zeros((len(targets), 1))
 
+            observer._altaz_cache[aakey] = dict(times=times,
+                                                altitude=altitude)
+
```
kvyh: I'm using these lines from AtNightConstraint which has them inside a "try:". Do I need to worry about what it is doing?

astroplan/constraints.py
```diff
@@ -394,7 +395,79 @@ def compute_constraint(self, times, observer, targets):
...
+        aakey = (tuple(times.jd), 'sun')
+
+        if aakey not in observer._altaz_cache:
+            # TODO: make sure this works properly
+            # Broadcast the lunar altitudes for the number of targets:
+            altaz = observer.altaz(times, get_moon(times))
+            altitude = altaz.alt
+            altitude.resize(1, len(altitude))
+            altitude = altitude + np.zeros((len(targets), 1))
```
kvyh: This line Errors if there are no targets, is there a way to fix that?

bmorris3: There should always be targets.

astroplan/constraints.py
```diff
@@ -402,8 +402,8 @@ class DarkGreyBrightConstraint(Constraint):
     """
     # TODO: make a new test
     @u.quantity_input(horizon=u.deg)
-    def __init__(self, lunar_horizon=90*u.deg, min_moon_illumination=0,
-                                                max_moon_illumination=1):
+    def __init__(self, lunar_horizon=90*u.deg, min_moon_illumination=None,
```
bmorris3: "Lunar horizon" to me sounds like a quantity for the observer on the moon. What about "maximum lunar altitude"?

kvyh:
The function _get_moon_data defined in #165 will significantly simplify the code for this constraint.

StuartLittlefair:
@kvyh This is very similar to the changes I made to the ```MoonIlluminationConstraint``` in #165. I don't think we should have both a ```MoonIlluminationConstraint``` and this ```DarkGreyBright``` constraint as well since they duplicate functionality.

I'm happy for the constraint you've written to become our Moon Brightness constraint and especially like the lunar altitude is configurable. I think your constraint should *replace* the ```MoonIlluminationConstraint``` currently defined.

astroplan/constraints.py
```diff
@@ -394,7 +395,81 @@ def compute_constraint(self, times, observer, targets):
...
+    """
+    # TODO: make a new test
+    @u.quantity_input(horizon=u.deg)
+    def __init__(self, lunar_horizon=90*u.deg, min_moon_illumination=None,
+                                                max_moon_illumination=None):
+        '''
+        Parameters
+        ----------
+        Lunar_horizon : `~astropy.units.Quantity` (optional)
+            The altitude of the ``horizon`` and defaults to 90 degrees
```
StuartLittlefair: I think this should default to 0 degrees. The way this ```Constraint``` should work is that it is satisfied if the moon illumination fractions are satisfied *or* if the moon is below the specified horizon. 

This is because this constraint will most often be used to ensure that the night sky is not too bright because of the Moon, and the night sky brightness drops dramatically once the Moon has set (see [here](https://confluence.lsstcorp.org/display/SIM/Sky+Brightness+Observations)).

The docstring should make this behaviour clear.

astroplan/constraints.py
```diff
@@ -394,7 +395,81 @@ def compute_constraint(self, times, observer, targets):
...
+            the maximum value of the fraction of the moon illuminated
+        
+        Example
+        -------
+        Constrain the observations to times where the moon :
+
+        >>> from astroplan import Observer
+        >>> from astroplan.constraints import DarkGreyBrightConstraint
+        >>> subaru = Observer.at_site("Subaru")
+        >>> darktime = DarkGreyBrightConstraint(lunar_horizon=-5*u.deg)
```
StuartLittlefair: "Dark time" is commonly defined as when the lunar illumination is <0.25.
"Grey time" is lunar illuminations between 0.25 and 0.65,
"Bright time" is >0.65 

See [here](http://www.ing.iac.es/astronomy/observing/manuals/ps/tech_notes/tn127.pdf) for more details. The docstring examples should reflect this I think.

astroplan/constraints.py
```diff
@@ -394,7 +395,81 @@ def compute_constraint(self, times, observer, targets):
...
+        >>> subaru = Observer.at_site("Subaru")
+        >>> darktime = DarkGreyBrightConstraint(lunar_horizon=-5*u.deg) 
+        >>> greytime = DarkGreyBrightConstraint(max_moon_illumination=.5)
+        '''
+        
+        self.min = min_moon_illumination
+        self.max = max_moon_illumination
+        self.horizon = lunar_horizon
+        
+    def _get_lunar_altitudes(self, times, observer, targets):
```
StuartLittlefair: It makes more sense to me to cache both the lunar altitudes and the lunar illumination here.

astroplan/constraints.py
```diff
@@ -394,7 +395,81 @@ def compute_constraint(self, times, observer, targets):
...
+            observer._altaz_cache = {}
+
+        aakey = (tuple(times.jd), 'sun')
+
+        if aakey not in observer._altaz_cache:
+            # TODO: make sure this works properly
+            # Broadcast the lunar altitudes for the number of targets:
+            altaz = observer.altaz(times, get_moon(times, observer.location))
+            altitude = altaz.alt
+            altitude.resize(1, len(altitude))
```
StuartLittlefair: I see that you've based this code on what is currently in the ```AtNightConstraint```. There are problems with that code (fixed in #165) because you're using the tuple of ```times``` as a key, but the shape of the object in the cache depends on both ```targets``` and ```times```.

Instead, I'd base this function on the ```_get_moon_data``` code in #165, and look at the ```MoonIlluminationConstraint``` in the same PR to see how to reshape the cached altitudes and illuminations to get the correct shape later on.

StuartLittlefair:
See above for some detailed comments on the codebase. My opinion is that the ```MoonIlluminationConstraint``` as implemented in #165 should be the starting point for this PR but that it would be really great if that could be enhanced to allow the user to set the lunar horizon, as you have done here.

Another thought is that classmethods to easily define ```MoonIlluminationConstraint```s for dark, grey and bright time would be very useful, along the same lines as the ones in ```AtNightConstraint``` allow one to easily create constraints for various twilights.

astroplan/constraints.py
kvyh: The reason that I defaulted it to 90 degrees was the assumption that it would be used to define what times were considered dark, grey or bright. For dark time you care about the horizon, for grey and bright you care about moon illumination, but assume the moon is up. My intended functionality would actually be better served by including a minimum lunar altitude that is modifiable and overwritten if a maximum lunar altitude is defined.

StuartLittlefair: The only issue with this is that most astronomers have a different conception of what dark, grey and bright mean. Dark time is moon illuminations < 0.25, even if the moon is up.

And then, since this is always going to be used as a constraint on the brightness of the night sky, it doesn't make sense to assume the moon is up, since the night sky can be darker when the Moon is 65% illuminated, but below the horizon, than it is when the moon is up but 30% illuminated.

wtgee: Just to chime in, I can easily foresee a use case for light-pollution
levels. Not sure if this is the appropriate place for that but it seems to
be in line with this Constraint without involving the moon at all.

On Wed, Jun 15, 2016 at 9:20 AM, StuartLittlefair <notifications@github.com>
wrote:

> In astroplan/constraints.py
> <https://github.com/astropy/astroplan/pull/174#discussion_r67182909>:
>
> > @@ -394,7 +395,81 @@ def compute_constraint(self, times, observer, targets):
> >                               "MoonSeparationConstraint.")
> >          return mask
> >
> > +
> > +class DarkGreyBrightConstraint(Constraint):
> > +    """
> > +    Constrain whether it is dark, gray or bright time or how bright.
> > +    """
> > +    # TODO: make a new test
> > +    @u.quantity_input(horizon=u.deg)
> > +    def __init__(self, lunar_horizon=90*u.deg, min_moon_illumination=None,
> > +                                                max_moon_illumination=None):
> > +        '''
> > +        Parameters
> > +        ----------
> > +        Lunar_horizon : `~astropy.units.Quantity` (optional)
> > +            The altitude of the ``horizon`` and defaults to 90 degrees
>
> The only issue with this is that most astronomers have a different
> conception of what dark, grey and bright mean. Dark time is moon
> illuminations < 0.25, even if the moon is up.
>
> And then, since this is always going to be used as a constraint on the
> brightness of the night sky, it doesn't make sense to assume the moon is
> up, since the night sky can be darker when the Moon is 65% illuminated, but
> below the horizon, than it is when the moon is up but 30% illuminated.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/astropy/astroplan/pull/174/files/8a8fcaebad00584066c950ea2c8161ce21d35d7e#r67182909>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAEUUM4YWUnV3x9I1o3iLZgvnvHzGKCaks5qMBhYgaJpZM4I1rL5>
> .
>



-- 
~Wilfred Tyler Gee

kvyh:
Thanks @StuartLittlefair, I'll try building it on top of your constraint_broadcasting with your comments in mind. The classmethods make a lot of sense to implement.

bmorris3:
@StuartLittlefair: 
> This is very similar to the changes I made to the MoonIlluminationConstraint in #165.

As the code stands now, I agree that this has a lot of overlap with `MoonIlluminationConstraint`, but it may become a bit more specific. @kvyh is this summer's Google Summer of Code student working on astroplan, and this was a task that he started before your PR came around. 

I think an additional feature that would make `DarkGreyBrightConstraint` unique and useful is to set a tolerance on the amount of time spent in violation of the specified brightness constraint. For example, at Apache Point Observatory, an observing block scheduled in "dark time" only has to be "dark" for >~80% of the time, and often in practice has a setting bright moon at the beginning of the run, or a rising bright moon at the end of the run. Maybe an additional keyword argument could be added like `bright_duration_tolerance=1*u.hour`, which allows the min/max illumination constraint to be violated for some amount of time.

StuartLittlefair:
@bmorris3 something like that would be useful, but might be confusing to implement with current ```Constraint``` API? ```Constraints``` are designed to be evaluated at instantaneous times, whereas you are suggesting a constraint whose validity depends on conditions over a time range.

I think it would make sense to have a new class for constraints that are valid if they're satisfied for a certain fraction of a time range.

bmorris3:
@StuartLittlefair That's something that I keep bumping into. Do you think it would be sufficient to have a time-resolution keyword somewhere that ensures that the constraints are always computed on a dense-enough grid of times to ensure that duration constraints are measured properly? 

In the `bright_duration_tolerance=1*u.hour` example, there could be a mechanism that interpolations between the input times so that the constraint gets computed with gaps no larger than `0.1 * bright_duration_tolerance`.

StuartLittlefair:
@bmorris3 I'm not sure I understand your point?

The current ```Constraints``` only calculate if they are valid at the exact ```times``` passed to ```compute_constraint```, no?

The only place the accuracy of gridding matters is in ```is_always_observable``` and ```is_observable```, is that right?

In fact, thinking about this makes me think that the best way to handle the case you describe above for Apache Point is to keep the Constraints as they are, but to define a new ```is_observable_fraction``` function, which takes the same arguments as is_observable and additionally the fraction of the time span for which the constraints must be valid. 

In that function you have to be a bit careful about the resolution of your time grid, but as you suggest, that could be worked out from the list of constraints passed to the function.

bmorris3:
That's right @StuartLittlefair, and that's more or less what I'm thinking. You could imagine that this would be useful for other constraints also, like airmass, i.e. "I want to take an hour-long exposure, and it should spend >50% of that visit at an airmass <2".

astroplan/constraints.py
ejeschke: Agree with this definition. This is what we use at Subaru.

bmorris3: To demonstrate that these definitions are not universal, there are different defaults at APO: 

> Bright time is more than 50% moon
> Grey time is less than 50% moon
> Dark time is no moon
> 
> In all cases, a half night is still called bright, grey or dark even if the moon condition is not met for up to an hour.

StuartLittlefair: Thanks Brett, 

This is good to know. I guess that the significant differences in definitions make the classmethods not very useful. 

It does makes the enhancement that @kvyh is working on all the more useful though!

kvyh:
This is being superceded by #201.

PR #182, opened 2016-06-21
adding a schedule class and modifying code

kvyh:
I added Schedule and Slot classes and rewrote portions of the PriorityScheduler to make use of them. I also modified Stephanie's example code to call the output correctly using schedule attributes. The new version of PriorityScheduler takes the exact same inputs as the old code, and returns essentially the same output.


kvyh:
I'm still doing ongoing work, hence the many TODOs, but I'm interested in suggestions or recommendations on the structure and what methods and attributes a schedule needs.

kvyh:
Also, if this is extended to cover the SequentialScheduler, this PR would overwrite most of #181 because it uses a different scheduler metaclass.

bmorris3:
> this PR would overwrite most of #181

@kvyh If so, would you like to close that PR? If you do close that PR, mention that it was superseded by this PR.

kvyh:
I will do that, and try to make it work for SequentialScheduler so that we can remove the old metaclass that had the issues.

astroplan/scheduling.py
```diff
@@ -299,14 +463,54 @@ def _make_schedule(self, blocks):
         return new_blocks, True
 
 
-class PriorityScheduler(Scheduler):
+class ScheduleScheduler(object):
```
bmorris3: Why is there a `Scheduler` and a `ScheduleScheduler`? Is there a more obvious name for `ScheduleScheduler`?

astroplan/scheduling.py
```diff
@@ -124,8 +126,170 @@ def components(self, val):
...
+                try:
+                    print(block.target.name, 'starting @',
+                          block.start_time.iso, 'lasting', block.duration)
+                except:
+                    print(block.name, 'starting @', block.start_time.iso,
+                          'lasting', block.duration)
+            else:
+                print('trans starting @', block.start_time.iso, 'lasting',
+                      block.duration, ':', block.components)
+        return 'done'
```
bmorris3: A `__repr__` should return the string that you want to print, not `print` the strings.

astroplan/scheduling.py
```diff
@@ -124,8 +126,170 @@ def components(self, val):
...
+    """
+    
+    def __init__(self, start_time, end_time):
+        self.start = start_time
+        self.end = end_time
+        self.duration = end_time-start_time
+        self.occupied = False
+        self.middle = False
+        self.OB = False
+        self.TB = False
```
bmorris3: Shouldn't a slot contain the block instance? For example, a slot filled with an OB will contain the `ObservingBlock` object, on an attribute called `self.block` or something.

Then you won't need the `Slot.OB` or `Slot.TB` attributes, you'd be able to check if `isinstance(self.block, ObservingBlock)`, for instance.

astroplan/scheduling.py
```diff
@@ -350,22 +559,21 @@ def from_timespan(cls, center_time, duration, **kwargs):
...
         _all_times = []
         _block_priorities = np.zeros(len(blocks))
-        for i,b in enumerate(blocks):
+        for i, b in enumerate(blocks):
             if b.constraints is None:
                 b._all_constraints = self.constraints
             else:
                 b._all_constraints = self.constraints + b.constraints
-            b._duration_offsets = u.Quantity([0*u.second, b.duration/2, b.duration])
+            b._duration_offsets = u.Quantity([0 * u.second, b.duration / 2, b.duration])
```
bmorris3: I think we shouldn't hard-code the times at which the constraints are computed – maybe some users will want to compute at higher time-resolution.

You could add a `time_resolution` argument+attribute, which controls how finely to compute the constraints. We have methods for making that time grid easy to parameterize, see [`astroplan.time_grid_from_range`](https://astroplan.readthedocs.io/en/latest/api/astroplan.time_grid_from_range.html?highlight=time%20grid#astroplan.time_grid_from_range).

astroplan/scheduling.py
kvyh: ScheduleScheduler is currently a "duplicate" of the Scheduler so that SequentialScheduler can run since it hasn't yet been changed to use the Schedule. Once that is written, they will be merged.

astroplan/scheduling.py
kvyh: Yes, is there a good way of writing a `__repr__` that returns a string that would print on multiple lines?

astroplan/scheduling.py
kvyh: Yes, that was a relic of my not using `isinstance`. I'll change that, it will also clean up the `schedule.scheduled_blocks` code.

astroplan/scheduling.py
bmorris3: Before we get too into this, let's note here that the `__repr__` is not of paramount importance yet. 

This could be a touchy subject, but I think they should really be single line unless there's a really good reason that they shouldn't be. The point of a `__repr__` isn't to display all of the information within the object, it's to summarize it. Maybe you could make the `__repr__` simply print the object name (like most objects do) and the number of scheduled OB's?

astroplan/scheduling.py
```diff
@@ -124,8 +126,170 @@ def components(self, val):
...
+    def __init__(self, start_time, end_time, constraints=None):
+        self.slots = [Slot(start_time, end_time)]
+        self.constraints = constraints
+        self.slew_duration = 4*u.min
+        # TODO: replace/overwrite slew_duration with Transitioner calls
 
+    def __repr__(self):
+        # once TransitionBlocks are defined, this should be scheduled_blocks
+        # TODO: make a better repr
+        for block in self.observing_blocks:
```
bmorris3: It looks like you're accessing an attribute `self.observing_blocks` that wasn't defined in the `__init__` method. Always initialize attributes in the `__init__` method, even if you have to initialize them with `None` or an empty list.

kvyh: I'm accessing a property  assigned below that using the same format as the `end_time` property of `TransitionBlock` defined on line 114.

kvyh:
I have now completely transferred over the current scheduling classes over to using a `Schedule`. There may be potential for more to be done using the `Schedule` during the scheduling process rather than as an output, but it is not included at this time.

astroplan/scheduling.py
```diff
@@ -163,6 +163,7 @@ def __init__(self, start_time, end_time, constraints=None):
 
     def __repr__(self):
         # once TransitionBlocks are defined, this should be scheduled_blocks
+        # TODO: make a better repr
         for block in self.observing_blocks:
             if hasattr(block, 'target'):
                 try:
```
bmorris3: Try your best to use meaningful commit messages.

astroplan/scheduling.py
StuartLittlefair: I wonder if this should be a property of the ```Constraint``` itself - how it is to be evaluated over an ```ObservingBlock```'s duration? 

I raise this suggestion partly as a result of the issues I've noticed with the ```MeridianTransitConstraint``` in #165. The way that's written at the moment, a user would only care if that constraint was valid at the start of the ```ObservingBlock```.

It may be possible to re-write that particular constraint, but perhaps there is a case that some constraints need to be evaluated on a fine time grid, others on a coarse time grid, and others just at the start of the ```ObservingBlock```?

astroplan/scheduling.py
```diff
@@ -124,8 +136,159 @@ def components(self, val):
...
+        self.constraints = constraints
+        self.slew_duration = 4*u.min
+        # TODO: replace/overwrite slew_duration with Transitioner calls
+
+    def __repr__(self):
+        return 'Schedule containing ' + str(len(self.observing_blocks)) + \
+               ' observing blocks between ' + str(self.slots[0].start.iso) + \
+               ' and ' + str(self.slots[-1].end.iso)
+
+    def apply_constraints(self):
```
StuartLittlefair: @kvyh - can you explain what the rationale is behind having ```Constraints``` in the ```Schedule``` is, and why it needs an ```apply_constraints``` method?

I wondered if the ```Schedule``` should just be a relatively  dumb bag of ```Slots```, and the job of applying ```Constraints``` and deciding where to place ```Slots``` should be delegated to the ```Scheduler```? Perhaps  I've missed some subtlety?

astroplan/scheduling.py
```diff
@@ -277,36 +439,35 @@ def _make_schedule(self, blocks):
...
                 # if even the best is unobservable, we need a gap
-                new_blocks.append(TransitionBlock({'nothing_observable': self.gap_time},
-                                                  current_time))
                 current_time += self.gap_time
             else:
                 # If there's a best one that's observable, first get its transition
                 trans = block_transitions.pop(bestblock_idx)
                 if trans is not None:
-                    new_blocks.append(trans)
+                    self.schedule.insert_slot(trans.start_time, trans)
```
StuartLittlefair: These will need to be inside ```try...except``` blocks, and you'll have to remember to remove the transition block if the main observing block cannot be inserted.

In fact, I wonder if this should be a loop over the blocks in order of priority, which inserts the first block which is insertable at the ```current_time```?

astroplan/scheduling.py
```diff
@@ -277,36 +439,35 @@ def _make_schedule(self, blocks):
...
 class PriorityScheduler(Scheduler):
     """
     A scheduler that optimizes a prioritized list.  That is, it
     finds the best time for each ObservingBlock, in order of priority.
     """
+
     @u.quantity_input(gap_time=u.second)
     def __init__(self, start_time, end_time, constraints, observer,
-                 transitioner=None, gap_time=30*u.min, slew_time=5*u.min):
+                 transitioner=None, gap_time=30 * u.min, slew_time=5 * u.min):
```
StuartLittlefair: I realise this is a hangup from someone else's code, but is the ```slew_time``` a stop-gap until we have working ```Transitioners```? I don't think we need both. If we do need both for some reason, it's probably  more useful to supply the slew *rate* and work out the slew time once the from-to coordinates are known.

astroplan/scheduling.py
kvyh: I could image extra keywords on `Constraints` indicating how fine of time spacing they need to be evaluated at, which could be handled by the schedulers. The issue is that it would not fit with the way that `Constraint`s are currently defined: to be evaluated at a single point, and that it would require more complex constraint evaluation by schedulers. If I can create an efficient way of allowing a `Slot` to calculate `Constraint` in itself, I could create exceptions for certain `Constraint`s so that they only needed evaluation at certain points.

astroplan/scheduling.py
kvyh: The having `Constraints` in the `Schedule` was to allow it to remove sections from the Schedule if, for example, they didn't fulfil the `AtNightConstraint` that was required, so that a `Scheduler` wouldn't need to handle those portions of the `Schedule`. I think that it can be removed since smart use and evaluation of `strided_arrays` within the `Scheduler` should be able to handle "global" `Constraints` more efficiently than changing the `Schedule`.

StuartLittlefair: I think that would be good, if possible. From a conceptual point of view, I think changing a ```Schedule``` should be the job of the ```Scheduler``` and not the ```Schedule``` itself. Otherwise the two overlap in functionality in a potentially confusing way IMO.

astroplan/scheduling.py
kvyh: True, I think that the main reason that the `slew_time` is defined here is that the `PriorityScheduler` as it is currently written can't handle having a `Transitioner` with the way that it schedules. If a much smaller time resolution were used for the `strided_array` then you could run a `Transitioner` between scheduled `ObservingBlocks` and round its duration up to a multiple of the time resolution to preserve the array.

astroplan/scheduling.py
StuartLittlefair: I've been thinking about this over the weekend and I'm more convinced that an extra keyword on ```Constraints``` is the way to do this. To illustrate my thinking, I'll set out my case below, and then provide some code which sets out how I think it might work. Feel free to ignore this comment if you disagree with the approach.

Firstly, the time spacing on which a ```Constraint``` needs to be calculated is intrinsically a property of the ```Constraint``` itself. Some ```Constraints``` change rapidly and need to be evaluated on a fine grid; others change slowly and can be evaluated on a coarse grid, or just at the start of the ```ObservingBlock```. So a ```time_resolution``` keyword should be a property of the ```Constraint```. 

Second, I'm not sure that this conflicts with the way ```Constraint```s are defined. A call to a constraint is still evaluated at a single point, so the current code and concept remain unchanged. 

All that changes is how the ```Scheduler```s evaluate the constraints over an ```ObservingBlock```, and I don't think this has to be more complex. This is probably best illustrated by some example code for the SequentialScheduler:

```python
constraint_res = []
for constraint in b._all_constraints:
    times = current_time + transition_time + constraint.grid(b.duration)
    constraint_res.append(constraint(self.observer, [b.target], times)
# things are a bit trickier here, because constraint_res is not square array
# i.e rows are different shapes
block_constraint_results.append(
    np.prod([np.prod(res) for res in constraint_res])
)
```

The ```constraint.grid``` call would create an array of ```TimeDelta```s with the appropriate spacing if ```b.duration``` was greater than ```constraint.time_resolution``` or zero if not (or if ```constraint.time_resolution``` was negative) 

This would probably make things more efficient; for example and ```AirmassConstraint``` would be evaluate a few times over a block, whereas we only really need to evaluate a Lunar constraint at the start time, for example. It would also fix my issues with the ```MeridianTransitConstraint```, which would get a negative ```time_resolution``` keyword.

kvyh:
I made a number of changes that didn't merge properly with this branch, so I'm closing this and will open a new PR.

PR #202, opened 2016-08-02
Require moon to be up when there is a minimum moon illumination 

kvyh:
I made `MoonIlluminationConstraint` check whether the moon is up (`and`) when a minimum moon illumination is given, and when there is no minimum if the moon is not up (`or`). I also added a new time to the `test_moon_illumination_constraint` where the moon is below the horizon.

### DIFF
--- a/astroplan/constraints.py
+++ b/astroplan/constraints.py
@@ -550,21 +550,24 @@ class MoonIlluminationConstraint(Constraint):
         # first is the moon up?
         cached_moon = _get_moon_data(times, observer)
         moon_alt = cached_moon['altaz'].alt
-        moon_alt_mask = moon_alt < 0
+        moon_down_mask = moon_alt < 0
+        moon_up_mask = moon_alt >=0
 
         illumination = cached_moon['illum']
         if self.min is None and self.max is not None:
             mask = self.max >= illumination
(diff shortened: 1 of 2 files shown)


astroplan/constraints.py
```diff
@@ -550,21 +550,24 @@ def compute_constraint(self, times, observer, targets):
...
         if self.min is None and self.max is not None:
             mask = self.max >= illumination
+            mask = np.logical_or(moon_down_mask, mask)
         elif self.max is None and self.min is not None:
             mask = self.min <= illumination
+            mask = np.logical_and(moon_up_mask, mask)
         elif self.min is not None and self.max is not None:
             mask = ((self.min <= illumination) &
                     (illumination <= self.max))
+            mask = np.logical_and(moon_up_mask, mask)
```
bmorris3: Would you swap these numpy logical operations for the notation used elsewhere in constraints, using `|` for `OR` and `&` for `AND` to keep the code consistent? 

For example, the three lines above could be: 
```python
mask = ((self.min <= illumination) & (illumination <= self.max)) & moon_up_mask
```

kvyh:
Changed the syntax to fit the rest of the code.

bmorris3:
Thanks @kvyh!

PR #204, opened 2016-08-03
Tonight convenience method

wtgee:
Cleaner implementation of #196 along with tests. Closes #196 

(small pep8 cleanup included in `observer.py`)

### DIFF
--- a/astroplan/observer.py
+++ b/astroplan/observer.py
@@ -1683,7 +1683,7 @@ class Observer(object):
 
         Parameters
         ----------
-        time : `~astropy.time.Time` (optional), default = `~astropy.time.Time.now()`
+        time : `~astropy.time.Time` (optional), default = `~astropy.time.Time.now`
             The start time for tonight, which is allowed to be arbitrary. See description
             above for behavior
         horizon : `~astropy.units.Quantity` (optional), default = zero degrees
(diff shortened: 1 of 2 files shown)


astroplan/observer.py
```diff
@@ -1667,3 +1670,38 @@ def target_hour_angle(self, time, target):
...
+
+        Returns
+        -------
+        times : `~astropy.time.Time`
+            A tuple of times corresponding to the start and end of current night
+        """
+        current_time = Time.now() if time is None else time
+        if self.is_night(current_time):
+            start_time = current_time
+            start_time.format = 'jd'
```
bmorris3: Out of curiosity – do you need to force the time format to `jd` here?

wtgee: The `sun_rise_time` (and `sun_set_time`) come out as `jd` for me, but `Time.now()` didn't, so if a `time` or `now()` were used then you would have inconsistent formats in the return. Happy to remove though.

bmorris3: Ah, I see. I think it's ok to have inconsistent formats here because the `Time` object behaves the same way for any format. Since we don't do this anywhere else in astroplan, let's remove that line, and then I'm ready to merge.

bmorris3:
Thanks @wtgee!

Issue #200, opened 2016-07-28
Move scheduling notebooks to tutorials in docs


bmorris3:
This will be a near-term upcoming task for @kvyh as part of his Google Summer of Code.

kvyh:
This has been completed.

PR #203, opened 2016-08-03
Scheduling tests

kvyh:
I have written tests for various parts of the scheduling as well as fixed a couple of errors that showed up as I was testing.

PR #208, opened 2016-08-03
Scheduling Tutorial

kvyh:
I have added  a scheduling tutorial to the docs. It relies on a couple of the changes from #203, though isn't branched from there, so it will fail tests (maybe?) until that is merged.


bmorris3:
This will eventually close #200.

Would you also remove the notebooks in this PR?

docs/tutorials/scheduling.rst
```diff
@@ -0,0 +1,273 @@
...
+Say we want to observe Vega, Arcturus, Deneb, and Polaris, as well as a few
+other stars in Ursa Minor. For this we will be using the Apache Point Observatory.
+
+First we define our `~astroplan.Observer` object (where we are observing from):
+
+.. code-block:: python
+
+    >>> from astroplan import Observer
+
+    >>> apo = Observer.at_site('apo)
```
bmorris3: That's going to need another `'`.

bmorris3: Don't forget to build locally with both Python 2 and 3 before pushing to GitHub to catch things like this, with 
```
python setup.py build_sphinx -lo
```

kvyh:
The only sphinx warning remaining is one that should be fixed by #203. The code examples are not good for doctests (line breaks apparently cause trouble) so they are now skipped.

kvyh:
This is superseded by #218 so I will close it.

PR #210, opened 2016-08-04
Creating a Scorer

kvyh:
I wrote up a scorer. It isn't currently being used, but once I take the time to move the `PriorityScheduler` over to it, it should make it easier to modify, optimize, and show how scoring is being done.

### DIFF
--- a/astroplan/scheduling.py
+++ b/astroplan/scheduling.py
@@ -13,10 +13,6 @@ import numpy as np
 
 from astropy import units as u
 from astropy.table import Table
-from astroplan.constraints import (AltitudeConstraint, AirmassConstraint, AtNightConstraint,
-                                   SunSeparationConstraint, MoonSeparationConstraint,
-                                   MoonIlluminationConstraint, LocalTimeConstraint, Constraint,
-                                   TimeConstraint)
 
 from .utils import time_grid_from_range, stride_array
 
--- a/astroplan/scheduling.py
+++ b/astroplan/scheduling.py
@@ -120,8 +120,6 @@ class Scorer(object):
         time_resolution : `~astropy.units.Quantity`
             the time between each scored time
         """
-        # TODO: iterate through blocks and pull out constraints instead
-        # make it not need the constraint_list
         start = self.schedule.start_time
         end = self.schedule.end_time
         times = time_grid_from_range((start, end), time_resolution)
--- a/astroplan/scheduling.py
+++ b/astroplan/scheduling.py
@@ -109,12 +109,6 @@ class Scorer(object):
         self.blocks = blocks
         self.observer = observer
         self.schedule = schedule
-        # the lists below should change if there are new constraints made
-        self.constraint_list = ['AltitudeConstraint', 'AirmassConstraint', 'AtNightConstraint',
-                                'SunSeparationConstraint', 'MoonSeparationConstraint',
-                                'MoonIlluminationConstraint', 'LocalTimeConstraint', 'Constraint',
-                                'TimeConstraint']
-        self.scheduling_constraints = None
 
     def create_score_array(self, time_resolution=1*u.minute):
         """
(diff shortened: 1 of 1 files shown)


astroplan/scheduling.py
```diff
@@ -13,6 +13,8 @@
 
 from astropy import units as u
 from astropy.table import Table
+import astroplan.constraints
+from astroplan.constraints import *
```
bmorris3: Please use `from .constraints import (AirmassConstraint...)`.

astroplan/scheduling.py
```diff
@@ -90,6 +92,81 @@ def from_exposures(cls, target, priority, time_per_exposure,
...
+        self.blocks = blocks
+        self.observer = observer
+        self.schedule = schedule
+        # the ranges below should change if there are new constraints made
+        self.constraint_list = [eval(name) for name in
+                                astroplan.constraints.__all__[:-6]
+                                if 'Constraint' in name]
+        self.scheduling_constraints = [eval(name) for name in
+                                       astroplan.constraints.__all__[-5:]
+                                       if 'Constraint' in name]
```
bmorris3: What's this about? `eval`s are kind of dangerous and should be avoided.

kvyh: The use of eval here could cause issues (pointed out by @StuartLittlefair). @bmorris3 is it possible to make these lists part of the `Constraints`? That way it could just call the list from `Constraints`.

kvyh: There are a few ways this can be resolved: have only global constraints be vectorized across targets a la #190 (which could work, but would require care in the scheduler to retain the order of `ObservingBlocks` compared to the vectorized `Constraints`); have it evaluate each target individually (this eliminates any use of the optimizations in vectorizing the constraints); or have a list of all the `Constraints` defined in `scheduling.py` or 'constraints.py`. All three of the solutions above solve the import issue above as a by-product. I will try to see if I can implement the third solution, as it preserves most of the potential optimisation, and fall back on the other two if I run into too many issues.

kvyh:
A change that I had made earlier, using `.__class__.__name__` means that I don't need the `import` statements or the 'eval'. They have been removed.

astroplan/scheduling.py
```diff
@@ -90,6 +90,80 @@ def from_exposures(cls, target, priority, time_per_exposure,
...
+        observer : `~astroplan.Observer`
+            the observer
+        schedule : `~astroplan.scheduling.Schedule`
+            The schedule inside which the blocks should fit
+        """
+        self.blocks = blocks
+        self.observer = observer
+        self.schedule = schedule
+        # the lists below should change if there are new constraints made
+        self.constraint_list = ['AltitudeConstraint', 'AirmassConstraint', 'AtNightConstraint',
```
StuartLittlefair: @kvyh - there may be a better way to do this. We can make the ```Constraint``` class have a MetaClass, which registers any subclass of ```Constraint``` that gets defined. This will then provide you with a list, or dictionary or whatever, of all defined ```Constraints```; be they ones that are implemented natively in astroplan or defined by the end user.

See http://www.effectivepython.com/2015/02/02/register-class-existence-with-metaclasses/ for details.

StuartLittlefair: On this subject, see also http://stackoverflow.com/questions/3862310/how-can-i-find-all-subclasses-of-a-given-class-in-python

```Constraint``` subclasses ```object```, so the approach in this thread will work, i.e you can get a list of all ```Constraints``` using

```python
from astroplan.constraints import Constraint
Constraint.__subclasses__()
```

Note that there's a bug that needs fixing for this to work properly; ```AirmassConstraint``` currently subclasses ```AltitudeConstraint``` rather than ```Constraint```.

kvyh: `AirmassConstraint` subclassing `AltitudeConstriant` is useful in other places: the schedule needs to make sure there is some `Constraint` on the altitude, so it can just check if any of the constraints are an `AltitudeConstraint`. The MetaClass idea seems interesting, but I'm not sure it's necessary. A recursive version of `__subclasses__()` should be able to handle it (and any user-created `Constraints`).

bmorris3: In our Google Hangout conversation today, we've decided that we can take a different approach: first loop over each block and append constraints that you find in each block to a list (if they're not already in the list). That way you don't have to know anything about what constraints have been implemented, which will naturally support user-defined constraints (which are really important!).

kvyh:
The new commits should fix most of the dangerous/rigid parts of the class.

astroplan/scheduling.py
```diff
@@ -90,6 +90,73 @@ def from_exposures(cls, target, priority, time_per_exposure,
...
+                score_array[i] *= applied_score
+        # for constraint in constraints:
+        #     constraint_list = [a[0] for a in constraints[constraint]]
+        #     index_list = [a[1] for a in constraints[constraint]]
+            # I'm not sure about the api of this, but this should eliminate need for importing constraints
+        #     vectorized = constraints[constraint][0][0].vectorize(constraint_list)
+        #     targets = [self.blocks[i].target for i in index_list]
+        #     scores = vectorized(self.observer, targets, times = times)
+        #     for i in len(index_list):
+        #         score_array[index_list[i]] *= scores[i]
```
bmorris3: Is there a reason these commented out sections are still in the code?

kvyh: I wanted to retain the code, I should move it to a different branch though.

bmorris3:
Thanks @kvyh!

PR #218, opened 2016-08-05
Tutorial for Scheduling

kvyh:
This contains the Scheduling Tutorial in the docs, but was rebased off of code that it requires to function. This will replace #208 and is intended to solve #200 .

There is a work left to do on the tutorial itself so that it better reflects how it would be used, so I will be rewriting specific parts of it over the coming days.


bmorris3:
@kvyh You'll have to rebase this branch.

kvyh:
The conflict was in a part of the code that I no longer needed, so I removed it.

kvyh:
It looks like the commit fixed the issue that was occurring with numpy1.7 (I don't really understand why), but the checks for this branch are running very slowly. Is there any way to help with the speed?

astroplan/tests/test_scheduling.py
```diff
@@ -21,70 +22,182 @@
                                      dec=89.26410897 * u.deg), name="Polaris")
 
 apo = Observer.at_site('apo')
-targets = [vega, rigel, polaris]
+targets = [vega, polaris, rigel]
+default_time = Time('2016-02-06 03:00:00')
```
bmorris3: Could you call this "tricky_time" or something that indicates that this is a time where float evaluations go bad, that you're testing specifically?

astroplan/tests/test_scheduling.py
```diff
@@ -21,70 +22,182 @@
...
+
+
+def test_schedule_insert_slot():
+    schedule = Schedule(default_time, default_time + 5*u.hour)
+    duration = 2*u.hour + 1*u.second
+    end_time = default_time + duration
+    block = TransitionBlock.from_duration(duration)
+    schedule.insert_slot(end_time - duration, block)
+    # even when float evaluation doesn't work, it should still schedule properly
+    assert not end_time - duration == default_time
```
bmorris3: Could you label this test to indicate that it's designed to test where the ordinary float comparison would fail? Could you also test a case where the float comparison wouldn't fail, and that `insert_slot` still works?

astroplan/tests/test_scheduling.py
```diff
@@ -21,70 +22,182 @@
...
+    assert len(schedule.slots) == 2
+    assert schedule.slots[0].start == default_time
+
+
+def test_schedule_change_slot_block():
+    schedule = Schedule(default_time, default_time + 5 * u.hour)
+    duration = 2 * u.hour
+    block = TransitionBlock.from_duration(duration)
+    schedule.insert_slot(default_time, block)
+    assert np.abs(schedule.slots[0].end - 2*u.hour - default_time) < 1*u.second
```
bmorris3: Would you change `2*u.hour` to `duration`?

astroplan/tests/test_scheduling.py
```diff
@@ -21,70 +22,182 @@
...
+    transition1 = trans(blocks[0], blocks[1], start_time, apo)
+    transition2 = trans(blocks[0], blocks[2], start_time, apo)
+    transition3 = trans(blocks[1], blocks[0], start_time, apo)
+    assert np.abs(transition1.duration - 2*u.minute) < 1*u.second
+    assert np.abs(transition2.duration - 2*u.minute - transition.duration) < 1*u.second
+    # to test the default transition
+    assert np.abs(transition3.duration - 5*u.minute) < 1*u.second
+    assert transition1.components is not None
+
 transitioner = Transitioner(slew_rate=1 * u.deg / u.second)
```
bmorris3: Would you give this a name like `default_transitioner`?

astroplan/tests/test_scheduling.py
```diff
@@ -21,70 +22,182 @@
...
+    constraints = [AtNightConstraint(), MoonIlluminationConstraint(max=0)]
+    scheduler1 = SequentialScheduler(start_time, end_time, constraints, apo,
+                                     transitioner)
+    schedule1 = scheduler1(block)
+    assert len(schedule1.observing_blocks) == 0
+    scheduler2 = PriorityScheduler(start_time, end_time, constraints, apo,
+                                   transitioner)
+    schedule2 = scheduler2(block)
+    assert len(schedule2.observing_blocks) == 0
+'''
```
bmorris3: Should this be commented out, or removed?

docs/tutorials/scheduling.rst
```diff
@@ -0,0 +1,258 @@
...
+combination. An observing block needs a target, a duration, and a priority;
+configuration information can also be given (i.e. filter, instrument, etc.).
+For each filter we want 17 exposures per target (100 seconds for M13 and 60
+seconds for Deneb) and the instrument has a read-out time of 20 seconds.
+We also need to consider that the moon can be a problem for observations
+in the B filter, so we want the moon to be down or dim for the B filter.
+
+.. code-block:: python
+
+    >>> from astroplan import ObservingBlock
```
bmorris3: If you're defining a code block that won't produce any printed outputs, don't use `>>>`.

docs/tutorials/scheduling.rst
```diff
@@ -0,0 +1,258 @@
...
+    >>> from astroplan.constraints import MoonIlluminationConstraint
+    >>> from astropy import units as u
+
+    >>> rot = 20 * u.second
+    >>> blocks = []
+
+    >>> for filter in ['B', 'G', 'R']:
+    >>>     if filter == 'B':
```
bmorris3: If you're indenting, use `...` rather than `>>>` (if you're using those symbols at all, see above comment).

docs/tutorials/scheduling.rst
```diff
@@ -0,0 +1,258 @@
...
+    >>> import matplotlib.pyplot as plt
+
+    >>> plt.figure(figsize = (14,6))
+    >>> plot_schedule_airmass(priority_schedule)
+    >>> plt.legend(loc = 1)
+    >>> plt.show()
+
+.. plot::
+
+    import astropy.units as u
```
bmorris3: It'd be great to put lots of comments in this code, since people will be able to click the `Source Code` button and see how you made the plot.

docs/tutorials/scheduling.rst
```diff
@@ -0,0 +1,258 @@
...
+
+Now we want to define our list of targets (`~astroplan.FixedTarget` objects),
+any object that is in SIMBAD can be called by an identifier.
+
+.. code-block:: python
+
+    >>> from astroplan import FixedTarget
+
+    >>> Deneb = FixedTarget.from_name('deneb')
+    >>> M13 = FixedTarget.from_name('m13')
```
bmorris3: When you can, label variables with lowercase letters. Upper case first letters are preferred for classes.

bmorris3:
Does this replace #203?

astroplan/scheduling.py
```diff
@@ -683,7 +683,7 @@ def _make_schedule(self, blocks):
                         end_idx = times_indices + start_idx
                         # this may make some OBs get sub-optimal scheduling, but it closes gaps
                         # TODO: determine a reasonable range inside which it gets shifted
-                        if (tb.duration > new_start_time - tb.start_time or
+                        if (new_start_time - tb.start_time < tb.duration or
```
bmorris3: lolwut

bmorris3:
Hey @bsipocz – some doctests are failing when we try to import matplotlib in the docs now. Could that be a result of any of the recent CI changes? Just want to check with you before I do some real digging.

bsipocz:
@bmorris3 - According to the docs, mpl is optional dependency only, so no wonder the tests fail when it is not available.

astroplan/tests/test_scheduling.py
kvyh: That default time is used in all of the tests, not just the insert_slot, would it be better to create the Time object separately for each test?

astroplan/tests/test_scheduling.py
kvyh: I will see if I can modify it so it isn't very slow. If I can't, I'll remove it into a separate branch.

kvyh:
That can easily be fixed by just skipping the doctests for it, like the plotting tutorial does.

astroplan/tests/test_scheduling.py
bmorris3: It's probably better to move the time into each test, because ideally in the future that will be updated to test multiple times (having only one test time is a pretty narrow test).

bmorris3:
@kvyh Doctests are useful and should be run whenever we can run them. 

@bsipocz Is there a way to ensure that matplotlib is installed for the doctests, without requiring matplotlib?

bsipocz:
There is ``__doctest_requires__`` where you can list the functions with
requirements as dictionary and also ``doctest-requires:: matplotlib`` for
docs files. There should be many examples in astropy core for the exact
usage.

bmorris3:
Reminder to rebase this branch after the changes in #219.

kvyh:
I rebased this and changed the tests to use the new `_call_` for the schedulers.

bmorris3:
I'm currently trying to build the sphinx docs from your tutorial, and they're taking forever. Either there's a problem with the code in there, or the code is way too slow.

kvyh:
The reason why it takes forever, is that it uses the still very slow `MoonIlluminationConstraint`.

docs/tutorials/scheduling.rst
```diff
@@ -0,0 +1,269 @@
...
+be observed at an airmass greater than 3, but we prefer a better airmass.
+Usually constraints scores are boolean, but with ``boolean_constraint = False``
+the constraint will output floats instead, indicated when it is closer to ideal.
+
+.. code-block:: python
+
+    >>> from astroplan.constraints import AtNightConstraint, AirmassConstraint
+
+    >>> global_constraints = [AirmassConstraint(max = 3, boolean_constraint = False),
+    ...                       AtNightConstraint()]
```
bmorris3: Could you change this to `AtNightConstraint.twilight_civil()`?

docs/tutorials/scheduling.rst
```diff
@@ -0,0 +1,269 @@
...
+We also need to define when we will be observing the targets, `~astropy.time.Time`
+objects for the start and end of our observing window. The half night goes from 7PM
+local time to 1AM local time, in UTC this will be from 2AM to 8AM.
+
+.. code-block:: python
+
+    >>> from astropy.time import Time
+
+    >>> start_time = Time('2016-06-07 02:00')
+    >>> end_time = Time('2016-06-07 08:00')
```
bmorris3: Make these times the sunset time and the sunrise time, or noon before and noon after.

docs/tutorials/scheduling.rst
```diff
@@ -0,0 +1,269 @@
...
+    >>> from astroplan import ObservingBlock
+    >>> from astroplan.constraints import MoonIlluminationConstraint
+    >>> from astropy import units as u
+
+    >>> rot = 20 * u.second
+    >>> blocks = []
+
+    >>> for filter in ['B', 'G', 'R']:
+    ...     if filter == 'B':
+    ...         constraints = [MoonIlluminationConstraint(max = 0.25)]
```
bmorris3: Swap for time constraint and make the observation fit in the A half (first half).

docs/tutorials/scheduling.rst
```diff
@@ -0,0 +1,269 @@
...
+    global_constraints = [AirmassConstraint(max = 3, boolean_constraint = False),
+                          AtNightConstraint()]
+    rot = 20 * u.second
+    blocks = []
+    for filter in ['B', 'G', 'R']:
+        if filter == 'B':
+            constraints = [MoonIlluminationConstraint(max = 0.25)]
+        else:
+            constraints = None
+        blocks.append(ObservingBlock.from_exposures(Deneb, 1, 60*u.second, 17, rot,
```
bmorris3: I think both targets need the same priority because we want to observe the target _and_ the standard.

docs/tutorials/scheduling.rst
```diff
@@ -0,0 +1,269 @@
...
+.. code-block:: python
+
+    >>> from astroplan import ObservingBlock
+    >>> from astroplan.constraints import MoonIlluminationConstraint
+    >>> from astropy import units as u
+
+    >>> rot = 20 * u.second
+    >>> blocks = []
+
+    >>> for filter in ['B', 'G', 'R']:
```
bmorris3: Avoid using the variable name `filter`, which overlaps with the Python built-in [`filter`](https://docs.python.org/2/library/functions.html#filter). 

Change this [here](https://github.com/astropy/astroplan/pull/218/files#diff-2adf7929c5a8b9d67c44fcc2e520c013R239) also.

docs/tutorials/scheduling.rst
```diff
@@ -0,0 +1,269 @@
...
+    >>> priority_schedule = prior_scheduler(blocks, schedule)
+
+Now that you have a schedule there are a few ways of viewing it.
+One way is to have it print a table where you can show, or hide,
+unused time and transitions with ``show_transitions`` and
+``show_unused`` (default is showing transitions and not unused).
+
+.. code-block:: python
+
+    >>> sequential_schedule.to_table()
```
bmorris3: Copy and paste the output from your terminal under this line as an output.

docs/tutorials/scheduling.rst
```diff
@@ -0,0 +1,269 @@
...
+                                        observer = apo, transitioner = transitioner)
+    # Create a schedule for the scheduler to insert the blocks into, and run the scheduler
+    schedule = Schedule(start_time, end_time)
+    priority_schedule = prior_scheduler(blocks, schedule)
+
+    # To get a plot of the airmass vs where the blocks were scheduled
+    plt.figure(figsize = (14,6))
+    plot_schedule_airmass(priority_schedule)
+    plt.legend(loc=1)
+    plt.show()
```
bmorris3: The final plot has three legend labels for M13 and three labels for Deneb. Is it possible to make it only add labels once per target name?

docs/tutorials/scheduling.rst
```diff
@@ -0,0 +1,269 @@
...
+    from astropy.time import Time
+    from astroplan import (Observer, FixedTarget, ObservingBlock, Transitioner, PriorityScheduler,
+                           Schedule)
+    from astroplan.constraints import AtNightConstraint, AirmassConstraint, MoonIlluminationConstraint
+    from astroplan.plots import plot_schedule_airmass
+    import matplotlib.pyplot as plt
+
+    # Now we define the targets, start time, end time, and observer.
+    Deneb = FixedTarget.from_name('deneb')
+    M13 = FixedTarget.from_name('m13')
```
bmorris3: Please capitalize Deneb and M13.

docs/tutorials/scheduling.rst
```diff
@@ -0,0 +1,269 @@
...
+targets.
+
+.. code-block:: python
+
+    >>> from astroplan.plots import plot_schedule_airmass
+    >>> import matplotlib.pyplot as plt
+
+    >>> plt.figure(figsize = (14,6))
+    >>> plot_schedule_airmass(priority_schedule)
+    >>> plt.legend(loc = 1)
```
bmorris3: Would you use the location key `"upper right"` rather than `1`? It's a bit more descriptive.

kvyh:
I have added the suggestions and rebased on the newest changes, it is much faster to build now.

bmorris3:
Thanks @kvyh! This is a big one.

Issue #217, opened 2016-08-04
Calling Schedulers with Schedules


kvyh:
The current  code creates a `Schedule` during the initialization. I will be removing the creation of a `Schedule` from the `__init__` and adding a `Schedule` as a required argument for the `__call__` (which will then require a list of `ObservingBlocks` and a `Schedule`). After the change it will be possible to input a `Schedule` that has already be added to into the `Schedulers`. 

Using a partially filled `Schedule` is **not** advised since neither `Scheduler` can accommodate it at this time. Adding this functionality is a future goal.

bmorris3:
I think this was addressed by the above mentioned PRs.

PR #219, opened 2016-08-05
Changed how Schedulers get their Schedule.

kvyh:
I moved the `Schedule` to be something created by the user and given to the `Scheduler` as part of the call. This means that the `Scheduler` initialisation doesn't need start/end times anymore. This also means that a user doesn't need to use `schedule = scheduler(blocks)` and is reflected in the changes to the tests.


kvyh:
This is the PR for the first part of #217.

bmorris3:
Should this be merged before your docs/tests updates for scheduling, so that you can rebase that PR afterwards and include this new API change?

kvyh:
That would be good.

PR #220, opened 2016-08-05
Plot Airmass Improvments

wtgee:
* Accept either a single target or a list of targets
* Brightness shading that actually works (from @joshwalawender in panoptes/PEAS)

**Note: Really this should all be redone to use a real plt.figure and axes**

### DIFF
--- a/astroplan/plots/time_dependent.py
+++ b/astroplan/plots/time_dependent.py
@@ -188,7 +188,7 @@ def plot_airmass(targets, observer, time, ax=None, style_kwargs=None,
         ]
 
         for i, twi in enumerate(twilights[1:], 1):
-            plt.axvspan(twilights[i - 1][0], twilights[i][0], ymin=0, ymax=1, color='grey', alpha=twi[1])
+            ax.axvspan(twilights[i - 1][0], twilights[i][0], ymin=0, ymax=1, color='grey', alpha=twi[1])
 
     # Invert y-axis and set limits.
     if ax.get_ylim()[1] > ax.get_ylim()[0]:
--- a/astroplan/plots/time_dependent.py
+++ b/astroplan/plots/time_dependent.py
@@ -45,7 +45,7 @@ def _has_twin(ax):
     return False
 
 
-def plot_airmass(target, observer, time, ax=None, style_kwargs=None,
+def plot_airmass(targets, observer, time, ax=None, style_kwargs=None,
                  style_sheet=None, brightness_shading=False,
                  altitude_yaxis=False):
     """
@@ -69,8 +69,9 @@ def plot_airmass(target, observer, time, ax=None, style_kwargs=None,
 
     Parameters
     ----------
-    target : `~astroplan.FixedTarget`
-        The celestial body of interest.
+    targets : list of `~astroplan.FixedTarget` objects
+        The celestial bodies of interest.
+        If a single object is passed it will be converted to a list.
 
     observer : `~astroplan.Observer`
         The person, telescope, observatory, etc. doing the observing.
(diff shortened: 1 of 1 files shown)


wtgee:
figure_1

bmorris3:
This looks nice. I have two questions – when you say "Really this should all be redone to use a real plt.figure and axes", what do you mean? And would you agree that it would be easier to read if the default `axvspan` color were gray?

wtgee:
My understanding is that pyplot should be used for interactive things but you should use the OO interface otherwise. But my understanding of matplotlib isn't that great and pyplot is always easier for me anyway. :).

From the site: "For non-interactive plotting it is suggested to use pyplot to create the figures and then the OO interface for plotting."

http://matplotlib.org/faq/usage_faq.html

I'll test it out with grey too and post a pic.

wtgee:
Yeah, I like the grey more, good call. I kept the slightly thicker lines.

figure_2

astroplan/plots/time_dependent.py
```diff
@@ -145,40 +147,49 @@ def plot_airmass(target, observer, time, ax=None, style_kwargs=None,
                       'to use a scalar. (Or maybe a list with length > 1?).',
                       PlotWarning)
 
-    # Calculate airmass
-    airmass = observer.altaz(time, target).secz
-    # Mask out nonsense airmasses
-    masked_airmass = np.ma.array(airmass, mask=airmass < 1)
+    if not isinstance(targets, list):
```
MSeifert04: What about testing for `collections.Sequence` or `collections.Iterable` here? For example the current implementation would wrap a tuple of targets as list. So I would end up with a tuple of a list of targets.

I don't think that will happen often but it's probably more user-friendly.

astroplan/plots/time_dependent.py
```diff
@@ -145,40 +147,49 @@ def plot_airmass(target, observer, time, ax=None, style_kwargs=None,
...
+            (observer.twilight_evening_astronomical(Time(start), which='next').datetime, 0.3),
+            (observer.twilight_morning_astronomical(Time(start), which='next').datetime, 0.4),
+            (observer.twilight_morning_nautical(Time(start), which='next').datetime, 0.3),
+            (observer.twilight_morning_civil(Time(start), which='next').datetime, 0.2),
+            (observer.sun_rise_time(Time(start), which='next').datetime, 0.1),
+        ]
+        twilights.sort(key=lambda x: x[0])
+
+        for i, twi in enumerate(twilights):
+            if i > 0:
```
MSeifert04: you could make the loop exclude the first element and then remove the following condition with `for i, twi in enumerate(twilights[1:], 1):` Starts with a counter of 1 and the second element of the twilights.

Another alternative would be to leave the loop and just check `if i` (which evaluates to true for all integers except 0).

astroplan/plots/time_dependent.py
wtgee: Yes, that probably is cleaner. Any preference between the two? It assumes it is a sequence of targets and I don't really see someone passing their own iterator.

astroplan/plots/time_dependent.py
```diff
@@ -145,40 +147,49 @@ def plot_airmass(target, observer, time, ax=None, style_kwargs=None,
...
+            (observer.sun_set_time(Time(start), which='next').datetime, 0.0),
+            (observer.twilight_evening_civil(Time(start), which='next').datetime, 0.1),
+            (observer.twilight_evening_nautical(Time(start), which='next').datetime, 0.2),
+            (observer.twilight_evening_astronomical(Time(start), which='next').datetime, 0.3),
+            (observer.twilight_morning_astronomical(Time(start), which='next').datetime, 0.4),
+            (observer.twilight_morning_nautical(Time(start), which='next').datetime, 0.3),
+            (observer.twilight_morning_civil(Time(start), which='next').datetime, 0.2),
+            (observer.sun_rise_time(Time(start), which='next').datetime, 0.1),
+        ]
+        twilights.sort(key=lambda x: x[0])
```
MSeifert04: One alternative to the `lambda` here is `operator.itemgetter(0)`.

Although for short lists like these the performance gain from the `itemgetter` is probably neglectable.

wtgee: Now that I look at it I'm not sure the sort is actually needed at all.

astroplan/plots/time_dependent.py
MSeifert04: The practical approach would be to use `Iterable` because that's the prerequisite of the `for` loop (you need an iterable to do a for loop). But I think the only (rare) real use-case where it makes a difference is a generator, which is an iterable but not a sequence.

wtgee: Yeah, that was my thought, and I don't see someone make a generator to output targets but I suppose it could happen. I set it to `Sequence` for now but we could switch it.

astroplan/plots/time_dependent.py
```diff
@@ -145,40 +148,47 @@ def plot_airmass(target, observer, time, ax=None, style_kwargs=None,
...
 
-    # Format the time axis
-    date_formatter = dates.DateFormatter('%H:%M')
-    ax.xaxis.set_major_formatter(date_formatter)
-    plt.setp(ax.get_xticklabels(), rotation=30, ha='right')
+        # Plot data.
+        ax.plot_date(time.plot_date, masked_airmass, label=target_name, **style_kwargs)
+
+        # Format the time axis
+        date_formatter = dates.DateFormatter('%H:%M')
```
MSeifert04: Why is this line and the following 2 inside the `for target in targets` loop? Wouldn't it suffice to do them once after the looping?

astroplan/plots/time_dependent.py
```diff
@@ -145,40 +148,47 @@ def plot_airmass(target, observer, time, ax=None, style_kwargs=None,
...
+    for target in targets:
+        # Calculate airmass
+        airmass = observer.altaz(time, target).secz
+        # Mask out nonsense airmasses
+        masked_airmass = np.ma.array(airmass, mask=airmass < 1)
 
-    # Plot data.
-    ax.plot_date(time.plot_date, masked_airmass, **style_kwargs)
+        # Some checks & info for labels.
+        if not hasattr(target, 'name'):
```
MSeifert04: I know that this is the original style but wouldn't a `try .. except ..` make more sense?

```
try:
    target_name = target.name
except AttributeError:
    target_name = ''
```

astroplan/plots/time_dependent.py
wtgee: Yup, thanks for checking. I'm taking that out of loop and making the suggestion below about the `try/except`

wtgee:
In the example the legend is added outside the `plot_airmass` function but with multiple targets it becomes a necessity to have and I think most users would just expect it to show. @bmorris3, what are your thoughts about adding it as part of the functionality?

wtgee:
I'm not sure why one of those tests started failing above. The only errors I see don't appear to be related and the reported error is about sphinx but nothing changed there. Let me know what I can do to fix it.

bmorris3:
> In the example the legend is added outside the plot_airmass function but with multiple targets it becomes a necessity to have and I think most users would just expect it to show.

@wtgee The reason we did not include multiple targets in the `plot_airmass` function was for that reason – we wanted them to be able to control the plot settings of each target individually. There will be users who still want control over the legend, so I think it should stay outside of the `plot_airmass` function.

Our intention is to keep the scope of the astroplan plotting functions narrow, so you can build on top of them as you need for your particular work.

astroplan/plots/time_dependent.py
```diff
@@ -167,18 +173,22 @@ def plot_airmass(target, observer, time, ax=None, style_kwargs=None,
...
+            (observer.twilight_evening_nautical(Time(start), which='next').datetime, 0.2),
+            (observer.twilight_evening_astronomical(Time(start), which='next').datetime, 0.3),
+            (observer.twilight_morning_astronomical(Time(start), which='next').datetime, 0.4),
+            (observer.twilight_morning_nautical(Time(start), which='next').datetime, 0.3),
+            (observer.twilight_morning_civil(Time(start), which='next').datetime, 0.2),
+            (observer.sun_rise_time(Time(start), which='next').datetime, 0.1),
+        ]
+
+        for i, twi in enumerate(twilights[1:], 1):
+            plt.axvspan(twilights[i - 1][0], twilights[i][0], ymin=0, ymax=1, color='grey', alpha=twi[1])
```
bmorris3: Can this be called as `ax.axvspan` rather than `plt.axvspan`, for consistency?

bmorris3:
Looks good, thanks @wtgee!

PR #223, opened 2016-08-05
moved moon_altaz to the non-location get_moon

kvyh:
This is a follow-up to #213 and removes the location kwarg from the `get_moon` call inside `Observer.moon_altaz` and uses the array for the following `.alt`, `.az`, and `.distance`s.

### DIFF
--- a/astroplan/observer.py
+++ b/astroplan/observer.py
@@ -1487,15 +1487,12 @@ class Observer(object):
             return moon
 
         else:
-            moon_coords = []
-            for t in time:
-                altaz_frame = AltAz(location=self.location, obstime=t)
-                moon_coord = get_moon(t, location=self.location, ephemeris=ephemeris).transform_to(altaz_frame)
-                moon_coords.append(moon_coord)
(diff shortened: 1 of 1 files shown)


bmorris3:
I left this out of #213 on purpose, because the moon alt/az will be more affected by an observer's location on the Earth than the moon phase. So I intended to leave this method as it is until astropy/astropy#5216 is merged, and then we can fix this method. Sorry I didn't explain that in #213.

kvyh:
The transformation to alt/az coordinates still uses the location. If this isn't changed, the `MoonIlluminationConstraint` needs to be rewritten, since it ends up calling `Observer.moon_altaz` which is currently still really slow for array times.

StuartLittlefair:
As I understand it, there's no issue with the general idea here. It doesn't matter what frame you transform _from_, the alt-az coordinate of the moon will be the same as long as the original coordinate points at the moon.

So using ```get_moon``` with no location and transforming to the correct alt_az frame makes sense.

PR #224, opened 2016-08-08
Fix `tonight` method to incorporate horizon at initial time check

wtgee:
Incorrect sunset times were being reported because of lack of `horizon` check on `is_night`. This happens if you check for `tonight` after `0 deg` but before passed `horizon`.

@bmorris3 I suppose wavelength should be included too?

### DIFF
--- a/astroplan/observer.py
+++ b/astroplan/observer.py
@@ -1696,7 +1696,7 @@ class Observer(object):
             A tuple of times corresponding to the start and end of current night
         """
         current_time = Time.now() if time is None else time
-        if self.is_night(current_time):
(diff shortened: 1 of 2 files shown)


bmorris3:
@wtgee Yes, wavelength could/should be included too if you want to be precise, but practically speaking I'm not sure it matters. Let me know if you intend to add it or not.

wtgee:
@bmorris3 It's only used in the `is_night` check and you are right that it probably won't make a big difference but it's easy enough to add and someone might be expecting it there. It keeps the `is_night` and `tonight` consistent.

bmorris3:
Great, thanks @wtgee!

PR #225, opened 2016-08-08
Reverse Altitude Constraint scores

wtgee:
Reverse the scores for altitude constraint so that lower altitudes give higher scores so long as they are above the `min`. If this is not the intended behaviour let me know. If is I can write a test for it (currently there are no tests for `boolean_constraint=False` constraints)

### DIFF
--- a/astroplan/constraints.py
+++ b/astroplan/constraints.py
@@ -999,11 +999,11 @@ def observability_table(constraints, observer, targets, times=None,
 
 def _rescale_minmax(vals, min_val, max_val):
     """ Rescale altitude into an observability score."""
-    rescaled = (vals - min_val) / (max_val - min_val)
+    rescaled = (max_val - vals) / (max_val - min_val)
     below = rescaled < 0
     above = rescaled > 1
-    rescaled[below] = 0
-    rescaled[above] = 1
+    rescaled[below] = 1
+    rescaled[above] = 0
 
     return rescaled
 


kvyh:
The intended behaviour is for `AltitudeConstraint` to output a higher (better) score for a higher altitude (closer to zenith). Is there a reason why we would want to flip the scoring?

`_rescale_minmax` (https://github.com/astropy/astroplan/blob/master/astroplan/constraints.py#L1000), is intended to be a general function that can be used in the non-boolean version of any `Constraint`. So if we wanted to flip the scoring it would be better to have `AltitudeConstraint` return `1-_rescale_minmax(...)` instead of changing how `_rescale_minmax` works.

wtgee:
Seems like it would depend on whether or not a meridian flip is needed or has been performed. If something is at 88° and is only observable for ten minutes before a flip is required it doesn't seem like that should have a higher score, but that could just be my thinking.

Also, if it's a generic function the docstring shouldn't specifically mention that it's for altitude. I'm happy to close this if it's not what is wanted and agree that the `1 - _rescale_minmax` is potentially the better solution.

kvyh:
In a earlier version of #165 than got merged, there was a `MeridianTransitConstraint` (https://github.com/StuartLittlefair/astroplan/blob/e96efcb81b29fb6637e26eebd98852ec1d299f66/astroplan/constraints.py#L709) that would be a solution to the case you described.

I'm going to try making a few additions to generalise `_rescale_minmax` and I'll give it a better docstring.

wtgee:
I was wondering why the `MeridianTransitConstraint` disappeared actually. :)

PR #228, opened 2016-08-09
improved docstring and generalised _rescale_minmax

kvyh:
In the case that a user wants to create a `Constraint` that is non-boolean, they will want to use `_rescale_minmax` or a similar function. This PR is trying to make the function more general and understandable by changing variable names and adding a docstring.

Any ideas for variable names, or better descriptions for the docstring are welcome.

### DIFF
--- a/astroplan/constraints.py
+++ b/astroplan/constraints.py
@@ -28,7 +28,8 @@ __all__ = ["AltitudeConstraint", "AirmassConstraint", "AtNightConstraint",
            "is_observable", "is_always_observable", "time_grid_from_range",
            "SunSeparationConstraint", "MoonSeparationConstraint",
            "MoonIlluminationConstraint", "LocalTimeConstraint", "Constraint",
-           "TimeConstraint", "observability_table", "months_observable"]
+           "TimeConstraint", "observability_table", "months_observable",
+           "rescale_minmax"]
 
 
 def _get_altaz(times, observer, targets,
@@ -286,7 +287,7 @@ class AltitudeConstraint(Constraint):
             uppermask = alt <= self.max
             return lowermask & uppermask
         else:
-            return _rescale_minmax(alt, self.min, self.max)
+            return rescale_minmax(alt, self.min, self.max)
 
 
 class AirmassConstraint(AltitudeConstraint):
@@ -345,7 +346,7 @@ class AirmassConstraint(AltitudeConstraint):
 
             mi = 1 if self.min is None else self.min
             # we reverse order so that airmass close to 1/min is good
-            return _rescale_minmax(secz, mi, mx, better_than=0)
+            return rescale_minmax(secz, mi, mx, better_than=0)
 
 
 class AtNightConstraint(Constraint):
@@ -997,9 +998,9 @@ def observability_table(constraints, observer, targets, times=None,
     return tab
 
 
-def _rescale_minmax(vals, worst_val, best_val, better_than=1, worse_than=0):
+def rescale_minmax(vals, worst_val, best_val, better_than=1, worse_than=0):
     """
-    rescales the input ``vals`` between 0 and 1
+    rescales an input array ``vals`` between zero and one
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


astroplan/constraints.py
```diff
@@ -997,24 +997,37 @@ def observability_table(constraints, observer, targets, times=None,
...
+    vals : array of values
+        the values that need to be rescaled to be between 0 and 1
+    worst_val : value
+        worst acceptable value (rescales to 0)
+    best_val : value
+        best value cared about (rescales to 1)
+    better_than : 0 or 1
+        what is returned for ``vals`` beyond the ``best_val``
+    worse_than : 0 or 1
+        what is returned for ``vals`` beyond the ``worst_val``
```
bmorris3: Would you include an example or two demonstrating how it works? Here's an [example](https://github.com/astropy/astroplan/blob/master/astroplan/observer.py#L317-L333) to refer to for the format. For example, you could use `vals = altitude = [0, 10, 20]` and `vals = airmass = [3, 2, 1]` and show how the outputs match what you'd expect.

kvyh:
I added some examples as well as some tests for it.

astroplan/constraints.py
```diff
@@ -997,24 +997,59 @@ def observability_table(constraints, observer, targets, times=None,
...
-    rescaled[below] = 1
-    rescaled[above] = 1
+    Examples
+    --------
+    rescale altitude = [20, 30, 40, 45, 55, 70] to between
+    0 and 1, with the best at 60->1 and worst at 35->0. 0
+    below 35 and 1 above 60.
+    >>> from astroplan.constraints import _rescale_minmax
+    >>> import numpy as np
+    >>> airmasses = np.array([20, 30, 40, 45, 55, 70])
```
bmorris3: I think you mean altitudes rather than airmasses here (and in the line below). Also – would you use a complete sentence in the description above, without repeating the variable definitions verbatim?

astroplan/constraints.py
```diff
@@ -997,24 +997,59 @@ def observability_table(constraints, observer, targets, times=None,
...
+    >>> _rescale_minmax(airmasses, 35, 60)
+    array([ 0. , 0. , 0.2, 0.4, 0.8, 1. ])
+
+    rescale airmasses = [1,1,1.5,2,2,2,3,3,3,2,2,1] to between
```
bmorris3: Would you reduce the number of redundant airmasses here? A good example might be `[3, 2, 1]` because it would show the full range.

kvyh:
@bmorris3, do you have any idea why https://travis-ci.org/astropy/astroplan/jobs/151028147#L751 might be happening?

bmorris3:
I don't, but the error goes away if I restart the build.

kvyh:
That's good, now I just need to find a fix for the `Quantity/TimeDelta` comparison.

astroplan/constraints.py
```diff
@@ -997,24 +997,59 @@ def observability_table(constraints, observer, targets, times=None,
...
 
 
-def _rescale_minmax(vals, min_val, max_val):
-    """ Rescale altitude into an observability score."""
-    rescaled = (vals - min_val) / (max_val - min_val)
-    below = rescaled < 0
-    above = rescaled > 1
-    rescaled[below] = 0
-    rescaled[above] = 1
+def _rescale_minmax(vals, worst_val, best_val, better_than=1, worse_than=0):
```
bmorris3: I think this function should be public (remove the leading `_`, added to `__all__`) since it will be used in custom constraints.

astroplan/constraints.py
```diff
@@ -997,24 +997,59 @@ def observability_table(constraints, observer, targets, times=None,
...
+    """
+    rescales the input ``vals`` between 0 and 1
```
bmorris3: You phrase this like `Rescales an input array ``vals`` between zero and one.`

kvyh:
Ok, I changed it to a public function and changed the description.

kvyh:
WIth #232's changes, this now passes.

kvyh:
It appears that this will be replaced by #236.

PR #230, opened 2016-08-10
Scorer testing

kvyh:
Added some tests for the scorer (#210).

### DIFF
--- a/astroplan/tests/test_scheduling.py
+++ b/astroplan/tests/test_scheduling.py
@@ -88,3 +88,20 @@ def test_schedule():
     assert np.abs(new_slots[1].duration - 3*u.hour) < 1*u.second
     assert np.abs(new_slots[2].duration - 20*u.hour) < 1*u.second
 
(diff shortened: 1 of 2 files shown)


astroplan/tests/test_scheduling.py
```diff
@@ -88,3 +88,17 @@ def test_schedule():
     assert np.abs(new_slots[1].duration - 3*u.hour) < 1*u.second
     assert np.abs(new_slots[2].duration - 20*u.hour) < 1*u.second
 
+
+def test_scorer():
+    constraint = AirmassConstraint(max=4)
+    c = constraint(apo, [vega], Time('2016-02-06 00:00:00'))
```
bmorris3: Would you do this test with vector times, even with just two times in the time vector? That way we can be sure the dimensions don't do anything funky.

kvyh:
Vector tests added for times and targets

bmorris3:
Thanks @kvyh.

PR #232, opened 2016-08-15
fix for the errors with numpy1.7

kvyh:
This seems to fix the errors that are appearing with numpy1.7 in other branches (#228, #218), which I'll have to rebase or modify after this gets merged.


bmorris3:
This error was a weird one, but that looks fine to me.

PR #234, opened 2016-08-16
using scorer in priority scheduler

kvyh:
With this, the priority scheduler makes use of the scorer. The main benefit is that it vectorizes global constraints.


bmorris3:
Is this covered by the existing tests? (I think so, but I want to check before merging).

kvyh:
I cherry-picked it into a branch that has all the tests for the scheduling and the priority-scheduler using the scorer passed all those tests.

bmorris3:
Great!

PR #236, opened 2016-08-17
two functions for rescaling

kvyh:
This is an alternate form of #228, with different functions for when the score gets better or worse for larger values. Would this be better than the single function in #228?

### DIFF
--- a/astroplan/constraints.py
+++ b/astroplan/constraints.py
@@ -28,7 +28,8 @@ __all__ = ["AltitudeConstraint", "AirmassConstraint", "AtNightConstraint",
            "is_observable", "is_always_observable", "time_grid_from_range",
            "SunSeparationConstraint", "MoonSeparationConstraint",
            "MoonIlluminationConstraint", "LocalTimeConstraint", "Constraint",
-           "TimeConstraint", "observability_table", "months_observable"]
+           "TimeConstraint", "observability_table", "months_observable",
+           "rescale_minmax"]
 
 
 def _get_altaz(times, observer, targets,
@@ -286,7 +287,7 @@ class AltitudeConstraint(Constraint):
             uppermask = alt <= self.max
             return lowermask & uppermask
         else:
-            return _rescale_minmax(alt, self.min, self.max)
+            return rescale_minmax(alt, self.min, self.max)
 
 
 class AirmassConstraint(AltitudeConstraint):
@@ -345,7 +346,7 @@ class AirmassConstraint(AltitudeConstraint):
 
             mi = 1 if self.min is None else self.min
             # we reverse order so that airmass close to 1/min is good
-            return _rescale_minmax(secz, mi, mx, better_than=0)
+            return rescale_minmax(secz, mi, mx, better_than=0)
 
 
 class AtNightConstraint(Constraint):
@@ -997,9 +998,9 @@ def observability_table(constraints, observer, targets, times=None,
     return tab
 
 
-def _rescale_minmax(vals, worst_val, best_val, better_than=1, worse_than=0):
+def rescale_minmax(vals, worst_val, best_val, better_than=1, worse_than=0):
     """
-    rescales the input ``vals`` between 0 and 1
+    rescales an input array ``vals`` between zero and one
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


kvyh:
`min_value`, `max_value` and `best_value`, `worst_value` are both easy to understand, the issue I'm having is in naming the regions beyond the best and worst values. The options I've though of are: (`greater_than`, `less_than`) ; (`above_max`, `below_min`) ; (`better_than` , `worse_than`) ; (`beyond_best` , `beyond_worst`). The first two are for `min,max` and latter two are for `best,worst`.

bmorris3:
Would you add a note on this in the User-Defined Constraint tutorial?

astroplan/constraints.py
```diff
@@ -997,24 +998,90 @@ def observability_table(constraints, observer, targets, times=None,
...
 
 
-def _rescale_minmax(vals, min_val, max_val):
-    """ Rescale altitude into an observability score."""
-    rescaled = (vals - min_val) / (max_val - min_val)
-    below = rescaled < 0
-    above = rescaled > 1
-    rescaled[below] = 0
-    rescaled[above] = 1
+def min_best_rescale(vals, min_val, max_val, greater_than=0, less_than=1):
```
eteq: How about naming the last two arguments (which might be just one if you make the changes I'm about to discuss in the comments) as "`greater_score` or similar?  That makes it clearer that you're saying what the *output* is, rather than what input value is.

Similarly, perhaps you should use the word "score" in the docstring.  E.g. 
```
Rescales the input array ``vals`` to be a score (between zero and one), where ``min_val`` 
is the best and ``max_val`` is whatever ``greater_score`` is. 
```

eteq:
This was discussed in a google hangout, but just to leave a note here on the topic: I'd suggest dropping the argument of whichever of ``greater_than``/``less_than`` has a default of ``0``.  The reasoning here is that for a function named e.g. ``min_best_rescale``, it won't ever make sense for values greater than max_val to be something other than ``0``.  In contrast,  ``less_than`` is still useful because sometimes you want values below that to be ``0`` but other times you might want it to be ``1``.

In general, though, I'm 👍  on having the two separate functions.  Much easier to understand.

kvyh:
I changed the docstring. I still need to add it into the constraints tutorial.

kvyh:
I changed the constraint tutorial to contain a non-boolean constraint that uses the `min_best_rescale`

astroplan/constraints.py
```diff
@@ -997,24 +998,90 @@ def observability_table(constraints, observer, targets, times=None,
...
+    Parameters
+    ----------
+    vals : array of values
+        the values that need to be rescaled to be between 0 and 1
+    min_val : value
+        worst acceptable value (rescales to 0)
+    max_val : value
+        best value cared about (rescales to 1)
+    less_than_min : 0 or 1
+        what is returned for ``vals`` below min_val. (in some cases
```
bmorris3: Would you put double ticks around min_val on this line?

astroplan/constraints.py
```diff
@@ -997,24 +998,90 @@ def observability_table(constraints, observer, targets, times=None,
...
+    Parameters
+    ----------
+    vals : array of values
+        the values that need to be rescaled to be between 0 and 1
+    min_val : value
+        worst acceptable value (rescales to 0)
+    max_val : value
+        best value cared about (rescales to 1)
+    greater_than_max : 0 or 1
+        what is returned for ``vals`` above max_val. (in some cases
```
bmorris3: Would you put double ticks around `max_val` here?

astroplan/constraints.py
```diff
@@ -997,24 +998,90 @@ def observability_table(constraints, observer, targets, times=None,
...
+    rescales an input array ``vals`` to be a score (between zero and one),
+    where the ``max_val`` goes to one, and the ``min_val`` goes to zero.
+
+    Parameters
+    ----------
+    vals : array of values
+        the values that need to be rescaled to be between 0 and 1
+    min_val : value
+        worst acceptable value (rescales to 0)
+    max_val : value
```
bmorris3: The type of `min_val` and `max_val` should be `float`, not value.

astroplan/constraints.py
```diff
@@ -997,24 +998,90 @@ def observability_table(constraints, observer, targets, times=None,
...
+    rescales an input array ``vals`` to be a score (between zero and one),
+    where the ``min_val`` goes to one, and the ``max_val`` goes to zero.
+
+    Parameters
+    ----------
+    vals : array of values
+        the values that need to be rescaled to be between 0 and 1
+    min_val : value
+        worst acceptable value (rescales to 0)
+    max_val : value
```
bmorris3: Same comment as [below](https://github.com/astropy/astroplan/pull/236/files#r75929188).

kvyh:
I changed the parameter types to "array-like" and "float" and ``ed the min_val and max_val.

bmorris3:
Thanks @kvyh!

Issue #212, opened 2016-08-04
Assumed TransitionBlock in PriorityScheduler


wtgee:
There are some issues with the `PriorityScheduler` in terms of an assumed TransitionBlock on lines [659](https://github.com/astropy/astroplan/blob/master/astroplan/scheduling.py#L659) and [677](https://github.com/astropy/astroplan/blob/master/astroplan/scheduling.py#L677). Since the `transitioner` attribute on the `PriorityScheduler` class is optional this can create problems. If `transitioner` is left as `None` then the callable on lines 658 and 676 fail, whereas if a default `transitioner` is created in the `PriorityScheduler` constructor, the `duration` attribute is not set. This is because on [776](https://github.com/astropy/astroplan/blob/master/astroplan/scheduling.py#L776) there is a check for the `components` of the `TransitionBlock` and if that doesn't exist `None` is returned.

If `None` is returned by the callable `Transitioner` this will create issues for the remainder of how `PriorityScheduler._make_schedule` works.

The solution will depend on the intended behavior, which I don't know, but it seems like `Transitioner.__call__` should return `self` instead of `None` on line 779 and that a `Transitioner` should also have a default `duration` (however that cannot be `0` or you will get division errors on lines 659 and 677)

wtgee:
Or just make a `TransitionBlock` a required param for `PriorityScheduler`

PR #240, opened 2016-08-20
Use an empty transitioner if None provided

kvyh:
This is  an initial attempt to address #212. This would be a temporary work-around (zero second long `TransitionBlocks` look strange on a schedule), but `PriorityScheduler` assumes and uses `TransitionBlocks` in a few places, so a true fix would likely require additions to `Schedule` as well as `PriorityScheduler`.


bmorris3:
I think the function should *require* a `Transitioner` as input, and raise an error if none is given. You can do this with the format: =
```python
if self.transitioner is None:
    raise ValueError("A Transitioner is required")
```

bmorris3:
👍

PR #269, opened 2016-12-13
Remove unneeded call to find Sun's position

migueldvb:
The `moon_altaz` method contains a call to find Sun's position that can be removed.


bmorris3:
Thanks @migueldvb, I'm restarting some of the failed tests which have to do with the IERS tables.

pllim:
Looks like that line is gone in `master` and this can be closed.

PR #279, opened 2016-12-30
Small documentation fixes that should clear up some test failures

wtgee:
Two tests are failing because of these sphinx documentation errors. This should clean up a lot of the failures on different PRs except for the tests using numpy prerelease and astropy dev.


eteq:
Thanks @wtgee !

PR #285, opened 2017-02-02
Fix broadcasting by using vector SkyCoords internally [WIP]

StuartLittlefair:
As described in #215, I think there's an argument for all internal calculations converting their arguments to ```Time``` and ```SkyCoord``` objects, which may be vector. This PR builds on the utility function in #215 and implements this functionality. It's a large change in the ```observer``` codebase, but outward facing functionality should be compatible with how it was before. However, there are so many unassociated test failures from other issues with astropy 1.3 it's hard to know, hence the WIP designation.

Once complete, this PR *could* fix remaining broadcasting issues with astropy 1.3 (e.g #282, #263, #272).

Some notes:

- This PR *requires* astropy 1.3 as a minimum version.
- ~~Although behaviour for scalar times and targets is unchanged, there is an improvement (IMO) in functionality for multiple times and targets. I briefly describe that here.~~

**This PR should give identical behaviour to the existing astroplan code, though the internals of the ```observer``` module have changed quite a lot. Please disregard the code samples below - I've left them in the comment for historical purposes**

First - for scalar objects:
```python
In [2]:  apo = Observer.at_site("APO")
In [3]:  times = Time(2457786.9048868627, format='jd')
In [4]:  m31 = coord.SkyCoord.from_name('M31')
In [5]:  apo.altaz(times, m31)
Out[5]: 
<SkyCoord (AltAz: obstime=2017-02-02 09:43:02.224949, location=(-1463969.3018517173, -5166673.342234327, 3434985.7120456537) m, pressure=0.0 hPa, temperature=0.0 deg_C, relative_humidity=0, obswl=1.0 micron): (az, alt) in deg
    ( 346.02166236, -13.98494769)>
```

However, if the times and targets are 1D and broadcastable, altaz and similar calcs will broadcast automatically:
```python
In [6]: ippeg = coord.SkyCoord.from_name('IP Peg')
In [7]: deneb = coord.SkyCoord.from_name('Deneb')
In [8]: times = times + np.linspace(-0.3, 0.3, 3)*u.d
In [9]: apo.altaz(times, [m31, ippeg, deneb])
Out[9]: 
<SkyCoord (AltAz: obstime=[datetime.datetime(2017, 2, 2, 2, 31, 2, 224949)
 datetime.datetime(2017, 2, 2, 9, 43, 2, 224949)
[...]

PR #286, opened 2017-02-04
Use numpy broadcasting for altaz and constraints

StuartLittlefair:
As discussed in #285 I think there is an advantage to relying on the broadcasting of times and targets in astropy, now that those objects are shape aware. This PR is basically to show how that would work, so the code maintainers can discuss if it's a good idea or not.

This PR passes all tests locally (some of which have been tweaked to accommodate changed behaviour - see below). To do this I have included changes in #281, #279, #273 as well.

The three potential gains are cleaner code, faster performance and greater flexibility. The downsides are that it requires astropy v1.3, and that it's a change which will probably break user code - see discussion below.

###### Cleaner Code

The philosophy is to always convert target arguments to ```SkyCoord``` objects. As a result, we no longer have to have two paths through the code for vector and non-vector targets. We can also get rid of constructions like:

```python
target_coos = [target.coord if hasattr(target, 'coord') else target
                         for target in targets]
target_altazs = [observer.altaz(times, coo) for coo in target_coos]
solar_separation = Angle([sunaltaz.separation(taa) for taa in target_altazs])
```
which becomes
```python
solar_separation = targets.separation(sunaltaz)
```

###### Faster performance

Performance is similar for single targets and single or a few times. However, when calculating the altaz or evaluating constraints over many targets and times it is much faster. For example evaluating the built-in constraints for 100 times and 2 targets has improved from 15s to 1.5s. Scheduling currently does every combination of target/time one-by-one so doesn't see any speed-up but the code could be adapted in a future PR.

###### Greater flexibility

This is also where the code-breaking behaviour arises. Previously, evaluating a constraint or calculating altaz for N targets and M times would always return an NxM array. This is no longer true. If N and M are equal, for example, the result will be a 1D array of size M. 

I'd argue this is an improvement - one doesn't always want to calculate every target at every time. To minimise the transition pain, both altaz and the constraint __call__ function have a ```grid``` argument, which restores the previous behaviour.


StuartLittlefair:
OK, I was going to ping people when this was passing tests, but I can't understand some of the failures. The ```TargetAlwaysUpWarning``` is fine on my system but fails on all travis tests. 

The older version of numpy also seem to fail because Time has no ```len``` attribute. Not sure what to do about that. 

So, for discussion, @eteq, @bmorris3, @bsipocz you might want to take a look at this PR sometime.

StuartLittlefair:
I have now rebased this so it is easier to see what has changed. I've also profiled the changes here against the current master which also uses ```SkyCoords``` internally. I evaluated each constraint for 100 times and 3 targets, and also for each combination of scalar time, vector target and vice versa. The results (in s) are shown below:

| Constraint      | Master           | This PR |
| ------------- | ------------- | ------------- |
| Altitude | 0.35 | 0.27 |
| Airmass | 0.06 | 0.09 |
| AtNight | 0.19 | 0.21 |
| SunSeparation | 1.07 | 0.53 |
| MoonSeparation | 29.46 | 1.09 |
| MoonIllumination | 0.76 | 0.59 |
| LocalTime | 0.02 | 0.09 |
| Time | 0.01 | 0.11 |

bmorris3:
@StuartLittlefair It looks like this PR is good to go! I just want to check before I click the green button that this one is ready to merge?

StuartLittlefair:
I'm happy with it. 

I'd be happier if @eteq took a look at it because it's a big change. Also, does it need a change log entry?

bmorris3:
Ok, then I'm going to do an intermediate release first, so that we can get v0.2.1 on pypi ASAP.

StuartLittlefair:
Good plan!

StuartLittlefair:
Hi @bmorris3. 

I'm thinking of finishing up this PR, but I can't make my mind up about the default behaviour. In the past astroplan has always done calculations for a grid of times and targets and as you know, this PR changes this default. It doesn't have to; we could make gridding the default again, and allow non gridding as an option. 

I'm going to finish up editing the docs and adding a changelog over the next week, but before I do that I wanted to check that you are happy with this decision?

StuartLittlefair:
p.s, this PR fixes #296

StuartLittlefair:
@bmorris3 you might want to take another look at this now. I've tidied it up, and the only test failure seems unrelated. In fact, I can't understand why that's suddenly appeared - perhaps you could restart that one failing test?

StuartLittlefair:
@bmorris3, @eteq - can I nudge you for a code review of this PR? I'm already starting to use it in our own telescope control/scheduling package due to the speedups involved in not gridding targets and times where not needed.

bmorris3:
@StuartLittlefair sorry for the long delay there! I just finished my candidacy exam and can take a look now. Reviewing!

docs/tutorials/constraints.rst
```diff
@@ -339,11 +310,15 @@ The resulting list of booleans indicates that the only target separated by
 5 and 30 degrees from Vega is Albireo. Following this pattern, you can design
 arbitrarily complex criteria for constraints.
 
-To see the (target x time) array for the non-boolean score::
+By default, calling a constraint will try to broadcast the time and target arrays
+against each other, and raise a `ValueError` if this is not possible. To see the
+(target x time) array for the constraint, there is an optional "grid_times_targets"
```
bmorris3: Small correction – would you please make `"grid_times_targets"` -> ``` `grid_times_targets` ```? That way it will be rendered nicely in the docs.

CHANGES.rst
```diff
@@ -3,6 +3,13 @@
 
 - No changes yet
```
bmorris3: We can remove this line with your addition below.

StuartLittlefair:
@bmorris3 congratulations!

I've made the minor changes requested

StuartLittlefair:
Closed and re-opened to start tests, which failed due to random lookup failure for APO site.

bmorris3:
Thanks for all of your hard work @StuartLittlefair!

PR #297, opened 2017-04-11
Constraints patch

lordaniket06:
@StuartLittlefair , @bmorris3 . This is in relation to the results I got from profiling the code. I have made some changes regarding that.
Adjusted the code slightly so that memory usage is minimized, the properties `_all_constraints` and `constraints` were being used repetitively . 
Also separately added the part which can append `AltitudeConstraints` in `b.constraints` if initially the constraints was not empty and there was no instance of `AltitudeConstraint` (default 0 deg) in it.

### DIFF
--- a/astroplan/observer.py
+++ b/astroplan/observer.py
@@ -15,7 +15,6 @@ from astropy.utils import isiterable
 from astropy.utils.compat.numpy import broadcast_to
 import astropy.units as u
 from astropy.time import Time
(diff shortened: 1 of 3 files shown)


lordaniket06:
This was supposed to be the patch I planned to create for GSoC-17 . However, due to some unavoidable reasons, I could not submit it before deadline.

PR #325, opened 2017-10-08
Implement moon_rise_time/moon_set_time

bmorris3:
This pull request implements two methods on the `Observer` object that previously raised a `NotImplementedError`. They are the `moon_rise_time` and `moon_set_time` methods, which follow the same call signature as the `sun_rise_time` and `target_rise_time` methods, just for the quickly moving moon. 

Before this PR, when calling `observer.sun_rise_time(time)`, the sun's sky coordinates were being calculated only once, at time `time`, and then the altitude of that _fixed_ coordinate was propagated forwards or backwards in time to approximate the Sun's motion throughout the day. Of course, the sun is slowly moving with respect to the celestial sphere throughout that 24 hour period, which can make a difference in rise/set times on the order of 3 minutes. 

The moon's apparent drift leads to a rise/set time offset of ~50 minutes, so I refactored the implementation of the rise/set/transit mechanics under-the-hood so that the rise/set/transit times of _moving_ objects (so far just the Sun and Moon) are treated properly. Here's a short outline of what "treated properly" means: 
* When `observer.sun_rise_time(time)` is called, the solar altitude is calculated over a grid of times spanning from `time` to `time + 24 hours + 3 minutes`, to account for the Sun potentially rising a bit later – this is meant to be similar to one sidereal day.
* When `observer.moon_rise_time(time)` is called, the lunar altitude is calculated over a grid of times spanning from `time` to `time + 24 hours + 60 minutes` to account for the Moon rising later. **Question:** I know that the sun rises ~50 min later each day at mid-latitudes. How much later could the moon rise at _any_ latitude?

* When `observer.target_rise_time(time, fixed_target)` is called, the old scheme is executed, where the target is fixed on the celestial sphere and only 24 hours need to be considered.

I've added some tests, and tweaked the tolerance in the tests against PyEphem to ensure that astroplan is getting better agreement with PyEphem than it used to. Before the sun rise/set times were consistent to within 8 min, but now they're all consistent to within 2 min.

***

**Opinions wanted:** In order to treat the sun and moon differently from `FixedTarget` objects, while still using the same rise/set API, I chose to introduce the class [`SpecialObjectFlag` and its subclasses `MoonFlag` and `SunFlag`](https://github.com/bmorris3/astroplan/blob/moon/astroplan/target.py#L266-L285), which get dropped into the `target` argument of [the workhorse function `observer._calc_riseset(time, target)`](https://github.com/bmorris3/astroplan/blob/moon/astroplan/observer.py#L696). These flags define approximate "drift times" for each object – 3 minutes, 60 minutes for Sun and Moon, respectively. 

You could imagine that every planet/minor planet/solar system object are "special" objects too. But since we don't yet have any concrete plans for implementing the `NonFixedTarget` class, I figured this is an acceptable drop-in solution to the sun/moon rise/set problem for now. Because the Sun and Moon _are_ special.

### DIFF
--- a/astroplan/observer.py
+++ b/astroplan/observer.py
@@ -18,7 +18,7 @@ import pytz
 # Package
 from .exceptions import TargetNeverUpWarning, TargetAlwaysUpWarning
 from .moon import moon_illumination, moon_phase_angle
-from .target import get_skycoord
+from .target import get_skycoord, SpecialObjectFlag, SunFlag, MoonFlag
 
 
 __all__ = ["Observer", "MAGIC_TIME"]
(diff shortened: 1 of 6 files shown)


bmorris3:
These tests will fail until astropy/astropy#6783 is merged.

bmorris3:
Build should pass, merging for speed!

PR #349, opened 2018-03-14
Support calling plot_airmass with an empty target list

lpsinger:
It will still set up the axes, add the brightness shading if requested, and return the axes.

This is helpful for my application because I want to start from an empty airmass chart and add my own plot elements, like this probabilistic airmass chart for a LIGO event.

airmass


bmorris3:
Thanks for the PR, and beautiful plot! The test failures here are unrelated to the PR, so I'm merging.

lpsinger:
When are you planning to do your next release?

bmorris3:
Soon, I suppose. Would you like one soon?

lpsinger:
It's not time-critical. I have a workaround for this change that I can use in the meantime (add a dummy target and then remove it).

PR #352, opened 2018-03-15
Fix doctest failures on new numpys

eteq:
There are two Travis tests failing that appear to be from the doctests.  The core problem appears to be that newer versions of numpy (I think 1.14 and later?) subtly changed the whitespace in how arrays get printed, which is causing doctest failures.  Adding the ``# doctest: +FLOAT_CMP`` magic here triggers a plugin that's in the helpers which changes the comparison from being a "is this string an exact match" to "are the values in this output numerically close to the expected values".  That also conveniently works around the whitespace problem.

cc @bmorris3


eteq:
(Note that, with this + #351 I *think* the tests should all be passing again)

PR #353, opened 2018-03-24
Add keyword in plot_airmass for Moon Airmass, brightness_shading multiple days

cagostino:
I couldn't find a way to do this with the existing astroplan system. Please correct me if I'm wrong and missed something, It seems like something people would want to be able to add on to their plot so I figured I should add it.
Also, I added in a quick fix to show brightness shading more than one day i.e. if I want to see it over a timeframe of several days, which it didn't do before.
Let me know what you all think.


bmorris3:
Hi @cagostino! I think the multiple-night shading is really useful, thanks for the PR! Would you mind making this PR focus on that aspect, and separating out the moon airmass part? I think some of the team will have opinions that the moon airmass should be left to the user to compute as with all other targets.

cagostino:
Will do, thanks for the suggestions!

cagostino:
Will do, thanks for the suggestions!

PR #354, opened 2018-03-26
Added in Brightness Shading for multiple day observations in plot_airmass

cagostino:

PR #355, opened 2018-03-26
Added in moon option for plot_airmass

cagostino:
I think this would be a useful feature to have in plot_airmass as it's easily calculable just with the information given and I think this would be a common enough calculation people have to do that it would be worth having to save a little bit of time.

### DIFF
--- a/astroplan/plots/time_dependent.py
+++ b/astroplan/plots/time_dependent.py
@@ -45,11 +45,10 @@ def _has_twin(ax):
             return True
     return False
 
-
 def plot_airmass(targets, observer, time, ax=None, style_kwargs=None,
                  style_sheet=None, brightness_shading=False,
                  altitude_yaxis=False, min_airmass=1.0, min_region=None,
-                 max_airmass=3.0, max_region=None):
+                 max_airmass=3.0, max_region=None, moon_airmass=False):
     r"""
     Plots airmass as a function of time for a given target.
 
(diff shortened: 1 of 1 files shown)


astroplan/plots/time_dependent.py
```diff
@@ -120,7 +118,8 @@ def plot_airmass(targets, observer, time, ax=None, style_kwargs=None,
     max_region : float
         If set, defines an interval between ``max_airmass`` and ``max_region``
         that will be shaded. Default is `None`.
-
+    moon_airmass : bool
```
wtgee: @cagostino it looks like you just forgot to include the `moon_airmass` parameter in the method declaration (lines 47-48).

I'd be curious about this PR. Can you make the fix above and post an example plot of what this looks like? Cheers!

cagostino: 3xmmhii0airmass
I've added an example. Here red is the corresponding airmass for the moon. I should be able to make these suggested changes as well as soon as my semester ends.

If you'll notice, the brightness shading here also covers multiple days, which is the subject of a separate pull request (#354) and is not currently implemented in the master branch iirc.

wtgee:
@cagostino any way you can clean this up and get submitted? If you don't have time let me know. I'm at a hack-week this week and might try to clean up some of the airmass plots so would just incorporate this. Same with #354

cagostino:
@wtgee yeah probably. I was not exactly sure what to do with #354 since its checks were no longer working and weren't showing any details as to why. Would be happy to get that one through though.

cagostino:
I fixed what you had suggested previously, let me know if there is anything else I can do to move this along @wtgee

wtgee:
> I fixed what you had suggested previously, let me know if there is anything else I can do to move this along @wtgee

Thanks! There's a conflict in one of the files still, if you could clean that up @bmorris3 can then take a look for merging.

PR #369, opened 2018-05-14
add tests for priority scheduler with short eclipses

StuartLittlefair:
As requested in #321 and finally delivered, a test which checks if the priority scheduler can implement short OBs which are only just shorter than the validity of their constraints


StuartLittlefair:
@bmorris3 failures here look related to download of IERS machinery, not this PR

bmorris3:
Perfect, thanks @StuartLittlefair!

PR #405, opened 2019-02-24
Access correct variable for longitude in EarthLocation

cosmicpudding:
In observer.py (and possibly elsewhere), it points to `self.location.lon` which is now incompatible with astropy, and should be changed to `self.location.longitude` instead. Tested this with the `local_sidereal_time` function which had given an error (`AttributeError: 'EarthLocation' object has no 'lon' member`). Changing both instances to `self.location.longitude` fixed the issue.


cosmicpudding:
Oops, I didn't realise my astropy was still 1.3.2! Never mind :p

Issue #423, opened 2019-10-16
Quicker versions of target_rise_time & target_set_time


iwellaway:
We are using the Astroplan Observer object extensively in the development of our robotic scheduler for the INT. We need to run a 10 year simulation so any way we can speed up the calculations would be a great help.

Is it possible to produce a faster, albeit probably slightly less accurate, version of the Observer object methods: target_rise_time, target_set_time and target_meridian_transit_time?

bmorris3:
Thanks for the note @iwellaway. I've implemented a faster "trigonometric solver" in pull request #327, which is currently under review. Would you please leave comments on that pull request if you have any feedback on the implementation? I'd like to make sure that it will work well for folks like you!

iwellaway:
Thanks Brett, I’ll take a look.

eteq:
@iwellaway - is there a specific test case (i.e. code snippet) you can provide that might let us do some profiling on your use case?  I have a suspicion that there are areas of these functions that could be *much* better optimized without going the "lower-accuracy" direction.  As I said in #327 it's not really clear to me that the coordinates machinery itself is the "hot spot" in these functions, especially given that there have been several major performance enhancements since #327 was first created.

bmorris3:
To begin this discussion, I've timed the `Observer.target_rise_time` function with and without the trig approximation, [see gist here](https://nbviewer.jupyter.org/gist/bmorris3/cab7eac37fc1afa5cdb96f919fe4a3d8), which shows a 2.4x speed-up (for a single target). The only machinery change between these functions is the coordinate engine swap.

bmorris3:
For the sake of transparency in this process, I'm going to briefly outline what astroplan is currently doing to compute rise/set times, mostly in the [`Observer._calc_riseset`](https://github.com/astropy/astroplan/blob/master/astroplan/observer.py#L695) function: 

1) Generate a grid of times with 5 min resolution over the 24 hours before or after the specified reference time [[code](https://github.com/astropy/astroplan/blob/master/astroplan/observer.py#L750)] (fast!)
2) Compute the alt/az of the target at each time on the 24 hour grid [[code](https://github.com/astropy/astroplan/blob/master/astroplan/observer.py#L759)] (slower)
3) Identify two grid-points where the target rises/sets above/below the user-defined horizon altitude, interpolate between the two to find the rise/set time [[code](https://github.com/astropy/astroplan/blob/master/astroplan/observer.py#L520)] (fast!)

[...]

PR #424, opened 2019-10-17
Adding `n_grid_points` parameter to rise/set/transit functions

bmorris3:
**tl;dr**: you can speed up rise/set times by an order of magnitude in exchange for a few minutes of rise/set time precision, without changing anything fundamental about how astroplan/astropy compute rise/set times.

This PR implements the changes I suggested in #423, which I summarize briefly below. There are also several PEP8 compliance fixes throughout the docstrings on the Observer object. If merged, this PR supersedes #327.

The primary goal of this PR is to expose the `N` parameter, which has been lurking within the guts of the Observer object since its inception. Rise/set/transit times are computed using a grid search, which scans for horizon crossings over `N=150` times spread over 24 hours relative to the reference time supplied by the user. By throttling that parameter up or down, the user can simultaneously change the precision and speed of rise/set/transit calculations. `N~100` gives precision better than 1 min from the true rise/set/transit time, `N~10` gives precision better than 5 minutes and is about 10x faster. This fix is designed to address the common complaint that rise/set times are slow (#402, #423).

The default behavior of the functions remains unchanged. When not supplied by the user, `N=150` by default for all functions. Thus none of the tests need to change, and introducing this new optional kwarg won't change people's results in legacy code.

Still to do: 
* ~add narrative docs explaining the update~
* include plots in docs comparing precision at different `N`

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,9 @@
   Galactic point sources is an issue. [#413]
 
 
+- Add `N` keyword argument to rise/set/transit functions which allows users to
+  trade off precision for speed. [#424]
+
 0.5 (2019-07-08)
 ----------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,8 +15,8 @@
   Galactic point sources is an issue. [#413]
 
 
-- Add `N` keyword argument to rise/set/transit functions which allows users to
-  trade off precision for speed. [#424]
+- Add ``N`` keyword argument to rise/set/transit functions which allows users
+  to trade off precision for speed. [#424]
 
 0.5 (2019-07-08)
 ----------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,8 +15,8 @@
   Galactic point sources is an issue. [#413]
 
 
-- Add ``N`` keyword argument to rise/set/transit functions which allows users
-  to trade off precision for speed. [#424]
+- Add ``n_grid_points`` keyword argument to rise/set/transit functions which
+  allows usersto trade off precision for speed. [#424]
 
 0.5 (2019-07-08)
 ----------------
--- a/astroplan/observer.py
+++ b/astroplan/observer.py
@@ -732,7 +732,7 @@ class Observer(object):
         return alt
 
     def _calc_riseset(self, time, target, prev_next, rise_set, horizon,
-                      N=150, grid_times_targets=False):
+                      n_grid_points=150, grid_times_targets=False):
         """
         Time at next rise/set of ``target``.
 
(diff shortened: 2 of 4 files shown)


bmorris3:
Tagging some users who might have opinions about this implementation: @wtgee, @StuartLittlefair, @iwellaway, @eteq.

StuartLittlefair:
@bmorris3 I like this approach a lot, and it does indeed make things a lot faster, and is simple to understand for the end user. 

For a single time/target I find the speedup between ```N=150``` and ```N=10``` is only x2, but this obviously improves with grids of times and targets, since ```N``` broadcasts against those grids, so the speedup will be more impressive if doing large numbers of calculations.

One thing I've often wondered is why we couldn't use a 2-pass scenario, where we use a small ```N``` to get an initial rough estimate of the event time, and then re-grid 10--15 mins either side of this estimate (using the same ```N```) to refine it, i.e use a few lines like the ones below at the end of these functions...

```python
    t = self._two_point_interp(jd1, jd2, al1, al2, horizon=horizon)
    # normally we'd return here, but let's use a finer grid, 
    # 15 mins either side of rough time
    times = _generate_24hr_grid(t, -0.01, 0.01, N)
    altaz = self.altaz(times, target, grid_times_targets=grid_times_targets)
    altitudes = altaz.alt
    al1, al2, jd1, jd2 = self._horiz_cross(times, altitudes, rise_set, horizon)
    return self._two_point_interp(jd1, jd2, al1, al2, horizon=horizon)
```

This is roughly the same speed for a single time/target for ```N=10``` vs ```N=150``` (as the two calculations offset the x2 speedup), but is still 5x faster for a grid of 300 times and agrees with the ```N=150``` case to within 1 second.

Or - giving the option to the user, you could switch a two-pass behaviour on or off using an argument, and allow the user to tune ```N``` and ```two_pass``` for their own accuracy requirements...

bmorris3:
@StuartLittlefair I seem to remember that it was originally our intention to use a two-pass solution, but looking back through the PR history, I can't find evidence that we ever actually did a two-pass solution. 

I think something @eteq and I discussed casually a few times is that atmospheric pressure and humidity along the line of sight can probably change rise/set times by a few minutes here or there, so we wondered whether it would be physically meaningful to provide rise/set times that are accurate to the minute anyways. I suppose I'm making the arguments that (1) perhaps the default `N` could be smaller, and (2) a two-pass solution is probably overkill.

StuartLittlefair:
I don't really disagree with any of that. But I would make the following counter-arguments:

1) for *any* given accuracy requirement, a two-pass solution is quicker than just using a finer grid over 24 hour

2) It seems like a simple change to the code, I think the lines pasted above work fine

bmorris3:
I've updated the PR to include narrative docs, which render like so: 

Screenshot 2019-10-24 at 11 32 56

StuartLittlefair:
Nice. I don't think my comments should hold up merging this PR BTW - I could always submit a separate PR for a two-pass approach at a later date.

tepickering:
this looks good and definitely helps in cases where precision isn't too important (like twilight times). i think there's still more to be gained from either a two-pass approach as suggested by @StuartLittlefair or an iterative approach like `skyfield` uses (see `find_discrete` in https://github.com/skyfielders/python-skyfield/blob/master/skyfield/almanac.py). those can be subjects for future PRs, though, and shouldn't hold this up...

PR #434, opened 2019-11-16
Adding sunrise/set precision discussion in FAQ, tests against MMTO

### LINKED ISSUE #409: Observer.sun_set_time not working properly
I'm trying to use Observer to determine the time of the sunset for a given day, the problem is that here on the site we have an official sunset time that does not match the one that astroplan gives. After some observation the official time seems to be the correct, the times differ for about 10 minutes.

What worries me the most is that while debugging I realized that every time the code was executed the sunset time returned was different.

Minimal example:
```python
from astroplan import Observer
from astropy.time import Time
from astropy.units import deg, m
from astropy.coordinates import EarthLocation
from datetime import datetime

obs = Observer(latitude=-24.622997508*deg, longitude=-70.40249839*deg, elevation=2635*m, name='pl', timezone='UTC')
print(obs.sun_set_time(Time(datetime.now()), which='next').to_datetime())
```

bmorris3:
Fixes #409, with some great help from @tepickering.

### DIFF
--- a/astroplan/tests/test_observer.py
+++ b/astroplan/tests/test_observer.py
@@ -1297,3 +1297,30 @@ def test_moon_rise_set():
             datetime.timedelta(minutes=threshold_minutes))
     assert (abs(pyephem_prev_set - astroplan_prev_set.datetime) <
             datetime.timedelta(minutes=threshold_minutes))
+
+usno_sunsets = [
+    Time('2019-01-01 00:31'),
+    Time('2019-02-01 00:58'),
+    Time('2019-03-01 01:21'),
+    Time('2019-04-01 01:43'),
+    Time('2019-05-01 02:03'),
+    Time('2019-06-01 02:24'),
+]
+
+@pytest.mark.parametrize('usno_sunset', usno_sunsets)
+def test_sun_set_vs_usno_almanac(usno_sunset):
+    """
+    Validates issue: https://github.com/astropy/astroplan/issues/409
+
+    USNO times to the nearest minute from the MMTO Almanac:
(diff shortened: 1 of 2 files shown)


bmorris3:
The test failures are a result of astropy/astroquery#1586.

tepickering:
looks good. i should actually go and audit the old fortran code we use to generate the MMTO almanac   😜

might also be worth adding some tests against `skyfield` at different locations. and against actual USNO calculations once they're back online...

tepickering:
for example, here's a comment i found in our code:
```
C This program is an ideal example of how NOT to write software.
C or it could be viewed as an enbalmed example of how people
C used to write software before they knew any better.
```
:rofl: :rofl: :rofl:

astroplan/tests/test_observer.py
```diff
@@ -1297,3 +1297,30 @@ def test_moon_rise_set():
...
+    Time('2019-01-01 00:31'),
+    Time('2019-02-01 00:58'),
+    Time('2019-03-01 01:21'),
+    Time('2019-04-01 01:43'),
+    Time('2019-05-01 02:03'),
+    Time('2019-06-01 02:24'),
+]
+
+@pytest.mark.parametrize('usno_sunset', usno_sunsets)
+def test_sun_set_vs_usno_almanac(usno_sunset):
```
tepickering: i would specify that this is the MMTO almanac. i don't know if our code is actually derived from any USNO code. however, the parameters used in the code are those published by the USNO in their Astronomical Almanac (currently offline: http://asa.usno.navy.mil/) and the times have been verified in the past against the USNO online calculator.

PR #437, opened 2019-11-19
Different strategy for measuring the last date in the IERS table

### LINKED ISSUE #436: Check for age of IERS table fails with untrapped PolPMFlag lookup
I'm having a lot of trouble getting astroplan to run after changing the location of the IERS table download to ftp://cddis.gsfc.nasa.gov/pub/products/iers/finals2000A.all . An example error message is shown here:
`Traceback (most recent call last):
  File "mmtscheduler.py", line 210, in <module>
    import astroplan

  File "/usr/local/lib/python3.6/dist-packages/astroplan/__init__.py", line 32, in <module>
    get_IERS_A_or_workaround()

  File "/usr/local/lib/python3.6/dist-packages/astroplan/utils.py", line 57, in get_IERS_A_or_workaround
    iers.IERS.iers_table = _get_IERS_A_table()

  File "/usr/local/lib/python3.6/dist-packages/astroplan/utils.py", line 93, in _get_IERS_A_table
    index_of_last_observation = ''.join(table['PolPMFlag_A']).index('IP')

  File "/usr/local/lib/python3.6/dist-packages/astropy/table/table.py", line 1346, in __getitem__
    return self.columns[item]

  File "/usr/local/lib/python3.6/dist-packages/astropy/table/table.py", line 190, in __getitem__
    return OrderedDict.__getitem__(self, item)
KeyError: 'PolPMFlag_A'
`

Basically, in utils.py, _get_IERS_A_table() is trying to use the PolPMFlag_A key to figure out how old the table is, the table does not have this key, and an unrecoverable error is thrown. I am not sure if the missing key is a difference between the IERS table from ftp.cddis.nasa.gov and the previous file or something else. However, the relevant code (currently lines 94-98 of utils.py) should be wrapped in a try / except or something similar to catch the error.

bmorris3:
[Seeking advice from: @mhvk @eteq @pllim]

Is this PR sufficient for measuring the age of an IERS table? In #436, we're getting an error when we download an old version of the IERS A table from `ftp://cddis.gsfc.nasa.gov/pub/products/iers/finals2000A.all` which appears to _not_ contain the `PolPMFlag_A` column. 

cc @tepickering and @bjweiner.

Fixes #436.

### DIFF
--- a/astroplan/utils.py
+++ b/astroplan/utils.py
@@ -92,9 +92,16 @@ def _get_IERS_A_table(warn_update=14*u.day):
     if IERS_A_in_cache():
         table = iers.IERS_Auto.open()
         # Use polar motion flag to identify last observation before predictions
-        index_of_last_observation = ''.join(table['PolPMFlag_A']).index('IP')
-        time_of_last_observation = Time(table['MJD'][index_of_last_observation],
-                                        format='mjd')
+        if 'PolPMFlag_A' in table.colnames:
+            index_of_last_observation = ''.join(table['PolPMFlag_A']).index('IP')
+            time_of_last_observation = Time(table['MJD'][index_of_last_observation],
+                                            format='mjd')
+            
+        # If time of last observation is not available, set it equal to the
+        # final prediction in the table:
+        else:
+            time_of_last_observation = Time(table['MJD'].max(),
+                                            format='mjd')
         time_since_last_update = Time.now() - time_of_last_observation
 
         # If the IERS bulletin is more than `warn_update` days old, warn user
--- a/astroplan/utils.py
+++ b/astroplan/utils.py
@@ -96,7 +96,7 @@ def _get_IERS_A_table(warn_update=14*u.day):
             index_of_last_observation = ''.join(table['PolPMFlag_A']).index('IP')
             time_of_last_observation = Time(table['MJD'][index_of_last_observation],
                                             format='mjd')
-            
+
         # If time of last observation is not available, set it equal to the
         # final prediction in the table:
         else:


mhvk:
I'm very much hoping part of the problem will be fixed by https://github.com/astropy/astropy/pull/9623 - at least if this is an astropy 4.0 only problem...

astroplan/utils.py
```diff
@@ -92,9 +92,7 @@ def _get_IERS_A_table(warn_update=14*u.day):
     if IERS_A_in_cache():
         table = iers.IERS_Auto.open()
         # Use polar motion flag to identify last observation before predictions
-        index_of_last_observation = ''.join(table['PolPMFlag_A']).index('IP')
-        time_of_last_observation = Time(table['MJD'][index_of_last_observation],
-                                        format='mjd')
+        time_of_last_observation = Time(table['MJD'].max(), format='mjd')
```
tepickering: for an IERS_A table, the original logic gives you the time of the last _observation_ while the new logic gives you the end of the table which is a _prediction_. you want to check for the `PolPMFlag_A` column and, if it exists, use the original logic while falling back to `table['MJD'].max()` if it doesn't.

astroplan/utils.py
```diff
@@ -92,9 +92,16 @@ def _get_IERS_A_table(warn_update=14*u.day):
...
-        index_of_last_observation = ''.join(table['PolPMFlag_A']).index('IP')
-        time_of_last_observation = Time(table['MJD'][index_of_last_observation],
-                                        format='mjd')
+        if 'PolPMFlag_A' in table.colnames:
+            index_of_last_observation = ''.join(table['PolPMFlag_A']).index('IP')
+            time_of_last_observation = Time(table['MJD'][index_of_last_observation],
+                                            format='mjd')
+            
+        # If time of last observation is not available, set it equal to the
+        # final prediction in the table:
```
tepickering: might be good to add a warning here stating that IERS_A is not available and older data is being used instead. maybe include the date of the last observation in the warning message.

bmorris3: In the interest of preventing more IERS warnings from popping up, which are already a significant source of headaches for people, I think I might let this one go by silently...? I want to move as much of the IERS responsibilities into astropy as possible. Does that sound ok?

bmorris3: If you have strong feelings, please open a new issue, I'm merging this one for now.

tepickering: this is great, thanks! i double-checked and astropy should indeed be throwing its own warning in this case so no need for redundancy.

PR #445, opened 2019-12-22
Avoid traceback when source is always up

keflavich:
This PR fixes an issue where using 'nearest' will result in a traceback when requesting rise/set times for an always-up source.


keflavich:
Includes #449

bmorris3:
Thanks @keflavich!

PR #446, opened 2019-12-25
Timezone and shading

keflavich:
This PR adds support for timezone-specific airmass plots, and in the process corrects an error in which nighttime would be incorrectly unshaded if it is not centered in the plot.

Example:
image
image

Calls look like this:
```
plot_airmass(pcyg_target, CTO, now.to_datetime(timezone=CTO.timezone), use_local_tz=True, brightness_shading=True)
```
`now` is an astropy time object, `CTO` is an observatory (UF Campus Teaching Observatory).

This PR is built on https://github.com/astropy/astroplan/pull/445 but does not depend on it; it can be refactored to be separate if needed.

### DIFF
--- a/astroplan/observer.py
+++ b/astroplan/observer.py
@@ -924,13 +924,20 @@ class Observer(object):
                 return previous_event
 
         if which == 'nearest':
-            if not (np.isfinite(previous_event.value) and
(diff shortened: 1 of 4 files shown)


keflavich:
Includes both #449 and #445

Issue #468, opened 2020-08-25
plot_sky error


Lagrang3:
Every call to the function `astroplan.plots.plot_sky` raises an exception.
Even with the tutorial [example](https://astroplan.readthedocs.io/en/latest/tutorials/plots-10.py).

Here a minimal code that triggers the error:
```
#!/usr/bin/env python3
import matplotlib.pyplot as plt
import astropy.units as u
import astroplan, astroplan.plots, astropy.time

apo = astroplan.Observer.at_site('APO')
vega = astroplan.FixedTarget.from_name('Vega')
observe_time = astropy.time.Time(['2000-03-15 15:30:00'])

astroplan.plots.plot_sky(vega,apo, observe_time)
plt.show()
```

Expected behaviour: a nice polar plot with the position of Vega.
Actual outcome: 
```
Traceback (most recent call last):
  File "./plots-10.py", line 12, in <module>
    astroplan.plots.plot_sky(vega,apo, observe_time)
  File "/opt/anaconda3/envs/astroplan/lib/python3.8/site-packages/astropy/units/decorators.py", line 234, in wrapper
    return_ = wrapped_function(*func_args, **func_kwargs)
  File "/opt/anaconda3/envs/astroplan/lib/python3.8/site-packages/astroplan/plots/sky.py", line 216, in plot_sky
    ax.set_thetagrids(range(0, 360, 45), theta_labels)
  File "/opt/anaconda3/envs/astroplan/lib/python3.8/site-packages/matplotlib/projections/polar.py", line 1346, in set_thetagrids
    self.set_xticklabels(labels)
  File "/opt/anaconda3/envs/astroplan/lib/python3.8/site-packages/matplotlib/axes/_base.py", line 63, in wrapper
    return get_method(self)(*args, **kwargs)
  File "/opt/anaconda3/envs/astroplan/lib/python3.8/site-packages/matplotlib/cbook/deprecation.py", line 451, in wrapper
    return func(*args, **kwargs)
  File "/opt/anaconda3/envs/astroplan/lib/python3.8/site-packages/matplotlib/axis.py", line 1793, in _set_ticklabels
    return self.set_ticklabels(labels, minor=minor, **kwargs)
  File "/opt/anaconda3/envs/astroplan/lib/python3.8/site-packages/matplotlib/axis.py", line 1714, in set_ticklabels
    raise ValueError(
ValueError: The number of FixedLocator locations (8), usually from a call to set_ticks, does not match the number of ticklabels (7).
[...]

PR #469, opened 2020-08-25
Fix to issue #468

Lagrang3:
The number of theta_labels (7) didn't match
the number of ticks.

### DIFF
--- a/astroplan/plots/sky.py
+++ b/astroplan/plots/sky.py
@@ -210,7 +210,7 @@ def plot_sky(target, observer, time, ax=None, style_kwargs=None,
             theta_labels.append('W' + '\n' + str(label_angle) + degree_sign)
         else:
             theta_labels.append(str(label_angle) + degree_sign)
-
+    theta_labels.append('')
     # Set ticks and labels.
     ax.set_rgrids(range(1, 106, 15), r_labels, angle=-45)
     ax.set_thetagrids(range(0, 360, 45), theta_labels)


wtgee:
Hi @Lagrang3, thanks for submitting this. I was just looking over #468 right now and while I can easily reproduce the bug, I don't think just appending a label here is necessarily the best solution until we figure out what has changed and why it was working but suddenly is not.

It seems to be some problem within the appending labels before the error thought not sure what yet.

https://github.com/astropy/astroplan/blob/001a1e123500d96979287db3a817d891cb87bd11/astroplan/plots/sky.py#L198-L212

wtgee:
@larrybradley If you append something besides a blank label, where is it appearing on the plot?

Lagrang3:
The blank label appears exactly where the 'Alt' label is. Therefore if you change 
`for chunk in range(0, 7):` for `for chunk in range(0, 8): ` you will get `315` on top of the `0 Alt.` which is ugly to see.

Lagrang3:
There you go
sky

keflavich:
Can this fix be merged as a workaround?  The "keep mpl<3.3" approach doesn't work right now, at least.

Also, is there a reasonable workaround that can be implemented without modifying the astroplan source?  I need to teach with this this semester, and my students are not going to have an easy time managing package versions.

wtgee:
@keflavich thanks for the prompt. Ideally we just get this merged and get a new release out, but that's going to need @bmorris3. Ideally we get this in and then can update #493.

The mpl<3.3 is currently annoying, that's for sure, but it should be working correctly.  Can you explain what doesn't work about that approach right now?

keflavich:
I already have mpl 3.3 installed, and I anticipate that's true for many of my students; _maybe_ it will just work if they `pip install astroplan/master`... should that work?  I don't know.

Also, is the mpl<3.3 fix already merged into master, or is it just on your branch?  I was a touch confused reading through the PR history.

wtgee:
> I already have mpl 3.3 installed, and I anticipate that's true for many of my students; 

Got it, thanks. I thought you meant you installed `<3.3` and it didn't work.

> _maybe_ it will just work if they `pip install astroplan/master`... should that work? I don't know.

It's mostly that one of the plotting tests was failing, but most functionality should be okay. Do a `pip install --no-deps astroplan` and it won't try to replace your existing matplotlib.  You would then need to make sure you have all the dependencies.

> 
> Also, is the mpl<3.3 fix already merged into master, or is it just on your branch? I was a touch confused reading through the PR history.

Yes, it's in the `setup.py` in `master`:

https://github.com/astropy/astroplan/blob/8f66690945e9acd1e372127097ee3fef90ceab05/setup.py#L107

keflavich:
Thanks.  Just on the second point - the reason I posted here is that the functionality my students will need is one of the failures.  I have matplotlib=3.3.2 and astroplan=0.8.dev1248 installed, and that's probably what my students have installed too.  `pip install`ing astroplan successfully installed astroplan despite the `plotting` extras requirement...  so I'm still a bit screwed here.  I don't know what the `extras` setup argument does, actually - is there a pip argument I can use to specify that I need the plotting extras?

keflavich:
I've reviewed this further, and I'm :+1: on this change - it is correct, and the old way was incorrect and relied on a bug in matplotlib.

The matplotlib change was made here:
https://github.com/matplotlib/matplotlib/commit/dbe6f358d347108e0ef7c31b77a7ce0c7ec39225

`zip` truncates to the shortest iterable.  `matplotlib`'s `set_ticklabels` zipped the tick location and the labels.  Prior to 3.3.2, it did this silently.  Now, if they don't match length, it raises an issue.

bmorris3:
Thanks to each of you!

Issue #466, opened 2020-08-21
LocalTimeConstraint does not consider local time-zone


Lagrang3:
`AltitudeConstraint` does not consider DST (daylight saving time).
For instance consider the following code:

```
#!/usr/bin/env python3

import astroplan, astropy.time
import astropy.units as u
import pytz,datetime

polse=astroplan.Observer(
    latitude=
        46+(28+9.1/60)/60,
    longitude=
        13+(0+42.6/60)/60,
    elevation=
        750*u.meter,
    timezone=
        pytz.timezone('Europe/Rome'))
        
target = astroplan.FixedTarget.from_name('Aldebaran')
constraint = astroplan.AltitudeConstraint(min=20*u.deg)
    
print(constraint(polse,target,times=astropy.time.Time('2020-04-10 21:00',
        location=polse.location)))

```

It determines whether Aldebaran is visible above 20 deg. on the evening of 2020-04-10.
In Stellarium at that time the star is above 25 deg. and at 22:00 the star is below the 20 deg.
However the code returns `False` after 20:00. When it should be `True` until 21:00.

I am running anaconda on Linux:
```
astroplan                 0.6             
astropy                   4.0.1.post1 
```

PR #470, opened 2020-08-25
LocalTimeConstraint fixed timezone (#466)

Lagrang3:
Constraint objects are fed an 'astropy.time.Time' variable.
LocalTimeConstraint must transform that time according
to the local time-zone.


bmorris3:
Hi @Lagrang3!

Thanks for this update. Would you mind entering a changelog entry, and a test that demonstrates that this local time transformation is working as expected?

Lagrang3:
@bmorris3: Hi. I am glad someone replied at last.
> Would you mind entering a changelog entry, and a test
Yes I will.

PR #474, opened 2020-09-27
Fix incompatibility w/astropy-dev

keflavich:
This change enables different imports.

It was incorrect to use a private function in the first place, but that's
partly on astropy-core for making this code private.
https://github.com/astropy/astropy/pull/10775 solves the issue moving forward.


keflavich:
This solution is not acceptable.  #471 is the right solution.

Issue #476, opened 2020-10-07
download_IERS_A() time out


jonaszubindu:
Hi,

since about yesterday the function download_IERS_A() almost always (works maybe 1 in 50 times) throws a time out error and is unable to download the IERS tables. Tested it in different networks and from different platforms. astropy and astroplan are up to date.

martinjohndyer:
I can confirm this. 

<details>
<summary>Traceback, on the off-chance it's any use...</summary>

```python
In [1]: from astroplan import download_IERS_A                                                                         

In [2]: download_IERS_A()                                                                                             
Downloading ftp://cddis.gsfc.nasa.gov/pub/products/iers/finals2000A.all
|===============>----------------------------------------------------------------| 655k/3.3M (19.51%) ETA     2s
---------------------------------------------------------------------------
timeout                                   Traceback (most recent call last)
<ipython-input-4-3953bedb3d4d> in <module>
----> 1 download_IERS_A()

~/.local/lib/python3.6/site-packages/astroplan/utils.py in download_IERS_A(show_progress)
    138         try:
    139             local_iers_a_path = download_file(url, cache=True,
--> 140                                               show_progress=show_progress)
    141         except urllib.error.URLError:
    142             if i == len(urls) - 1:

~/.local/lib/python3.6/site-packages/astropy/utils/data.py in download_file(remote_url, cache, show_progress, timeout, sources, pkgname, http_headers)
   1173                 f"included as a valid source.")
   1174         elif len(sources) == 1:
-> 1175             raise errors[sources[0]]
   1176         else:
   1177             raise urllib.error.URLError(

~/.local/lib/python3.6/site-packages/astropy/utils/data.py in download_file(remote_url, cache, show_progress, timeout, sources, pkgname, http_headers)
   1145                     remote_url=remote_url,
   1146                     pkgname=pkgname,
-> 1147                     http_headers=http_headers)
   1148             # Success!
   1149             break

~/.local/lib/python3.6/site-packages/astropy/utils/data.py in _download_file_from_source(source_url, show_progress, timeout, remote_url, cache, pkgname, http_headers)
   1013                         bytes_read += len(block)
   1014                         p.update(bytes_read)
-> 1015                         block = remote.read(conf.download_block_size)
   1016                         if size is not None and bytes_read > size:
   1017                             raise urllib.error.URLError(

/usr/lib/python3.6/tempfile.py in func_wrapper(*args, **kwargs)
    622             @_functools.wraps(func)
[...]

PR #477, opened 2020-10-07
Automatically use backup IERS URL if first one fails to download

martinjohndyer:
When looking into what could be causing #476 I saw that the `timeout` error was not being caught within `astroplan.utils.download_IERS_A()`. This results in the function failing if the primary URL times out, without trying the backup.

One solution might be to catch `timeout` in addition to `urllib.error.URLError`, but when looking at the equivilent function in [`astropy.utils.iers.IERS_Auto()`](https://github.com/astropy/astropy/blob/625c6e6257af40e4b5c1070c9b3f707e7b0953dd/astropy/utils/iers/iers.py#L670-L680) it seems you can just pass a list of source URLs to `astropy.utils.data.download_file()`, which will run through the list if the first one fails.

So this PR makes that minor change to mean `download_IERS_A()` mirrors the code in `IERS_Auto()`. Letting `download_file()` catch the errors also means that there's no longer any reason the enumerate through the list of URLs here, making things simpler here too.

### DIFF
--- a/astroplan/utils.py
+++ b/astroplan/utils.py
@@ -134,13 +134,8 @@ def download_IERS_A(show_progress=True):
         for url in urls:
             clear_download_cache(url)
 
-    for i, url in enumerate(urls):
-        try:
-            local_iers_a_path = download_file(url, cache=True,
-                                              show_progress=show_progress)
-        except urllib.error.URLError:
-            if i == len(urls) - 1:
-                raise
+    local_iers_a_path = download_file(urls[0], sources=urls, cache=True,
+                                      show_progress=show_progress)
 
     # Undo monkey patch set up by get_IERS_A_or_workaround
     iers.IERS.iers_table = iers.IERS_A.open(local_iers_a_path)


wtgee:
Hi @martinjohndyer , thanks for contributing this and #476. The IERS table has a long history of being painful.  The patch you've submitted here somewhat reverses what was done in #418, which was an attempt to compensate for invalid urls.

There are also quite a few other Issues and PRs that were related to IERS items. I think some of the underlying astropy mechanism has also changed (or is about to change).

I'll try to look at this later but maybe @bmorris3 could comment on if there is a master plan for IERS within astroplan.

martinjohndyer:
@wtgee I wouldn't necessarily say this reverses any of #418 , the core principle of using the URLs from the astropy config is still respected. This just uses the astropy function to select the mirror if the primary URL fails, rather than the code that @lpsinger added which doesn't catch timeout errors we're now seeing.

If you really want a minimum change then all you need to do to fix #476 is just catch timeout errors explicitly, replace:
```python
    for i, url in enumerate(urls):
        try:
            local_iers_a_path = download_file(url, cache=True,
                                              show_progress=show_progress)
        except urllib.error.URLError:             # <---- Doesn't catch timeout
            if i == len(urls) - 1:
                raise
```
with
```python
    for i, url in enumerate(urls):
        try:
            local_iers_a_path = download_file(url, cache=True,
                                              show_progress=show_progress)
        except (urllib.error.URLError, timeout):  # <---- Now it does
            if i == len(urls) - 1:
                raise
```
or even just a general `except Exception` would do the job, I don't know if there is a paticular reason only `URLError`s are caught.

I'm well aware dealing with the IERS tables in astroplan have been an ongoing issue (e.g. [1](https://github.com/astropy/astroplan/issues/356), [2](https://github.com/astropy/astroplan/issues/192#issuecomment-448707531)). I'd certainly welcome any more involved changes, ideally using as much of the astropy code as possible (which is what I suggested doing in this PR), but hopefully this is a fairly small change that would fix an obvious shortcoming with the existing function.

wtgee:
Hi @martinjohndyer, thanks, you are right that it isn't a reversal of the work done, I guess I was not looking at it quite closely enough.

However, in addition to this issue there is also the issue that `IERS_A_in_cache` is relying on the private `astropy` method `_get_download_cache_locs` (as reported in #471).  Since `IERS_A_in_cache` is called both before and as part of `download_IERS_A`, this is always being called and the above changes will fix one problem, but not the other.

Really `astroplan` shouldn't be doing any of this since it just duplicates what `astropy` does.  `astropy.utils.iers.IERS_Auto()` should just handle everything.

wtgee:
#481 is an attempt to move entirely over to the astropy mechanism, which should hopefully solve this problem.

martinjohndyer:
@wtgee Thanks, that looks like a much better solution. I'm happy for this PR to be closed in favour of #481.

PR #491, opened 2020-11-30
DOC: Replace astropy channel with conda-forge

pllim:
Similar to astropy/astropy#11111

cc @mwcraig


bmorris3:
Is this good to go @mwcraig?

mwcraig:
@bmorris3 --  yes. The idea here is to try this in a few places and see what we hear back (if anything) from users.

PR #493, opened 2020-12-15
Move to APE 17 infrastructure and replace travis with github actions (Ready for review)

mwcraig:
This pull request will move astroplan from `astropy-helpers` to [APE 17](https://github.com/astropy/astropy-APEs/blob/master/APE17.rst) (`tox`, `setuptools_scm`, etc.) and replaces Travis-CI with GitHub Actions.

**Link to builds:** https://github.com/mwcraig/astroplan/actions

This PR:

1. Sets the minimum python version to 3.7
1. Bumps the minimum versions for numpy and astropy 
1. Drops astropy helpers
1. Switches from Travis-CI to GitHub actions 

 To do:

- [x] Add action workflow with a few essential tests
- [x] Revise sphinx configuration for APE17
- [x] Ask about test I had to modify
- [x] Make GA matrix match, roughly, old travis matrix.
- [x] Verify that minimum python version is correct
- [x] Ask about better way to do Sphinx links to astropy
- [x] Restore restriction on pushes to the default branch 👈👈👈 Do this BEFORE merging


astroplan/tests/test_scheduling.py
```diff
@@ -75,7 +75,7 @@ def test_schedule_insert_slot():
     end_time = start + duration
     block = TransitionBlock.from_duration(duration)
     schedule.insert_slot(end_time - duration, block)
-    assert not end_time - duration == start
+    assert end_time - duration == start
```
mwcraig, line 78: @bmorris3 @wtgee -- I wasn't sure about this change but had to make it to get tests passing. Was it still the expectation that `end_time - duration != start`?

astroplan/plots/sky.py
```diff
@@ -195,7 +195,7 @@ def plot_sky(target, observer, time, ax=None, style_kwargs=None,
     ]
 
     theta_labels = []
-    for chunk in range(0, 7):
+    for chunk in range(0, 8):
```
mwcraig, line 198: @bmorris3 @wtgee -- made this change to match the tick labels that are defined below. Without it, the doc build fails.

setup.cfg
```diff
@@ -1,33 +1,85 @@
...
+long_description = file: README.rst
+long_description_content_type = text/x-rst
 edit_on_github = False
 github_project = astropy/astroplan
 
-[entry_points]
+[options]
+zip_safe = False
+packages = find:
+python_requires = >=3.6
```
mwcraig, line 17: @bmorris3 @wtgee -- I took a guess here as to the correct version. I would recommend not supporting Python 2 anymore, and even python 3.5 is pretty old at this point.

setup.cfg
```diff
@@ -1,33 +1,85 @@
...
+setup_requires = setuptools_scm
+install_requires =
+    numpy>=1.10
+    astropy>=1.3
```
mwcraig, line 21: @bmorris3 @wtgee  -- Can the astropy and/or numpy minimum versions be updated? I'd suggest something more like `numpy>=1.17` and `astropy>=4` though maybe astropy 3 works.

setup.cfg
```diff
@@ -1,33 +1,85 @@
...
+[options.extras_require]
+all =
+    matplotlib>=1.4,<3.3
+    astroquery
+test =
+    pytest-astropy
+docs =
+    sphinx-astropy
+    sphinx_rtd_theme
+    matplotlib>=1.4,<3.3
```
mwcraig, line 34: @bmorris3 @wtgee  -- Any idea if the matplotlib restriction `<3.3` is still needed?

mwcraig:
@tepickering -- if you have some time to look at this that would be great. The only thing I had to do that there really seems like a better solution for are the links to the Astropy project that I changed in this commit: 27e8b633fe02217befdeed8e3b83257ddec5473b

That was the only way I found to avoid warnings that there was a duplicate reference.

setup.cfg
tepickering: i would set minimum to 3.7 at this point unless there's a specific reason to support 3.6. iirc, astropy has dropping support for 3.6 in its timeline, at least.

setup.cfg
tepickering: yeah, since 4.0.x is the new LTS, i hard agree that these should be updated.

tepickering:
might it now be worth including python 3.9 in the test matrix? maybe just one test with default deps as a sanity check.

tepickering (commented):
looks good to me! main changes i'd recommend would be to make sure the min/max versions of things are handled consistently and updated where appropriate.

docs/references.txt
```diff
@@ -1,4 +1,3 @@
-.. _Astropy: http://astropy.org
```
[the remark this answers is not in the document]
mwcraig: Yes -- I'm still mystified as to why RST doesn't easily allow this kind of thing.

setup.cfg
mwcraig: Matplotlib started checking whether the number of tick labels matches the number of ticks. See #468 and #469. I've removed my temporary fix for the issue, which was the same as #469 and kept the matplotlib pinning for now.

astroplan/tests/test_scheduling.py
wtgee: Yeah, I'm confused as to why that was passing before. @bmorris3?

docs/references.txt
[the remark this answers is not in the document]
wtgee: Sorry, what's the reason this stopped working? And why would the other urls still be working?

setup.cfg
wtgee: Yeah, #469 is a silly reason to hold us back on matplotlib, so would be nice to get that merged (@bmorris3)

wtgee (approved):
Thanks @mwcraig and sorry it took so long.  I'm assuming most of this just conforms with the updates to APE17, so I didn't nitpick some of the items.  We'll need @bmorris3 to merge #469 to get the matplotlib update, which would be nice, but could happen afterward.

docs/references.txt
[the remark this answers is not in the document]
mwcraig: It builds without removing this but generates warnings, and the warning were promoting to errors, so I moved the definition inline everywhere. I believe none of the other names defined here show up in more than one place.  An alternative would be to revert the change in links and change the documentation build so that warnings are not treated as errors.

wtgee:
Pinging @bmorris3 on a merge.

astroplan/tests/test_scheduling.py
bmorris3: I'm also stumped by this. The change was introduced in https://github.com/astropy/astroplan/commit/221d0d68bc23f82d540e22ffa8e8e5b5137655ab.

PR #498, opened 2021-01-27
Unit error in periodic tutorial

mrtommyb:
pl_trandur already has the unit of days. I assume that this is from an upstream change in astroquery


bmorris3:
Thanks @mrtommyb!

PR #507, opened 2021-04-26
Added a "name" attribute to ObservingBlock

thusser:
Added a "name" attribute to ObservingBlock, which can be used to assign a user-defined name or ID to a block. Should probably be used in __repr__ as well, if present.


bmorris3:
Looks good, thanks @thusser.

PR #529, opened 2022-11-01
`months_observable` update + fix

tygger7:
changed `months_observable` to `target_observable`  by adding ability to chose ('days','weeks', or 'months') to check against constraints. Default is 'month's to maintain functionality.

 Also fixed `_current_year_time_range` as it was not including the 31st of Dec when run through `months_observable` (found by using 'days'). `months_observable` was the only function to use this so the change is simple.

### DIFF
--- a/astroplan/constraints.py
+++ b/astroplan/constraints.py
@@ -40,7 +40,7 @@ __all__ = ["AltitudeConstraint", "AirmassConstraint", "AtNightConstraint",
 _current_year = time.localtime().tm_year  # needed for backward compatibility
 _current_year_time_range = Time(  # needed for backward compatibility
     [str(_current_year) + '-01-01',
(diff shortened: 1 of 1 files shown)


tygger7:
found error in my request will recommit

bmorris3:
Hi @tygger7, thanks for this contribution. People are already using `months_observable` since it was released six years ago in #152, so I we shouldn't remove it. If you'd like to add more  functions like `months_observable`, you might consider opening another pull request which adds a new function that supports outputs in days/weeks/months, and then modify the contents of the `months_observable` function to call the more generic function that you've written. Some names to consider for the generic function could be something like `observable_interval(interval='months')`.

tygger7:
Thank you. I didn't consider that. I'll make some changes and open a new pull_request.

PR #531, opened 2022-11-02
`Observable_interval` function added and `months_observable` aliased

tygger7:
Added a generic version `months_observable` called `observable_interval`. I also aliased `months_observable` to maintain backwards comparability, and help clean the code base. `weeks_observable` and `days_observable` aliases were also added; I felt it appropriate with `months_observable` remaining in the code. Lastly, I fixed a small issue with the "_current_year_time_range" as I noticed that it was not calculating visibility for December 31st in its current form.

Let me know if there is anything you think might need changed.

### DIFF
--- a/astroplan/observer.py
+++ b/astroplan/observer.py
@@ -7,7 +7,6 @@ from six import string_types
 import sys
 import datetime
 import warnings
-
 # Third-party
 from astropy.coordinates import (EarthLocation, SkyCoord, AltAz, get_sun,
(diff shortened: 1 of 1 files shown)


tygger7:
I have fixed the code style issues that were preventing check completion.

tygger7:
I have correct the issues that were indicated in the workflow.

bmorris3:
This is looking good. Could you please add tests for the new methods `days_observable` and `weeks_observable`?

PR #532, opened 2022-11-15
Enabled calculation of lunar and solar transit

lunarspectrum:
copied code from Observer._calc_riseset into ._calc_transit to enable calculations of lunar and solar transit.

### DIFF
--- a/astroplan/observer.py
+++ b/astroplan/observer.py
@@ -512,6 +512,11 @@ class Observer(object):
         >>> target_altaz = apo.altaz(time, target) # doctest: +SKIP
         """
         if target is not None:
+            if target is MoonFlag:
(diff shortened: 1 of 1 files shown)


bmorris3:
Hi @lunarspectrum, I've pushed a small improvement to your improvement. Rather than copying/pasting code to multiple places, I put the `*Flag` handling blocks in `Observer.altaz`, which seems more efficient.

lunarspectrum:
Hi @bmorris3 , thanks for doing that. I copied it from the other location because I didn't have the full picture of dependencies and didn't want to break any other code. Much appreciated.

bmorris3:
Thanks so much for your first contribution! 🎉

PR #538, opened 2023-02-27
Tutorial: fix indentation to plot multiple charts

SultanOrazbayev:
This will allow viewing charts for multiple targets (seems to be more consistent with the tutorial text).


bmorris3 (approved):
Thanks for this fix, you're right!

PR #540, opened 2023-03-06
Improve plot_schedule_airmass() performance by drawing night bands only once

### LINKED ISSUE #539: `plot_schedule_airmass` with `show_night=True` is immensely slow
Greetings,

I've noticed that running `plot_schedule_airmass()` is very slow when I activate the argument `show_night=True`, I'm talking like 8 seconds for a very simple plot. As [this comment](https://github.com/astropy/astroplan/blob/main/astroplan/plots/time_dependent.py#L496) realized there is a lot of overlap here. My suggestion is calculating all midnights first and coloring in the nights once:

```python
if show_night:
        midnights = []
        test_time = schedule.start_time
        while (midnight := schedule.observer.midnight(test_time, which='next')) < schedule.end_time:
            test_time = midnight + 6 * u.hour
            midnights.append(midnight)

        # Ceating darker bands
        for midnight in midnights:
            previous_sunset = schedule.observer.sun_set_time(
                midnight, which='previous')
            next_sunrise = schedule.observer.sun_rise_time(
                midnight, which='next')

            previous_twilight = schedule.observer.twilight_evening_astronomical(
                midnight, which='previous')
            next_twilight = schedule.observer.twilight_morning_astronomical(
                midnight, which='next')

            plt.axvspan(previous_sunset.plot_date, previous_twilight.plot_date,
                        facecolor='lightgrey', alpha=0.5)
            plt.axvspan(previous_twilight.plot_date, next_twilight.plot_date,
                        facecolor='lightgrey', alpha=0.7)
            plt.axvspan(next_twilight.plot_date, next_sunrise.plot_date,
                        facecolor='lightgrey', alpha=0.5)
```

This makes it a lot faster.

michaelbaisch:
Fixes #539

### DIFF
--- a/astroplan/plots/time_dependent.py
+++ b/astroplan/plots/time_dependent.py
@@ -160,7 +160,7 @@ def plot_airmass(targets, observer, time, ax=None, style_kwargs=None,
         tzname = time.tzname()
         tzinfo = time.tzinfo
     else:
-        tzoffset = 0
+        tzoffset = 0 * u.hour
         tzname = 'UTC'
         tzinfo = None
     # Populate time window if needed.
--- a/astroplan/plots/time_dependent.py
+++ b/astroplan/plots/time_dependent.py
@@ -202,40 +202,51 @@ def plot_airmass(targets, observer, time, ax=None, style_kwargs=None,
     # Shade background during night time
     if brightness_shading:
         start = time_ut[0]
+        end = time_ut[-1]
+        nights = []
 
-        # Calculate and order twilights and set plotting alpha for each
-        twilights = [
-            (observer.sun_set_time(start, which='next'), 0.0),
-            (observer.twilight_evening_civil(start, which='next'), 0.1),
-            (observer.twilight_evening_nautical(start, which='next'), 0.2),
-            (observer.twilight_evening_astronomical(start, which='next'), 0.3),
-            (observer.twilight_morning_astronomical(start, which='next'), 0.4),
-            (observer.twilight_morning_nautical(start, which='next'), 0.3),
-            (observer.twilight_morning_civil(start, which='next'), 0.2),
-            (observer.sun_rise_time(start, which='next'), 0.1),
-        ]
+        next_sun_rise = observer.sun_rise_time(start, which='next')
+        sun_set = observer.sun_set_time(next_sun_rise, which='previous')
+        while (sun_set < end):
 
-        # add 'UTC' to each datetime object created above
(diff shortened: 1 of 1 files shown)


wtgee:
Hi @michaelbaisch, thanks for the PR. I’ll admit this is the first time I’ve looked at this code in detail, but I’m a bit confused. The night shading is already handled by `plot_airmass`, so it seems like `plot_schedule_airmass` could conform to that by simply passing the `show_night` variable into the `plot_airmass` call on line 493 and then removing the midnight plotting contained within `plot_schedule_airmass`.

Doing so produces the following plot:

image


@bmorris3?

michaelbaisch:
Some thoughts: `plot_airmass` is called for each target in `plot_schedule_airmass`. So `brightness_shading` should only be set to True for one of those calls, otherwise the shading would be drawn multiple times again.
I was using the `:=` operator, which was added in python 3.8, but there is still a test for python 3.7, is that a hard dependency?

bmorris3:
> Some thoughts: plot_airmass is called for each target in plot_schedule_airmass. So brightness_shading should only be set to True for one of those calls, otherwise the shading would be drawn multiple times again.

Right, so it should be possible to set `brightness_shading=True` for only the first iteration in this loop that calls `plot_airmass`, right?

https://github.com/astropy/astroplan/blob/fda875841bded74a3e82daee8c85b002e56400c5/astroplan/plots/time_dependent.py#L492-L493


> I was using the := operator, which was added in python 3.8, but there is still a test for python 3.7, is that a hard dependency?

I'm not sentimental about 3.7, but I'm not sure that one use of the walrus op is a sufficient reason to drop 3.7 support.

bmorris3:
> I'm not sentimental about 3.7, but I'm not sure that one use of the walrus op is a sufficient reason to drop 3.7 support.

This might force our hand though: https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table

michaelbaisch:
Not sure if I can do something about the failed "Test building of Sphinx docs (pull_request)"?

bmorris3:
@michaelbaisch I'll investigate that failing test in a separate PR and follow up here when I sort it out. Thanks!

bmorris3:
@michaelbaisch I've dug into it a bit, and I haven't sorted it out why this test fails on GitHub. It passes for me locally, so I'm going to merge and hope that #544 has already solved the problem. 🤞🏻 

Thanks!

michaelbaisch:
@bmorris3 why was this reverted?

PR #548, opened 2023-04-05
Adding telescope constraints to the time_dependent plots

SunilSimha:
Hello!

I'm proposing the addition of `Constraint`s to the `time_dependent` plots, specifically `plot_airmass`. Currently, my changes add a thick black line where the telescope constraints are not satisfied (i.e. the object is not observable when it's marked black). 
image
The figure above shows an example image with my custom `Constraint` class called `KeckVisibility` which takes into account the various telescope pointing constraints at Keck I and Keck II.

I've not made any changes to the other plots yet as they're plotted somewhat differently but am happy to take suggestions for those as well.

Regards,
Sunil

### DIFF
--- a/astroplan/plots/time_dependent.py
+++ b/astroplan/plots/time_dependent.py
@@ -2,6 +2,7 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 import copy
+from xmlrpc.client import boolean
 import numpy as np
 import operator
 import astropy.units as u
@@ -12,6 +13,7 @@ import pytz
 
 from ..exceptions import PlotWarning
 from ..utils import _set_mpl_style_sheet
+from ..constraints import Constraint
 
 __all__ = ['plot_airmass', 'plot_schedule_airmass', 'plot_parallactic',
            'plot_altitude']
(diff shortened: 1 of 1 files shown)


astroplan/plots/time_dependent.py
```diff
@@ -2,6 +2,7 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 import copy
+from xmlrpc.client import boolean
```
bmorris3, line 5: Is this used anywhere?

astroplan/plots/time_dependent.py
```diff
@@ -176,6 +183,23 @@ def plot_airmass(targets, observer, time, ax=None, style_kwargs=None,
...
+        for cons in constraints:
+            assert isinstance(cons, Constraint), "All constraint objects must be Constraints"
+
+            # Make sure they're boolean
+            assert hasattr(cons, 'boolean_constraint'), "Each constraints element must have the 'boolean_constraint attribute."
+            assert cons.boolean_constraint, "Cannot work with non-boolean constraints."
+            # Then make sure they're not some weird user-generated constraint. 
+            assert hasattr(cons, 'compute_constraint'), "Constraints must have a 'compute_constraint' method."
+            
+
```
bmorris3, line 202: Rather than including `assert`s in the function here, a more common approach is to extract the elements of the list that meet the validation criteria, without raising an error.

bmorris3 (changes requested):

PR #551, opened 2023-05-29
AltitudeConstraint: Adjust max_best_rescale() to prevent unexpected output

michaelbaisch:
In previous versions, when setting the max parameter for altitude, max_best_rescale() would return 1.0 if the altitude exceeded the maximum limit. This behavior was unexpected.

To rectify this, the greater_than_max argument in max_best_rescale() has been set to 0. This adjustment ensures that the output aligns with expectations when the altitude surpasses the specified maximum constraint.


bmorris3:
Thanks @michaelbaisch!

PR #554, opened 2023-06-01
Remove deprecated call to astropy helpers.

### LINKED ISSUE #553: Remove deprecated astropy helper functions.
Need to remove this old function. `pytest` now warns by default but that seems different from an exception being raised. Not sure we really need exception raised since the warning is shown at end of test run.

              Failure is from astropy removing old code, which you should definitely fix, but unrelated to this PR.

_Originally posted by @pllim in https://github.com/astropy/astroplan/issues/552#issuecomment-1572878114_

wtgee:
Removes the function call and defaults to pytest just showing the warning at end of test run. Thanks @pllim for pointing out.

Edit: Astropy issue here https://github.com/astropy/astropy/issues/9718 and PR https://github.com/astropy/astropy/pull/14670

Closes #553

P.S. I feel like there is some kind of meta irony in us not noticing this was deprecated...


pllim:
To retain the functionality you are removing here, you need to change

https://github.com/astropy/astroplan/blob/cb9517e381a3e8338e8d182e5f130201139f7fc7/setup.cfg#L44

to turn warning into exceptions using

https://docs.pytest.org/en/stable/how-to/capture-warnings.html

wtgee:
Now I'm confused why this wasn't working before. Currently passing PRs show deprecations as warnings rather than raising actual exceptions.

pllim:
The pytest way is probably more sensitive than what `astropy` was providing.

pllim:
You should probably drop Python 3.7. It is picking up a very old `astropy` (4.3.1).

bmorris3:
Hi @wtgee, sorry to begin reviewing this PR after making and merging #558, which I believe has the same effect as this PR. Shall we close this one?

wtgee:
Hey @bmorris3, this PR properly turns on the deprecations as exceptions via pytest, which replaces the functionality that was supposed to be working with `enable_deprecations_as_exception`. #558 removes that but doesn't add the pytest filter. 

It's up to you if that should be in the repo but adding it in shows a bunch of other deprecations that should probably be dealt with soon (hence the follow-up issues about python 3.7, etc), so it's arguably nice to retain that functionality.

PR #562, opened 2023-07-18
Add __hash__, __eq__, and __ne__ support for Observer objects

grant-nations:
Adds the ability to hash and compare Observer objects. This is helpful for adding LRU cache decorators to functions that take an Observer as a parameter.


bmorris3:
This is handy, thanks!

PR #572, opened 2023-09-29
Add remote_data on all tests that require network

sergiopasra:
This PR implements a first step towards the goal of running the tests of astroplan without network access. I'm using the decorator `@pytest.mark.remote_data` from the package `pytest-remotedata`, which is currently in the dependencies. In the long term, I would like to check one by one the marked tests and see if the network access is actually required or if it is an 
*accident*, typically by using `Observer.at_site`


bmorris3:
Neat, thanks! We used to prevent problems due to internet access this with mocks like this:

https://github.com/astropy/astroplan/blob/7880daad50733fc4d5c6df0a6fd077bac3fefe39/docs/conf.py#L159-L160

Could you add the `--remote-data` flag to the test matrix, e.g. here:

https://github.com/astropy/astroplan/blob/f1bd5036534667e81339dc8f582c7ca0a03bdca2/tox.ini#L76

sergiopasra:
I have added a factor `-remote` that enables remote data in all tests. The tests without `-remote` use only local data. If you prefer that `-cov` tests are run always with remote data, it can be done also.

astroplan/constraints.py
```diff
@@ -734,7 +734,7 @@ def __init__(self, min=None, max=None):
         >>> from astroplan import Observer
         >>> from astroplan.constraints import LocalTimeConstraint
         >>> import datetime as dt
-        >>> subaru = Observer.at_site("Subaru", timezone="US/Hawaii")
+        >>> subaru = Observer.at_site("Subaru", timezone="US/Hawaii")  # doctest: +REMOTE_DATA
```
pllim, line 737: Why is `subaru` needed here at all? 🤔

astroplan/target.py
```diff
@@ -127,7 +127,7 @@ def from_name(cls, query_name, name=None, **kwargs):
         Examples
         --------
         >>> from astroplan import FixedTarget
-        >>> sirius = FixedTarget.from_name("Sirius")
+        >>> sirius = FixedTarget.from_name("Sirius")  # doctest: +REMOTE_DATA
```
pllim, line 130: Problem with marking lines with remote data, you also need to mark all the subsequent lines derived from here onwards.

sergiopasra: Yes, there are some tests where I should mark also the next line

astroplan/constraints.py
sergiopasra: It's true, that variable is not used

astroplan/target.py
sergiopasra: I have also noticed that I was using REMOTE_DATA in some tests that don't need it. I have removed them

tox.ini
pllim, line 1: There is a way to control remote data without touching this file. I will revert the changes here. FYI.

pllim:
Thank you very much for smoking out the remote data tests, @sergiopasra ! I took some liberty to push follow up commits here. Hope you don't mind.

PR #589, opened 2024-08-09
Added near meridian constraint

ramonefoster:
Created NearMeridianContraint, that can be useful for observations using German-Equatorial Mounts, to avoid flipping the side of pier during exposures.


astroplan/constraints.py
```diff
@@ -938,6 +938,46 @@ def compute_constraint(self, times, observer=None, targets=None):
...
+class NearMeridianConstraint(Constraint):
+    """
+    Constraint near the Meridian.
+    """
+    def __init__(self, min=None, boolean_constraint=True):
+        """
+        Parameters
+        ----------
+        min : `~astropy.units.Quantity` or `None`, optional
+            Minimum acceptable distance to meridian.
```
bmorris3, line 950:
```suggestion
            Minimum acceptable angular distance from the meridian.
```

bmorris3 (commented):
Thanks @ramonefoster, this is great! 

My only real comment is about the name. When you first read it, it may be a bit ambiguous if this constraint includes or excludes targets are near the meridian.

Could we rename it `MeridianSeparationConstraint`, and add a `max` keyword argument? This would make it clear that the constraint sets angular limits relative to the meridian.

ramonefoster:
Hello, @bmorris3 . Thanks for the suggestion, it was indeed a needed change. I have it fixed.

astroplan/constraints.py
bmorris3: @ramonefoster Could you add this change too?

ramonefoster: I thought I did, but I didnt haha... anyways, I made the changes :)

PR #591, opened 2024-08-25
Remove nonfunctional grid keyword in finder plot

### LINKED ISSUE #588: SkyView
Not sure if this is an issue for others, but in case it is:

In order to use plot_finder_image with Astropy v 6.1.2, Astroquery 0.4.8, and Astroplan 0.11, I had to remove the grid=grid argument in SkyView.get_images in the file finder.py (line 88-89) under astroplan because the get_images function doesn't take a grid parameter. So even though it is by default False in finder.py, astroquery threw an error my way from SkyView/core.py.

keflavich:
Fixes #588.

As noted on #588, a recent change to the SkyView module resulted in the grid keyword being removed:
https://github.com/astropy/astroquery/pull/2979

### DIFF
--- a/astroplan/plots/finder.py
+++ b/astroplan/plots/finder.py
@@ -12,7 +12,7 @@ __all__ = ['plot_finder_image']
 
 @u.quantity_input(fov_radius=u.deg)
 def plot_finder_image(target, survey='DSS', fov_radius=10*u.arcmin,
-                      log=False, ax=None, grid=False, reticle=False,
+                      log=False, ax=None, reticle=False,
                       style_kwargs=None, reticle_style_kwargs=None):
     """
     Plot survey image centered on ``target``.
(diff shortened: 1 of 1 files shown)


keflavich:
I added the `skyview_kwargs` option to pass kwargs to skyview, since the default arguments result in some very badly pixelated images if you change the field of view.

pllim:
You have to fix code style and rebase. Thanks!

keflavich:
working on it, but I can't see where the code style is failing b/c it's out of date.  We need to update the testing infrastructure to get anything green.

pllim:
Are the failures related? I cannot tell.

Also you might wanna rebase again as I also just merged  #624

keflavich:
no the failures were only in documentation, unrelated... though I don't know why they'd be red here and green in the infrastructure PR

pllim:
@keflavich , I think it is related. But I don't understand why the simpler solution at https://github.com/astropy/astroplan/pull/622 fixes better than this PR.

astroplan/plots/finder.py
```diff
@@ -13,7 +13,8 @@
 @u.quantity_input(fov_radius=u.deg)
 def plot_finder_image(target, survey='DSS', fov_radius=10*u.arcmin,
                       log=False, ax=None, grid=False, reticle=False,
-                      style_kwargs=None, reticle_style_kwargs=None):
+                      style_kwargs=None, reticle_style_kwargs=None,
+                      skyview_kwargs={}):
```
pllim, line 17: I don't think this is clean enough. It let the `grid=` through from example code in docs, and then fails the doc build.

pllim:
I think I am closing this in favor of #622 . Is there a reason why we need to catch other kwargs?

PR #593, opened 2024-09-16
Add type hints

### LINKED ISSUE #592: Add type hints
It would be great to have type hints. I started adding them in my local version, I can open a PR when more significant work has been done, but I wanted to open this issue for tracking.

vandalt:
Hi! This adds type hints and fixes #592.

For astropy quantities, I used `Quantity` without any physical type or units annotations. The reasons are:

- This caused error when using `Union` with non-Quantity types (see https://github.com/astropy/astropy/issues/17017)
- In retrospect, this would have been equivalent to changing the behaviour of astroplan, by introducing new unit validation that was not present before this PR, which I think would have been out of scope anyways.

The tests are all passing on my machine.

I was not sure what the formatting guidelines were for this package. Let me know if there are issues on that side!

### DIFF
--- a/astroplan/constraints.py
+++ b/astroplan/constraints.py
@@ -9,7 +9,7 @@ from __future__ import (absolute_import, division, print_function,
 
 # Standard library
 from abc import ABCMeta, abstractmethod
-from typing import Optional, Sequence, Union
+from typing import Optional, Self, Sequence, Union
 import datetime
 import time
 import warnings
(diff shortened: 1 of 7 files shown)


vandalt:
I updated formatting so that flake8 check would pass.

I also patched the `typing.Self` to be backward compatible for Python<3.11.

Currently the tests for Python 3.7 are failing because I used types with subscripts (e.g. `np.ndarray[float]`). Python 3.7 is EOL, but let me know if I should still make the type hints backward compatible.

PR #594, opened 2024-09-20
Fix some warnings in test

vandalt:
This fixes two warnings I saw when running tests:

1. Deprecation warning for `ax.plot_date`: fixed by using `ax.plot` with `datetime64` instead
2. Warnings about non-rotation transforms when calculating Sun and Moon separations. Fixed by adding `origin_mismach="ignore"` to those two calculations. There is already a comment about this before both instances and removing the warning avoids end users seeing it and wondering if something is wrong with how they specified coordinates.

### DIFF
--- a/astroplan/constraints.py
+++ b/astroplan/constraints.py
@@ -551,7 +551,10 @@ class SunSeparationConstraint(Constraint):
         # 'get_sun' returns ICRS coords.
         sun = get_body('sun', times, location=observer.location)
         targets = get_skycoord(targets)
-        solar_separation = sun.separation(targets)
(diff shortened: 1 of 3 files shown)


pllim:
> separation() got an unexpected keyword argument 'origin_mismatch'

The problem with this is that astroplan pins to too old of minversion of astropy etc. So we either have to bump minversion or do a version check.

Would love to see this clean up go in!

Also would have to un-ignore the relevant warnings here:

https://github.com/astropy/astroplan/blob/bf5e26d0697aaaf710d5879322c5793866249d24/setup.cfg#L54-L57

Thanks!

PR #611, opened 2025-01-06
Fix `TimeDeltaMissingUnitWarning` in `plot_airmass`

H-xie:
The function `plot_airmass` triggers the warning `TimeDeltaMissingUnitWarning`. This issue arises from unit checks implemented in newer versions of Astropy (refer to commit d8968b29a209fdaebd0f48ff30863718bca9db78).

The warning is generated because a time value without a unit is being added to an Astropy Time object. To resolve this warning, it is recommended to add the unit `u.hour` to the time value.

### DIFF
--- a/astroplan/plots/time_dependent.py
+++ b/astroplan/plots/time_dependent.py
@@ -161,7 +161,7 @@ def plot_airmass(targets, observer, time, ax=None, style_kwargs=None,
         tzname = time.tzname()
         tzinfo = time.tzinfo
     else:
-        tzoffset = 0
+        tzoffset = 0 * u.hour
         tzname = 'UTC'
         tzinfo = None
     # Populate time window if needed.


m4cd4r4 (commented):
Clean fix for the TimeDeltaMissingUnitWarning. One thing worth checking:

In the datetime branch above this line, `tzoffset` is set to a bare float (from `time.utcoffset().total_seconds() / 3600`). If that value is later added to an astropy `Time` object, the same warning would appear for timezone-aware datetime inputs.

Worth adding `* u.hour` to the datetime branch too:

```python
tzoffset = time.utcoffset().total_seconds() / 3600 * u.hour
```

Otherwise this is a straightforward one-liner that fixes a real warning. Looks like it needs a rebase to trigger CI.

keflavich:
failures are unrelated - they're dead links and there's a link checker

PR #605, opened 2024-12-06
Infrastructure clean-up

pllim:
Note to whoever merges this: Please use **Squash and Merge** button.

Stuff needs modernization.

- [x] Where is codecov bot?
- [x] Switch to OpenAstronomy
- [x] Add publish workflow and have it also run in PR

Need to do this after merge:

* #607 

Would make the following fixes easier going forward:

* #601 
* #603

Might naturally close the following when you next do a release after merging this:

* #583

PR #623, opened 2025-09-15
Update Python versions in CI testing

cgobat:
CI was previously trying to run on Python 3.7 and 3.9, neither of which even exist in [tox.ini](../blob/main/tox.ini) anymore. 3.7 is [end-of-life](https://devguide.python.org/versions/) and 3.9 is about to be.


pllim:
Superseded by #605

PR #625, opened 2025-10-07
DOC: Remove grid keyword in plot examples

pllim:
I have no idea why it was happening. Just trying to get things green so Adam can do his thing.


pllim:
Ooops wrong tree

PR #622, opened 2025-09-12
Remove `grid` kwarg from `SkyView.get_images()` call

cgobat:
As of astropy/astroquery#2979 (May 2024), `SkyView.get_images()` no longer accepts a `grid` keyword argument. This pull request removes it from `plot_finder_image()` to prevent breakage/maintain compatibility.

Close https://github.com/astropy/astroplan/pull/591

Fixes https://github.com/astropy/astroplan/issues/588

PR #626, opened 2025-10-07
MNT: SkyView no longer takes grid

pllim:
Since https://github.com/astropy/astroquery/pull/2979

Fix https://github.com/astropy/astroplan/issues/588

Close https://github.com/astropy/astroplan/pull/591


pllim:
Oh wait this turns out to be same as #622

PR #566, opened 2023-08-05
Implement support for targets defined by TLE (Two-Line Element set)

michaelbaisch:
* Add TLETarget class that uses skyfield internally to return RA/Dec values for specific times
* The get_skycoord() function now takes a 'time' argument because the returned SkyCoord RA/Dec values are time-dependent for TLETargets
* get_skycoord() behavior for non-TLETargets and multiple times can be controlled with the 'backwards_compatible' argument
* Only perform 'grid_times_targets' when targets and times are not broadcastable
* Update the schedule table so it also displays sensible info for TLETargets
* Update Tests

Fixes #565

For get_skycoord(), I opted for the backwards compatibility option by default. There are more details in the code, but the basic question is: Do we want to be 100% backwards compatible, or do we prefer consistent output? Having said that, everything that I've tested works well like this.

The new table looks like this:

```
              target                  start time (UTC)         end time (UTC)     duration (min)  type          coordinates/info         configuration
--------------------------------- ----------------------- ----------------------- -------------- ------ -------------------------------- -------------
  Capella (Alpha Aurigae) (State) 2023-08-04 10:09:25.239 2023-08-04 10:09:45.239         0.3333 RA/Dec         05h16m41.4s +45d59m52.8s            {}
                  TransitionBlock 2023-08-04 10:09:45.239 2023-08-04 10:10:05.239         0.3333                                                    []
 Sirius A (Alpha Canis Majoris A) 2023-08-04 10:10:05.239 2023-08-04 11:09:25.239        59.3333 RA/Dec           06h45m08.9s -16d42m58s            {}
                  TransitionBlock 2023-08-04 11:09:25.239 2023-08-04 11:09:45.239         0.3333                                                    []
     Castor A (Alpha Geminorum A) 2023-08-04 11:09:45.239 2023-08-04 12:06:45.239        57.0000 RA/Dec             07h34m36s +31d53m18s            {}
                  TransitionBlock 2023-08-04 12:06:45.239 2023-08-04 12:07:05.239         0.3333                                                    []
       Betelgeuse (Alpha Orionis) 2023-08-04 12:07:05.239 2023-08-04 12:16:45.239         9.6667 RA/Dec         05h55m10.3s +07d24m25.4s            {}
                  TransitionBlock 2023-08-04 12:16:45.239 2023-08-04 12:17:05.239         0.3333                                                    []
                      Unused Time 2023-08-04 12:17:05.239 2023-08-04 12:24:25.244         7.3334                                                      
[...]

PR #634, opened 2026-02-12
ENH: Add AltAzTarget and support time-dependent targets

michaelbaisch:
- Add AltAzTarget for targets defined in horizontal coordinates at a fixed EarthLocation, evaluated to time-dependent ICRS SkyCoord when needed; support vacuum (pressure=0) and apparent (pressure>0) interpretations via stored atmospheric parameters.
- Export AltAzTarget and make get_skycoord(time-aware): accept `times`, evaluate time-dependent targets via get_skycoord(times), and broadcast coords.
- Pass `time` into get_skycoord from Observer._preprocess_inputs to enable time-dependent target evaluation during constraint calculations.
- Update Constraint.__call__ to delegate gridding and time-dependent evaluation to Observer._preprocess_inputs.
- In scheduling add `target_type`/`target_info` columns in Schedule.to_table.
- Add tests.

#566 was updated to be compatible with this PR.

### DIFF
--- a/astroplan/constraints.py
+++ b/astroplan/constraints.py
@@ -15,7 +15,7 @@ import astropy.units as u
 import numpy as np
 from astropy import table
 from astropy.time import Time
-from astropy.coordinates import get_body, get_sun, Galactic, SkyCoord
+from astropy.coordinates import get_body, get_sun, Galactic
 from numpy.lib.stride_tricks import as_strided
 
 # Package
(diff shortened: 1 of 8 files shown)


m4cd4r4 (commented):
Solid addition to astroplan. The AltAzTarget class fills a real gap for users working with local pointing directions rather than celestial coordinates. Test coverage is thorough (208 lines).

A few things I noticed:

**Breaking change in `Schedule.to_table()`:**
The `ra` and `dec` columns are replaced with `target type` and `target info`. Any downstream code that accesses `table['ra']` or `table['dec']` will break silently. Would it be possible to keep the old columns (populated where applicable) alongside the new ones? Or add a deprecation warning for one release cycle?

**Time-dependent detection heuristic:**
The `_is_time_dependent` check (has `get_skycoord` callable and no `coord` attribute) is duck-typing based. If someone later adds a `get_skycoord` method to `FixedTarget`, this logic breaks silently. A dedicated `is_time_dependent` property on the `Target` base class would be more explicit and safer against future changes.

**Atmospheric parameter defaults:**
The docstring says pressure defaults to 0 hPa (vacuum), but the code passes `pressure=None` to `AltAz`. Depending on the astropy version, `AltAz(pressure=None)` could use a different default. Setting `pressure = 0 * u.hPa` explicitly in `__init__` would guarantee vacuum behavior regardless of astropy version.

**Minor:**
- The `except Exception` blocks in `_format_target_info` are broad. Narrowing to `AttributeError` or `ConvertError` would avoid masking real bugs.
- The error message when `times=None` for a time-dependent target could mention that the caller needs to provide `times` (currently just raises a bare `ValueError`).

The refactoring of `_preprocess_inputs` to handle time-dependent gridding is clean, and the score_key rounding in the scheduler is a good catch. Nice work.

michaelbaisch:
I hope my changes address your points. I was just not sure what you meant by the last point about a bare `ValueError` when `times=None`, since the code currently raises `ValueError("`times` is required to evaluate an AltAzTarget.")`.

I have also updated https://github.com/astropy/astroplan/pull/566 accordingly.

m4cd4r4 (approved):
Thanks for the updates Michael. These look good.

You're right on the ValueError - I misread that, it already has a clear message. My mistake.

The rest of the changes look solid.

m4cd4r4:
Fair point - I was referring to the case where someone passes times=None to a method that feeds into _altaz(), but looking again, the ValueError message is clear enough. That last point is addressed.

The rest looks good after your latest push. Nice work on the scheduling integration.

evanmayer:
User here. It would be great to get this in. I'm having real pains writing observing plans that include planets.

